PR #1, opened 2016-08-13
Using chests

TheAIBot:


Danielv123:
Fits perfectly, thank you so much! I will se about getting the JS side of things working with this.

Danielv123:
Committed some changes to both the LUA mod and the filehooks to make everything work.

Need help to add #2 to the mod, as anything else would essentially make it a free items mod.

PR #19, opened 2016-10-02
Mod Support for circuit networks over clusterio

justarandomgeek:
Add Transmit/Receive Combinators. Transmit combinators are Deciders which also output into txbuffer.txt. Receive Combinators are Constants which replay frames received from external sources via remote calls to receiveFrame and recieveMany.

### DIFF
--- a/clusterio_0.0.2/config.lua
+++ b/clusterio_0.0.2/config.lua
@@ -18,9 +18,15 @@ OUTPUT_TANK_NAME = "get-tank"
 OUTPUT_TANK_PICTURE_PATH = nil
 OUTPUT_TANK_ICON_PATH = nil
 
+MAX_RX_BUFFER_SIZE = 40 -- slightly more than one second
+RX_COMBINATOR_NAME = "get-combinator"
+TX_COMBINATOR_NAME = "put-combinator"
+
 
 OUTPUT_FILE = "output.txt"
 ORDER_FILE  = "orders.txt"
+TX_BUFFER_FILE = "txbuffer.txt"
 
 
 UPATE_RATE = 60
+CIRCUIT_UPATE_RATE = 5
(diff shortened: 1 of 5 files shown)


Danielv123:
Thanks for the work! I will pull this into my dev branch and see about integrating it.

justarandomgeek:
It is necessary for the transmitter to check every tick, or it may miss pulses from the circuits - these devices are intended to work with pulses, which are only around for a single tick. It is likely that the clusterio client would then read multiple frames at once, which is perfectly fine - they're going into a buffer that feeds a buffer anyway, right?

The inter-game linking is missing because I don't (yet) node.js - I just said I'd do this part, so I did ;)

If two factories send multiple signals, they should be relayed one after another by the node.js master distributing them to worlds - in fifo order as seen by the master, presumably. 

And if the sum of all factories transmits signals faster than any world can play them back, frames are dropped in that world when receiving into it's buffer. It is the transmitting world's responsibility to wait/retransmit as needed.

justarandomgeek:
Yeah, perhaps in a future update frames will get a destination address and only be sent to that world if set. You'll see i have commented out signals for src/dest address - intended for this exact purpose! I just wasn't sure how to assign world IDs at this point, and left that for later.

Danielv123:
I am already delaying the reading of files to every 3 seconds because of problems and stuff, so thats not an issue. There is no need for it to request the signals from the master from the mod, client.js can just ask master every 3 seconds or so. I believe I can do this, but it will take a bit of time as I have a lot of stuff going on this week.

justarandomgeek:
> @Danielv123  
> I will pull this into my dev branch and see about integrating it.

Should I send pullreqs to dev in the future and/or resubmit this one there?

Danielv123:
Since I have to do a bit of work in the devbranch anyways, I just use normal merges so it does not matter. Otherwise, I guess pull requests to dev would be the best.

Danielv123:
Merged into dev.

PR #54, opened 2017-08-29
Clean up web interface storage tab.

DDR0:
- Use diffing algorithm so you can select text and images.
	- This uses textcontent, plugs a vulnerability where you could inject arbitrary javascript if you could upload a modded item with html in the name, eg, "iron-plate".
- Fall back to transparent png if item image not found.
	- This stops the flashing box when the image does not exist.
- (internal) Simplify `sortByKey(array, key)`.


DDR0:
hang on, forgot about filtering

PR #55, opened 2017-08-29
ui improvements for the web interface storage tab

DDR0:
See the two commit messages for details.


Danielv123:
These look like good changes I never got around to work on, will test tomorrow.

DDR0:
I have updated the storage tab, it uses much less data to do it's thing now.

PR #56, opened 2017-08-30
Another branch, built on top of my dev branch, which contains a scarce resource sharing implementation.

DDR0:
Initial stab at making the master server dole out resources more fairly when running a deficit. Seems to work. ¯\\_(ツ)_/¯

The server keeps an in-memory list of what it's recently doled out. Each element has a count associated with it, which is doubled when the server can't dole out the full amount of the resource. When the server can fulfil a request, the count is decremented. (I haven't been able to get steam to start multiple copies of Factorio, but the system ought to be impartial to who initiates the request.)


static/slaveDisplay.js
```diff
@@ -83,27 +85,26 @@ function post(url, data, callback) {
...
+			y: Number(y)
 		}
 	}
 	chartData = sortByKey(chartData, "x");
 	let xyz = {};
 	xyz.name = nameKey;
 	xyz.type = "line";
-	if(nameKey == "copper-wire"||nameKey == "iron-plate"||nameKey == "copper-plate"||nameKey == "electronic-circuit"||nameKey == "steel-plate"||nameKey == "advanced-circuit"||nameKey == "crude-oil"||nameKey == "petroleum-gas"){
+	
+	if(legendKeys.indexOf(nameKey) >= 0){
```
Danielv123: in ES6, use Array.includes("") instead.

PR #92, opened 2018-05-06
swapped from remote call to custom commands

MCRcortex:
Plz check first, should work


Danielv123:
Closing as command size is no longer that relevant

PR #130, opened 2018-07-18
This will build correctly

Sir3lit:
The build is successful :)


Danielv123:
Doesn't quite seem that way. 

```npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.```

Danielv123:
Oh, this was for docker. Welp, I guess I am dumb.

PR #150, opened 2019-04-02
NN update

axapaxa123:
Should mostly work, but sometimes overdraws.


Danielv123:
Wow I have no idea what happened to the master.js. Did you somehow replace all the newlines with windows or linux newlines? I see that in /routes/api/remove.js your editor has inserted tabs instead of spaces, which looks inconsistent. Now Im wondering what kind of editor you are using :P

axapaxa123:
Notepad++ I used for JS modifications. I don't think I replaced newlines in master.js, but I reuploaded file which automatically marks whole content as new afaik. How to fix it :P

axapaxa123:
Will add new PR with fixed master.js :P

PR #151, opened 2019-04-03
Neural Dole v2

axapaxa123:
This time with proper master.js :P


axapaxa123:
Added new version of NN along with other improvements. Should be fairly stable from my testing, NN divides items pretty nicely, no problems. Added gauge also works fine, but looks way better when averaged in Prometheus, seems to even make sense in severity. Also doesn't seem to falsely trigger unless items are indeed missing.

PR #153, opened 2019-05-13
player to character inventory renames and extra info for EACCESS problem

veladon:
Updated defines.inventory.player_<x> to defines.inventory.character_<x> as per 0.17.35 change to rename character entity type from "player" to "character". Removed player_quickbar and player_tools inventories as they no longer exist.

Also added extra instructions to readme for the EACCESS common problem.


Danielv123:
That Lua file isn't actually used directly, only for development. The functions are stringified and put in interfacingFunctions.js - Will still merge though, so its easier to resume development.

PR #166, opened 2019-07-15
 fix for stats exporter console errors

veladon:


MrRozdor:
@veladon `forceData: force.data` doesn't work because it is passed as `x-form-urlencoded` and every single value becomes a string
You need `needle.post(url, data, {json: true}, callback)`

PR #172, opened 2019-07-17
Remove GPS pings in crosschat

Sir3lit:
Cross-server GPS pings are annoying, let us fix that

Tested in test cluster with TheNetworkDoctor + Hornwitser,

Sin~ AYS

### DIFF
--- a/sharedPlugins/globalChat/index.js
+++ b/sharedPlugins/globalChat/index.js
@@ -22,7 +22,7 @@ module.exports = class remoteCommands {
 					words.shift();
 					words.shift();
 					words.shift();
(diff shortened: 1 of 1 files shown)


Sir3lit:
Yeh, it would be, but for the sake of the current event it would be useful to just dump the whole message, It was my quick fix <3. Feel free to improve it so it only strips the gps part.

psihius:
I have to agree with @Danielv123 here - should update the code to strip only the gps pings, not the whole message. Should be easy to do. Ask for help if needed :)

Master-Guy:
It might be useful for admins and other people to still be able to see the coordinates, without it being a ping in the other maps. I would therefor suggest something like this:
`replace("[gps=", "gps[")`

PR #173, opened 2019-07-17
Add "info" field between client and master

Master-Guy:
Adds a "info" field to config.json and allows the client to send the contents of this info field to the master, so it also shows up (i.e. with /api/slaves)


client.js
```diff
@@ -317,7 +317,7 @@ write-data=${ path.resolve(config.instanceDirectory, instance) }\r\n
 		"clientPort": args["rcon-port"] || process.env.RCONPORT || Math.floor(Math.random() * 65535),
 		"__comment_clientPassword": "This is the rcon password. Its also used for making an instanceID. Make sure its unique and not blank.",
 		"clientPassword": args["rcon-password"] || Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 8),
-		"info": {},
+		"info": {}
```
MrRozdor: It's JS. Commas at the object end doesn't matter

PR #176, opened 2019-07-19
/api/slaves POST (optimized)

Master-Guy:
Adds a /api/slaves POST call that requires authentication. It also returns the rconPasswords, unlike the GET call.
Optimized version of https://github.com/Danielv123/factorioClusterio/pull/175

### DIFF
--- a/master.js
+++ b/master.js
@@ -440,37 +440,60 @@ app.post("/api/uploadMod", authenticate.middleware, function(req,res) {
     }
 });
 /**
+Prepare the slaveCache for both /api/slaves calls
+*/
+let slaveCache = {
+	timestamp: Date.now(),
+};
+function getSlaves() {
+	if(!slaveCache.cache || Date.now() - slaveCache.timestamp > 5000){
+		let copyOfSlaves = JSON.parse(JSON.stringify(slaves));
+		slaveCache.cache = copyOfSlaves;
+		slaveCache.timestamp = Date.now();
+	}
+	return slaveCache;
+}
+/**
 GET endpoint for getting information about all our slaves
 @memberof clusterioMaster
 @instance
 @alias /api/slaves
 */
-let slaveCache = {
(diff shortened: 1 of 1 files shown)


master.js
```diff
@@ -440,37 +440,60 @@ app.post("/api/uploadMod", authenticate.middleware, function(req,res) {
...
+POST endpoint for getting information about all our slaves, requires auth, responds including RCON passwords
+@memberof clusterioMaster
+@instance
+@alias /api/slaves
+*/
+app.post("/api/slaves", function(req, res) {
+	let token = req.headers['x-access-token'];
+	if(!token) return res.status(401).send({ auth: false, message: 'No token provided.' });
+	jwt.verify(token, config.masterAuthSecret, function(err, decoded) {
+		if(err) return res.status(500).send({ auth: false, message: 'Failed to authenticate token.' });
```
psihius: I don't want to be an ass, but I honestly always require full block in any code review I do (and it's a forced transformation setting in my IDE)
```
if (condition) {
    // do stuff
}
```

Practice and experience show that non-block structure will get messed up or fucked with a 99% probability and will lead to bugs.

It also took me a while to parse it visually and see what is exactly executed in the condition.
Also, the line is just long :)

Could we please edit it?

master.js
```diff
@@ -440,37 +440,60 @@ app.post("/api/uploadMod", authenticate.middleware, function(req,res) {
...
-var sentItemStatisticsBySlaveID = {};
+/**
+POST endpoint for getting information about all our slaves, requires auth, responds including RCON passwords
+@memberof clusterioMaster
+@instance
+@alias /api/slaves
+*/
+app.post("/api/slaves", function(req, res) {
+	let token = req.headers['x-access-token'];
+	if(!token) return res.status(401).send({ auth: false, message: 'No token provided.' });
```
psihius: same as what I said bellow

PR #180, opened 2019-07-25
add compression to research sync

Godmave:


Danielv123:
See no reason not to have this enabled by default. Can I suggest ```needle.defaults({
  open_timeout: 60000,
  user_agent: 'MyApp/1.2.3',
  parse_response: false });``` and using that to somehow enable compression by default?

Godmave:
Would only save something if you have more than one request in a script, but should work.

needle.defaults({compressed:true})

most likely

Danielv123:
Actually, if you put that in the top of client.js it will apply to all plugins as well. require("needle") is cached and returns the same instance every time its used.

Godmave:
Aaaan, needle says "no". compressed is not a valid key for defaults ...

PR #184, opened 2019-07-27
Added "Doomsday" scenario

psihius:


Danielv123:
Err, should this be in the main repo, enabled by default? Not sure if everyone are going to want to play with this.

PR #190, opened 2019-08-02
Move node_modules/_app/ into lib/ as a package

Hornwitser:
Move the contents of node_modules/_app into lib/ as there shouldn't be
any content in node_modules.  This directory is automatically managed
and pruned by npm.

Add lib/ as a local package in package.json, this allows scripts
everywhere to require it's content as "lib/path" instead of relying on
local paths.  This shadows the StdLib Node.js Bindings package, I could
care less.

Change all requires with lib as a local path and _app to use just lib.

<hr>

This should fix the "Cannot find module '_app/fileOps'" error occurring due a change in npm that causes node_modules to be pruned by default.  But note that this change hasn't been tested any more than just ensuring that master.js and client.js runs at all.


Danielv123:
Didn't realize this was a thing, that looks great!

PR #191, opened 2019-08-04
Add tailing of files and move socketio connect after plugin load

Godmave:
Prevents losing of data between factorio and clusterio.
Prevents socket getting connections before all plugins are read to catch the connection event

Works fine on my little cluster, but should be reviewed and tested!


master.js
```diff
@@ -927,70 +924,64 @@ class wsSlave {
 		if(commandID) this.commandsWaitingForReturn[commandID] = {callback, timestamp: Date.now()};
 	}
 }
+
 io.on('connection', function (socket) {
-	// cleanup dead sockets from disconnected people
```
Danielv123: Everything else in this file uses tabs, was this change to spaces necessary?

Godmave: not intentional. just the ide doing I guess

Hornwitser:
I am strongly against pulling this entangled mess into the repository.
- You're modifying stuff all over the place that has absolutely nothing to do with implementing your feature.  This is guaranteed to give me headaches with my refactoring.
- You're doing back and forth merges that's littering the commit history with nonsense like `Merge branch 'master' into master` and `Merge pull request #11 from Danielv123/master` the latter even breaks GitHub.  Use feature branches, and if you have merge conflicts or need features added on the master, rebase your feature branch on top of it.  Also keep in mind that GitHub's web interface is completely unusable for the simple task of updating a remote tracking branch.
- Your commit messages are barbaric, the gold standard is a less that 50 character summary followed by an empty line, and then an extended description that's line wrapped at 72 characters with paragraphs separated by a blank line (see f76c84c).  While I don't care too much about long descriptions or that the summary is less than 50 characters, not having any summary and writing on the second line of the commit message is a sin.

PR #192, opened 2019-08-05
Prune unneeded runtime dependencies

Hornwitser:
Slim down on the runtime's near 1k npm depenedncies:
- Replace app-root-path, cache-require-paths, hash-files, and underscore with trivial code.
- Remove the unused dependencies bcryptjs, and right-pad.
- Remove snyk as a dependency, while I get that security is important npm already has vulnerability reporting builtin.  And I can't be bothered to register on yet another site only for it to send me irelevant security notices about dependencies not being used having issuse.
- Remove documentation as a dependency.  This was a two part process as node-factorio-api was also pulling it in for no reason.  I don't fancy the documentation module as it's overkill for our needs and pulls in 475 other dependencies.
- Move html5-validator to devDependencies.
- Recommend using the `--only=production` flag to `npm install`

In total this removes 625 dependencies from the runtime requirement of clusterio, bringing down the total from 920 to 295 dependencies.


psihius:
So far loos good to me, I assume this has been tested with a few major plugins like playerManager and teleportingTrains? What about a few other plugins in major use?

Hornwitser:
I have casually tested this with the playerManager and trainTeleports and so far it appears to work fine.  Would probably be better if the automated testing actually worked.  (I have a branch that fixes it, but it's based on this one so I'll post a pull request for it once this is merged.)

psihius:
@Hornwitser  Can't you just post a PR for PR? :D It does not really matter into what branch the PR is being merged into.

Hornwitser:
Git commits are intrinsically tied to _all_ of its ancestors.  That means if I want to post a PR for my fix-the-tests branch which is based on this branch it _must_ pull in all the commits of this branch into that PR.  Which is obviously not something you want.  I could rebase my fix-the-tests branch on top of master so it doesn't pull in this branch, but that creates a merge conflict with this branch so that just seems like pointless busywork to me.

Any reason why you can't get this merged in anytime soon?

psihius:
I just want @Danielv123 and @Godmave to take a look - I pinged them as reviewers, but they take their sweet time :)

psihius:
Well...

YOLO

PR #195, opened 2019-08-07
Fix the coverage and test scripts

Hornwitser:
Replace istanbul with nyc as istanbul has been deprecated is no longer developed and disable the remoteMap tests as it's missing the linvodb3 dependency.  I guess linvodb3 could be added, but since remoteMap is very much an not working in-dev thing I haven't investigated it, and I sort of wanted to keep the unneeded dependencies down.


Danielv123:
linvodb3 was removed and is not coming back. The version of remoteMap in the repo is old and should be deleted - the new version (v3) is a full rewrite.

Hornwitser:
I added .nyc_output to the .gitignore now.

PR #204, opened 2019-08-20
Refactor master to fix unit tests

Hornwitser:
This is a minor refactor of master.js and master.spec.js in order to fix the unit tests for it.

### DIFF
--- a/master.js
+++ b/master.js
@@ -1012,7 +1012,7 @@ async function getPlugins(){
 }
 
 
-masterPlugins = await getPlugins();
+var masterPlugins = await getPlugins();
 
 
 io.attach(server);
--- a/master.js
+++ b/master.js
@@ -14,15 +14,6 @@ node master.js
 // const updater = require("./updater.js");
 // updater.update().then(console.log);
 
-(async ()=>{
-
-// Set the process title, shows up as the title of the CMD window on windows
-// and as the process name in ps/top on linux.
-process.title = "clusterioMaster";
-
-// add better stack traces on promise rejection
-process.on('unhandledRejection', r => console.log(r));
-
 // configgy stuff
 debug = false;
 
@@ -54,24 +45,6 @@ const getFactorioLocale = require("lib/getFactorioLocale");
 
 // homemade express middleware for token auth
 const authenticate = require("lib/authenticate")(config);
-
-/** Sync */
-function randomStringAsBase64Url(size) {
-  return base64url(crypto.randomBytes(size));
-}
-if (!fs.existsSync("secret-api-token.txt")) {
-	config.masterAuthSecret = randomStringAsBase64Url(256);
-	fs.writeFileSync("config.json",JSON.stringify(config, null, 4));
-	fs.writeFileSync("secret-api-token.txt", jwt.sign({ id: "api" }, config.masterAuthSecret, {
-		expiresIn: 86400*365 // expires in 1 year
-	}));
-	console.log("Generated new master authentication private key!");
-	process.exit(0);
-}
-// write an auth token to file
-fs.writeFileSync("secret-api-token.txt", jwt.sign({ id: "api" }, config.masterAuthSecret, {
-	expiresIn: 86400*365 // expires in 1 year
-}));
 
 const express = require("express");
 const compression = require('compression');
@@ -174,9 +147,6 @@ const prometheusMasterInventoryGauge = new Prometheus.Gauge({
 	help: 'Amount of items stored on master',
 	labelNames: ["itemName"],
 });
-setInterval(()=>{
-	fs.writeFileSync("database/items.json", JSON.stringify(db.items));
-},config.autosaveInterval || 60000);
 /**
 GET Prometheus metrics endpoint. Returns performance and usage metrics in a prometheus readable format.
 @memberof clusterioMaster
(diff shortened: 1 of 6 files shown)


master.js
```diff
@@ -1007,15 +962,74 @@ async function getPlugins(){
...
+
+	if (httpsPort) {
+		var certificate = fs.readFileSync(path.resolve(config.databaseDirectory, 'certificates/cert.crt'));
+		var privateKey = fs.readFileSync(path.resolve(config.databaseDirectory, 'certificates/cert.key'));
+
+		let httpsServer = require("https").createServer({
+			key: privateKey,
+			cert: certificate
+		}, app)
+		httpsServer.listen(args.sslPort || config.sslPort);
```
TheAIBot: You probably meant to use httpsPort here.

master.js
```diff
@@ -1007,15 +962,74 @@ async function getPlugins(){
...
+		fs.writeFileSync("database/items.json", JSON.stringify(db.items));
+	},config.autosaveInterval || 60000);
+	process.on('SIGINT', shutdown); // ctrl + c
+	process.on('SIGHUP', shutdown); // terminal closed
+
+	// Make sure we're actually going to listen on a port
+	let httpPort = args.masterPort || config.masterPort;
+	let httpsPort = args.sslPort || config.sslPort;
+	if (!httpPort && !httpsPort) {
+		console.log("Error: no port configured to listen to");
```
TheAIBot: This error message should be more descriptive of the error and use console.error.
A better error message would be something along the lines of "Missing port for both http and https where at least one is required. Either specify them through the arguments --masterPort=<port> and --sslPort=<port> or in the config with the names masterport and sslPort."

master.js
```diff
@@ -1007,15 +962,74 @@ async function getPlugins(){
...
+		process.exit(1);
```
TheAIBot: Don't call `process.exit` right after writing to either `console.log` or `console.error` [as that may prevent the message from being sent to the console](https://github.com/nodejs/node/blob/master/doc/api/process.md#a-note-on-process-io). The [solution](https://nodejs.org/api/process.html#process_process_exit_code) is to set `process.exitCode = 1;` and return from the function so the program ends naturally.

Hornwitser: This is the convention in 4 other places.  Setting exit code and waiting for it to die doesn't work as there's a minefield of interval and timed callbacks everywhere, so node doesn't exit.

master.js
Hornwitser: I'm thinking more along the lines of "at least one of httpPort and sslPort must be configured".  Specifying config on the command line is something I wonder if it should just be axed.  You cannot modify options specified on the command-line through the planned management web interface.

master.js
Hornwitser: Right.

master.js
psihius: @Hornwitser well, the thing is that we probably should rewrite stuff to do proper exiting have timeouts or/and intervals exit properly. Existing like this can lose data or get into an inconsistent state.
So I do agree for this PR this is okay, but we probably should have an issue for a refactoring.

master.js
```diff
@@ -1007,15 +962,74 @@ async function getPlugins(){
...
 
-masterPlugins = await getPlugins();
+	/** Sync */
+	function randomStringAsBase64Url(size) {
+	  return base64url(crypto.randomBytes(size));
+	}
+
+	if (!fs.existsSync("secret-api-token.txt")) {
+		config.masterAuthSecret = randomStringAsBase64Url(256);
+		fs.writeFileSync("config.json",JSON.stringify(config, null, 4));
```
psihius: Had a thought while reviewing - we have all the file names hardcoded in the code right now with no ability to pass config file name when starting the process and all other file paths can actually be defined in the config file - that way you can use the same install with multiple files. But more importantly - no hard coded file names in the code all over the place.
Thoughts?

master.js
```diff
@@ -1007,15 +962,74 @@ async function getPlugins(){
...
+		fs.writeFileSync("secret-api-token.txt", jwt.sign({ id: "api" }, config.masterAuthSecret, {
```
psihius: Like this example - this file name is used in multiple places in the code and cannot be changed unless you edit the code.

master.js
```diff
@@ -1007,15 +962,74 @@ async function getPlugins(){
...
+			expiresIn: 86400*365 // expires in 1 year
+		}));
+		console.log("Generated new master authentication private key!");
+		process.exit(0);
+	}
+	// write an auth token to file
+	fs.writeFileSync("secret-api-token.txt", jwt.sign({ id: "api" }, config.masterAuthSecret, {
+		expiresIn: 86400*365 // expires in 1 year
```
psihius: Maybe expire time should also be a config option?

master.js
```diff
@@ -1007,15 +962,74 @@ async function getPlugins(){
...
-io.attach(server);
+	// Only start listening for connections after all plugins have loaded
+	if (httpPort) {
+		let httpServer = require("http").Server(app);
+		httpServer.listen(httpPort);
+		console.log("Listening for HTTP on port %s...", httpServer.address().port);
+	}
+
+	if (httpsPort) {
+		var certificate = fs.readFileSync(path.resolve(config.databaseDirectory, 'certificates/cert.crt'));
```
psihius: This can also go into config and use, just have a default value in it.

master.js
Hornwitser: I think secret-api-token.txt is kind of a weird way to do it.  Would much rather have a command line interface to generate a token of desired time length.  But also this is kind of out of scope for this PR.

master.js
Hornwitser: This already has a config option called sslCert, and I already have a fix this.  See the "Use sslCert and sslPrivKey config entries" commit of my config-rewrite branch.

master.js
Hornwitser: Another option would be to allow you to generate a token on the web interface.  Or maybe a whole slave config file.  But again.  Out of scope for this PR.  Open an issue about it.

PR #218, opened 2019-08-29
Remove remanent remoteMap code

Hornwitser:
Remove remoteMap specific code left behind after the removal of the remoteMap plugin in #197.  The plugin interface should be versatile enough to allow all of this code to reside inside the plugin itself.


psihius:
That's a lot of plugin code in master.js file nuked. Much wow. Merging.

PR #219, opened 2019-09-01
Add development warning

Hornwitser:
Add notice warning that this is the development branch for 2.0, along with an improvement of startup error reporting to master.js.

Marked as draft as it's referencing an non-existent branch 1.2.x.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,13 @@
 
 Discord for development/support/play: https://discord.gg/5XuDkje
 
+## Important notice
+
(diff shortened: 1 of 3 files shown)


Hornwitser:
This PR conflicts with the current master after 7d61cd9 added a different warning to the README.md.  I also think the wording of that could be better.  The 2.0 branch should be stable enough that it runs but that it's very much an alpha right now.

psihius:
I'm merging, we can adjust the wording separately with a documentation update

PR #221, opened 2019-09-03
Fix master socket.io broken by a3fc1dd

Hornwitser:
Add missing attach calls to the master socket.io object that was accidentally removed by the refactoring in #204 by a3fc1dd.  Unfortunately I've been unable to determine if socket.io actually supports listening on multiple servers at the same time, this needs further investigation.

### DIFF
--- a/master.js
+++ b/master.js
@@ -945,6 +945,7 @@ async function startServer() {
 	if (httpPort) {
 		let httpServer = require("http").Server(app);
 		await listen(httpServer, httpPort);
+		io.attach(httpServer);
 		console.log("Listening for HTTP on port %s...", httpServer.address().port);
 	}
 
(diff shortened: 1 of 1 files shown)


psihius:
@Hornwitser this has the answer: https://stackoverflow.com/questions/44277101/socket-io-listening-on-multiple-ports

Hornwitser:
No it does not.  That question is whether you can have two separate socket.io servers in the same node program (which you can, make two different instances of the io server and attach them to two different http server instances).  My question is whether you can attach the same socket.io server to two separate http(s) server instances at the same time.

Oh and btw, answer no.2 does not answer the question and lists code for making one port set with a command line parameter.  And answer no.3 just lists code that doesn't even use socket.io, instead it makes an express server supposedly listen on two different ports.

psihius:
Well, pretty much it listens on one port - multiple ports you need to use 2 socket.io sockets, which, actually, makes sense.
Basically I say just use SSL port if it is defined instead of HTTP. The only way you can use both is to have a reverse-proxy in front to terminate traffic and then send it to the single WebSocket port.

Hornwitser:
Well, here's the thing, attaching the socket.io server to multiple http(s) servers works as far as I have been able to tell.  But I don't know if it _should_ work, or if there are issues with it.

BTW: a socket.io socket is not the same as a socket.io server.

psihius:
@Hornwitser conflict needs resolving :)

Hornwitser:
The conflict has been solved.

psihius:
Well, let's deal with it as things go further. Merging.

Issue #200, opened 2019-08-09
[Design Proposal] New configuration system


Hornwitser:
The current way configuration is handled is to put it mildly terrible.  Everything is dumped into config.json, which is loaded at startup and never touched again.  This presents a lot of problems:

 - Plugin config is not contained, and there's no mechanism for a plugin to introduce a config entry into the config.json file.  Causing some plugins to make their own config file.
 - The loaded config is passed around, but some places wrongly require the config.json directly with a relative path
 - There's no clear distinction between what config is for the master, what is for slaves/client/instances, what is for factorio, what is or client plugins and what is for master plugins.
 - Each node in a cluster with multiple physical servers end up having to replicate a lot of the config manually
 - There's no mechanism implemented to write back a modified config at runtime, so a web based config is not possible.
 - There's no mechanism to make tests source a different config.

Thefore I propose that the whole configuration system is rewritten to have a lib/configManager module that manages all of the configuration needed by instances, slaves, the master server, and plugins.  And that is should have at least the following features:

 - Per plugin configuration.  This means that plugin a and b have separate spaces for their configs and do not see the values of each other's configs in their own.
 - Per instance configuration so that you can set for example a different password for different servers.
 - Base configs for instances and client plugins that applies to all slaves that can be overridden on a per slave/instance basis
 - A factorio_config entry that's passed directly to Factorio to allow any option for Factorio to be set, not just the hard coded ones.
 - Option to store the slave/instance config on the master server instead of locally enabling the config for the whole cluster to be managed at one spot.
 - Writing back the config on shutdown.

Things that would be nice extensions to add to this system, but is a bit broad in scope for the rewrite:

 - Events for validating and applying changes that plugins can respond to.
 - A JSON schema for the config (that plugins can extend) that can be validated with for example `ajv`
 - Web interface for the config

<hr>

I will come up with a more concrete plan for implementation later if this is accepted.

psihius:
This just makes too much sense ;)

Danielv123:
This makes a lot of sense, although you got a few things wrong about the current system.

* Adding a master web interface for modifying the config at runtime is possible, although any modules that load the config for themselves break it (see serverManager, it has a thing for this)

* There is a command line parameter for the client to use a different config (--config I think, might be for the master as well)

Everything else is about right. We need a new more robust config system if we are to properly decouple plugins from the core repo.

[...]

PR #222, opened 2019-09-03
Config Rewrite Part 1

Hornwitser:
These are preparations that are needed in order to make the new config system as proposed in #200 to work with the master server.  They mostly revolve around removing the dependency of the config in the master server until after `startServer` is invoked.


lib/authenticate.js
```diff
@@ -1,28 +1,48 @@
...
-					} else {
-						resolve({ok:true, msg:"Successfully authenticated"});
-					}
-				});
-			});
-		}
+function middleware(req, res, next) {
+	// This should not happen
+	if (!authSecret) {
+		return res.status(500).send({ message: 'Secret not set' });
```
TheAIBot: To be consistent with the other error paths, the body of the send should be 

`{ auth: false, message: 'Secret not set.', endpoint: req.route.path}`

Hornwitser: Fixed.

lib/database.js
```diff
@@ -0,0 +1,142 @@
...
+ * specified by items.
+ *
+ * @param {string} name - The name of the item to add.
+ * @param {number} count - The count of item to remove.
+ * @param {Map} items - The items database to add to, this is usually
+ *                      db.items.
+ */
+function addItem(name, count, items) {
+	let existingCount = items.get(name);
+	if (typeof existingCount === "number" && !isNaN(existingCount)) {
```
TheAIBot: We are really doing this backwards. Instead of checking if the db contains `NaN`, we should check whether we try to put `NaN` into the db.

Before inserting into the db, check if `existingCount` is `NaN` and if it's then don't insert into the db and instead write a warning to the console with `console.warn`.

lib/database.js
```diff
@@ -0,0 +1,142 @@
...
+ * stored item count will become negative.
+ *
+ * @param {string} name - The name of the item to remove.
+ * @param {number} count - The count of items to remove.
+ * @param {Map} items - The items database to remove from, this is
+ *                      usually db.items.
+ */
+function removeItem(name, count, items) {
+	let existingCount = items.get(name);
+	if (typeof existingCount === "number" && !isNaN(existingCount)) {
```
TheAIBot: Same as the comment for `addItem`.

lib/manager/pluginManager.js
```diff
@@ -3,114 +3,123 @@ const path = require("path");
...
+	plugins.forEach(plug => {
+		console.log(`${plug.name}`.padEnd(22).substr(0,22)+` - ${plug.description}`.substr(0,80-22));
+	});
+}
+
+async function getPlugins() {
+	let directories = await fs.readdir("./sharedPlugins");
+	let returnArray = [];
+	for(let i in directories){
+		let dir = directories[i];
```
TheAIBot: This seems redundant. `i` and `dir` are the same.

lib/manager/pluginManager.js
```diff
@@ -3,114 +3,123 @@ const path = require("path");
...
-					await fs.remove(plug.pluginPath);
-					return {
-						ok:true,
-						msg:`Deleted plugin ${plug.name}`
-					}
-				}
+
+			// check if plugin is enabled
+			try{
+				await fs.stat(path.join(pluginConfig.pluginPath, "DISABLED"))
```
TheAIBot: I know you didn't write this code but we might as well fix it now.
`fs.stat` [isn't made to check if a file exists](https://nodejs.org/api/fs.html#fs_fs_stat_path_options_callback). Use `fs.access` instead.

lib/manager/pluginManager.js
```diff
@@ -3,114 +3,123 @@ const path = require("path");
...
+			}catch(e){
+				pluginConfig.enabled = true;
 			}
+			returnArray.push(pluginConfig);
+		}
+	}
+	return returnArray;
+}
+
+function addPlugin(path) {
```
TheAIBot: This method should to be async as it returns a promise.

Hornwitser:
I couldn't get myself to [use the existing code](https://github.com/clusterio/factorioClusterio/blob/ef66ab37dae3052cca13346501080e5c6e36dc68/master.js#L177-L219) and that's why it was rewritten.  But I'm wondering now if `db.items` should be an instance of a `ItemDatabase` class defined in lib/database.  And then have `addItems` and `removeItems` be methods on it, as well as a load and save database method.

lib/authenticate.js
```diff
@@ -1,28 +1,48 @@
...
+
+	jwt.verify(token, authSecret, function(err, decoded) {
+		if (err) {
+			return res.status(401).send({ auth: false, message: 'Failed to authenticate token.', endpoint: req.route.path });
+		}
+		next();
+	});
+}
+
+function check(token){
```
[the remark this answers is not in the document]
Hornwitser: How odd that I missed that.

lib/database.js
Hornwitser: I was merely re-implementing the existing logic (which was broken too.)  There are checks before the calls to addItem/removeItem, but I'm wondering if having them throw an Error on bad input would be a better idea.

lib/manager/pluginManager.js
Hornwitser: No.  `i` is the index in the array of directories and `dir` is the given directory.  This would be better written as `for (let dir of directories) {` though.

lib/database.js
```diff
@@ -0,0 +1,142 @@
+/**
+ * Database tools.
+ * @module lib/database
+ * @author Hornwitser
+ */
```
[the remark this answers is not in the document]
Hornwitser: The `@module` directive is required for jsdoc to even show the module.  Author is not redundant either, as it tells you who wrote the majority of the module.

lib/database.js
```diff
@@ -0,0 +1,142 @@
+/**
+ * Database tools.
```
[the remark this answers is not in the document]
Hornwitser: Indeed.  But I think I'm going to rewrite this to implement an item database.

lib/manager/pluginManager.js
Hornwitser: What's the difference between an async function returning a promise and a ordinary function returning a promise?  They both do the same thing.

lib/manager/pluginManager.js
Hornwitser: No.  The commit is called "Remove unused config parameter to pluginManager", not "fix everything wrong with pluginManager".  It's also a pointless fix as this method of disabling plugins will be removed by the config rewrite.

Hornwitser:
I have pushed a revised version of "Refactor database and item retreival" that makes the item database a separate class that keeps its integrity.  I do wonder if mapToObject in lib/database should belong in lib/objectOps instead.  The load and save map as json functions I'm also not sure about.

lib/manager/pluginManager.js
Danielv123: One returns a promise, the other returns a promise that immediately resolves to another promise. There might be a microoptimization, but we shouldn't care about that.

lib/database.js
```diff
@@ -0,0 +1,230 @@
...
+	 *     database from.  An empty database will be created if this parameter
+	 *     is left undefined.
+	 */
+	constructor(serialized) {
+		this._items = new Map();
+
+		// Verify the content of the serialized database
+		if (serialized !== undefined) {
+			for (let [name, count] of Object.entries(serialized)) {
+				if (typeof count !== "number" || isNaN(count)) {
```
TheAIBot: Can be replaced wit ha call to the new `checkNameAndCount`

lib/database.js
```diff
@@ -0,0 +1,230 @@
...
+	 * Removes count of an item from the database.
+	 *
+	 * Remove count copies of the item specified by name from the item database
+	 * If count is greater than the currently stored count the stored item
+	 * count will become negative.
+	 *
+	 * @param {string} name - The name of the item to remove.
+	 * @param {number} count - The count of items to remove.
+	 */
+	removeItem(name, count, items) {
```
TheAIBot: Forgot to remove the `items` argument.

lib/database.js
Hornwitser: I decided to split it into checkName and checkCount as that allows checName to be using is getItemCount as well.

routes/api/remove.js
```diff
@@ -139,32 +129,33 @@ function doleDivider({
     const originalCount = Number(object.count) || 0;
     object.count /= ((_doleDivisionFactor[object.name]||0)+doleDivisionRetardation)/doleDivisionRetardation;
     object.count = Math.round(object.count);
-    if(item.length > 40) console.info(`Serving ${object.count}/${originalCount} ${object.name} from ${item} ${object.name} with dole division factor ${(_doleDivisionFactor[object.name]||0)} (real=${((_doleDivisionFactor[object.name]||0)+doleDivisionRetardation)/doleDivisionRetardation}), item is ${Number(item) > Number(object.count)?'stocked':'short'}.`);
-    
-    // Update existing items if item name already exists
-    if(Number(item) > Number(object.count)) {
+	if (itemCount > 40) {
+		console.info(`Serving ${object.count}/${originalCount} ${object.name} from ${itemCount} ${object.name} with dole division factor ${(_doleDivisionFactor[object.name]||0)} (real=${((_doleDivisionFactor[object.name]||0)+doleDivisionRetardation)/doleDivisionRetardation}), item is ${itemCount > Number(object.count)?'stocked':'short'}.`);
```
TheAIBot: `object.count` is already a number.

routes/api/remove.js
```diff
@@ -139,32 +129,33 @@ function doleDivider({
...
+	}
+
+	// Update existing items if item name already exists
+	if (itemCount > Number(object.count)) {
```
TheAIBot: `object.count` is already a number.

routes/api/remove.js
Hornwitser: Is it though?  Doesn't object come from the request body of `POST /api/remove`?

TheAIBot: Even if that's the case, you are setting it `object.count = Math.round(object.count);` a little higher up.

PR #227, opened 2019-09-09
Config Rewrite Part 2

Hornwitser:
These are more preparations for the config rewrite.  Some notes:
- The change from masterIP+masterPort to masterURL is not strictly necessary, but it fixes the problem with http being hardcoded so that you couldn't make the slave connect to an https port, and it's simpler to do before the new config system is introduced.
- Since the change from masterIP+masterPort to masterURL sort of breaks lib/Model I decided to remove it.  This module is compiled TypeScript from https://github.com/veladon/clusterioModel, any fixing should be with those TypeScript sources, not the compiled output.  I also don't think this stand alone tool belongs in the factorioClusterio repository as it's specific to the trainTeleports plugin.
- The client.js refactor here is only the beginning of what's needed to move the configuration loading to after the plugins have been discovered.


client.js
```diff
@@ -563,9 +537,61 @@ write-data=${ path.resolve(config.instanceDirectory, instance) }\r\n
...
+	if(!config.masterAuthToken || typeof config.masterAuthToken !== "string"){
+		console.error("ERROR invalid config!");
+		console.error("Master server now needs an access token for write operations. As clusterio slaves depends \
+		upon this, please add your token to config.json in the field named masterAuthToken. \
+		You can retrieve your auth token from the master in secret-api-token.txt after running it once.");
+	}
+
+
+	// handle commandline parameters
+	if (!command || command == "help" || command == "--help") {
```
psihius: I think this is better handled by a switch-case and, maybe, can be extracted into a function

config.json.dist
```diff
@@ -7,8 +7,8 @@
 	"__comment1": "IP players should connect to to join your servers, should be portforwarded and stuff",
 	"publicIP": "localhost",
 	
-	"__comment2": "IP the master web server runs at, can be ignored on the master system",
-	"masterIP": "localhost",
+	"__comment2": "URL the master web server runs at, can be ignored on the master system.  Must not end with a slash",
+	"masterURL": "http://localhost:8080",
```
psihius: Hm, maybe we should not add port here, cause we have the "masterPort" setting and you need to change both - that can be error-prone and a bit annoying.
Maybe in the code, we should just combine config.masterURL = config.masterURL + config.masterPort.toString()` or use variables like `masterHost` and `masterPort` and combine them into `masterURL` ?

sharedPlugins/clusterioMod/index.js
```diff
@@ -78,7 +78,7 @@ module.exports = class remoteCommands {
 							}
 						}
 						console.log("Recorded flows, copper plate since last time: " + payload["copper-plate"]);
-						needle.post(that.config.masterIP + ":" + that.config.masterPort + '/api/logStats', {timestamp: timestamp, instanceID: that.config.unique, data: payload}, needleOptionsWithTokenAuthHeader, function (err, response, body) {
+						needle.post(that.config.masterURL + '/api/logStats', {timestamp: timestamp, instanceID: that.config.unique, data: payload}, needleOptionsWithTokenAuthHeader, function (err, response, body) {
```
psihius: This is a very long line. Need to chop it into multiple lines.

sharedPlugins/UPSdisplay/index.js
```diff
@@ -25,7 +25,7 @@ class UPSdisplay {
 				let UPS = Math.round(ticksInPeriod / (timePeriod/1000));
 				// console.log("UPS: " + UPS);
 				try{
-					needle.post(this.config.masterIP+':'+this.config.masterPort+'/api/editSlaveMeta', {instanceID: this.config.unique, password: this.config.clientPassword, meta: {UPS:UPS, tick:this.tick}}, {headers: {'x-access-token': this.config.masterAuthToken}}, function(err, resp) {
+					needle.post(this.config.masterURL+'/api/editSlaveMeta', {instanceID: this.config.unique, password: this.config.clientPassword, meta: {UPS:UPS, tick:this.tick}}, {headers: {'x-access-token': this.config.masterAuthToken}}, function(err, resp) {
```
psihius: This is a very long line. Need to chop it into multiple lines.

sharedPlugins/clusterioMod/index.js
```diff
@@ -168,7 +168,7 @@ module.exports = class remoteCommands {
 				// request our items, one item at a time
 				for (let i = 0; i < Object.keys(preparedPackage).length; i++) {
 					// console.log(preparedPackage[Object.keys(preparedPackage)[i]]);
-					needle.post(that.config.masterIP + ":" + that.config.masterPort + '/api/remove', preparedPackage[Object.keys(preparedPackage)[i]], needleOptionsWithTokenAuthHeader, function (err, response, body) {
+					needle.post(that.config.masterURL + '/api/remove', preparedPackage[Object.keys(preparedPackage)[i]], needleOptionsWithTokenAuthHeader, function (err, response, body) {
```
psihius: This is a very long line. Need to chop it into multiple lines.

sharedPlugins/inventoryImports/interfacingFunctions.js
```diff
@@ -60,7 +60,7 @@ function handleInventory(json, config, callback){
 						}
 						//console.log("Requesting: "+JSON.stringify(request));
 						// request the difference from master
-						needle.post(config.masterIP + ":" + config.masterPort + '/api/remove', request, {json:true,headers: {'x-access-token': config.masterAuthToken}}, function (err, response, body) {
+						needle.post(config.masterURL + '/api/remove', request, {json:true,headers: {'x-access-token': config.masterAuthToken}}, function (err, response, body) {
```
psihius: This is a very long line. Need to chop it into multiple lines.

sharedPlugins/inventoryImports/interfacingFunctions.js
```diff
@@ -87,7 +87,7 @@ function handleInventory(json, config, callback){
 				for(let i = 0; i < stackArray.length; i++){
 					if(stackArray[i].count > 0){
 						// console.log("Returning overflow: " + JSON.stringify(stackArray[i]));
-						needle.post(config.masterIP + ":" + config.masterPort + '/api/place', stackArray[i], {headers: {'x-access-token': config.masterAuthToken}}, function (err, resp, body) {
+						needle.post(config.masterURL + '/api/place', stackArray[i], {headers: {'x-access-token': config.masterAuthToken}}, function (err, resp, body) {
```
psihius: This is a very long line. Need to chop it into multiple lines.

config.json.dist
Hornwitser: masterPort is only for the master server and specifies what port to listen on, masterURL is only for slaves and specifies where to connect.  It should be really clear after the config a split into two.

I don't think we should do any combining of host and port into the url because that limits your option on how to host the master.  With a url you can specify for example `https://example.com/master` (notice how port isn't needed when running on the default one) and proxy that to the master server with for example apache or nginx as the front-end.  This could also be used to serve all the files that doesn't need any template logic once those have been separated out.

client.js
Hornwitser: For this round of refactoring I'm going to say no.  There are far better ways to do this than switches and if else trees.

PR #228, opened 2019-09-09
Fixes editSlaveMeta and db AddItem errors

AlienXAXS:
While trying out the latest version of Clustorio, i found the following errors causing the mod to not work correctly.

### DIFF
--- a/lib/database.js
+++ b/lib/database.js
@@ -190,7 +190,7 @@ class ItemDatabase {
 	 * @param {number} count - The count of item to remove.
 	 */
 	addItem(name, countInput) {
-		var count = Number(count);
+		var count = Number(countInput);
 		checkName(name);
 		checkCount(count);
 
--- a/lib/database.js
+++ b/lib/database.js
@@ -189,7 +189,8 @@ class ItemDatabase {
 	 * @param {string} name - The name of the item to add.
 	 * @param {number} count - The count of item to remove.
 	 */
-	addItem(name, count) {
+	addItem(name, countInput) {
+		var count = Number(count);
 		checkName(name);
 		checkCount(count);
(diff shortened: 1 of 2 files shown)


lib/database.js
```diff
@@ -189,7 +189,8 @@ class ItemDatabase {
 	 * @param {string} name - The name of the item to add.
 	 * @param {number} count - The count of item to remove.
 	 */
-	addItem(name, count) {
+	addItem(name, countInput) {
+		var count = Number(countInput);
```
Hornwitser: This is not right.  addItem is supposed to throw an error if you pass it something that's not a number.  Have you got a stack trace for when a non-number was passed to it?

TheAIBot: I looked into why count isn't a number and it turns out it was never converted to a number when it's sent from the client.
https://github.com/clusterio/factorioClusterio/blob/master/sharedPlugins/clusterioMod/index.js#L115 You need to change that line to `count: Number(g[1]),` so the client sends the count as a number.

AlienXAXS: > This is not right. addItem is supposed to throw an error if you pass it something that's not a number. Have you got a stack trace for when a non-number was passed to it?

It simply happend every time I put something in the sender object in Factorio.

Then the fix should be here instead:
Within /api/place (app.post), req.body.count is a string at this point.
db.items.addItem(req.body.name, req.body.count);

psihius: It is the correct way of doing type conversion at the calling side, not inside the function. If anything, function should check if the incoming parameter is a number.

So yeah, please update the addItem call, not the addItem function itself.

Hornwitser:
Looks good to me.

PR #230, opened 2019-09-10
Add authentication to socket.io server

Hornwitser:
Authenticate connections made to the socket.io server via a token query parameter that's validated in a similar manner to the x-access-token for HTTP requests.  This is based on the [With query parameters](https://socket.io/docs/client-api/#With-query-parameters) example to socket.io.

Note that this isn't the best method of doing this as there's a risk the URL with the token in stored in web server logs.

### DIFF
--- a/client.js
+++ b/client.js
@@ -570,7 +570,10 @@ async function instanceManagement(instanceconfig) {
     console.log("Started instanceManagement();");
 
     /* Open websocket connection to master */
-	var socket = ioClient("http://"+config.masterIP+":"+config.masterPort);
+	var socket = ioClient("http://"+config.masterIP+":"+config.masterPort+"?token="+config.masterAuthToken);
+	socket.on("error", err => {
+		console.log("SOCKET | Error: ", err);
+	});
 	socket.on("hello", data => {
 		console.log("SOCKET | registering slave!");
 		socket.emit("registerSlave", {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+Changelog
+=========
+
+Version 1.2.1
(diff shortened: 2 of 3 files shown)


client.js
```diff
@@ -570,7 +570,10 @@ async function instanceManagement(instanceconfig) {
     console.log("Started instanceManagement();");
 
     /* Open websocket connection to master */
-	var socket = ioClient("http://"+config.masterIP+":"+config.masterPort);
+	var socket = ioClient("http://"+config.masterIP+":"+config.masterPort+"?token="+config.masterAuthToken);
```
psihius: Shouldn't this use the updated masterful already? We merged that config change already

Hornwitser: This PR is against the 1.2.x branch if you haven't noticed.

psihius:
Oh, sorry. But then this is a major break, we need a 1.3 release then. This type 9f fix should not be done in patch versions.

client.js
psihius: Oh, sorry. But then this is a major break, we need a 1.3 release then. This type of fix should not be done in patch versions.

Hornwitser:
The things that break with this change is that a previous version of the client cannot connect to the update master server, as it requires authentication.  Additionally any third party code that connects over the socket.io server also needs to be updated to pass authentication (I'm not aware of any such code.)  Plugins are not affected by this change.

If you consider this a major break then you need a 2.0 release if you are to follow semantic versioning.  I do not consider this a breaking change, as nowhere is it specified that an old version of the client should be able to connect to a newer version of the master server.

psihius:
@Hornwitser cherry-pick this into master too.

PR #238, opened 2019-09-18
Move EJS templates

Hornwitser:
This builds the idea that static content and templated pages should be separated.  The EJS templates are EJS and not HTML so I've changed the extension for them.  This makes them syntax highlight correctly in editors too.  For plugins the sharedPlugin directory is also added to the views path so that they can easily include their own as well as other plugin's templates and pages without having to resort to ugly relative path traversal.  This also enables them to use the Express.js builtin res.render() function instead of having to do the rending themselves.


psihius:
Let's see how it goes, looks okay

PR #239, opened 2019-09-18
Spawn factorio server deteached in client.js

### LINKED ISSUE #217: SIGTERM is sent twice on Linux
Sending ^C via terminal on Linux sends SIGTERM to the factorio process twice, causing it to exit without completing the saving.  I.e:

```
$ node client.js start foo
...
^CFact:    2.822 Received SIGINT, shutting down   2.822 Quitting: signal.
   2.822 Info MainLoop.cpp:400: Saving map as instances/foo/saves/_autosave1.zip

Caught interrupt signal, disconnecting RCON
Rcon disconnected, Sending ^C
Fact:    2.825 Received second SIGINT, terminating immediately
child process exited with code 1
```

The reason for this is that SIGINT sent via the terminal is sent by default to all members of the active _process group_ (not to be confused by the permission group the process runs as, that's something else entirely.)  So both client.js and factorio is receiving the SIGINT, then client.js sends another SIGINT to the factorio process.

It's still possible that only the client.js process is sent SIGTERM, process managers may be doing this instead of sending to the process group.

Fixing this might be a simple as [setting the detached option](https://nodejs.org/docs/latest-v10.x/api/child_process.html#child_process_options_detached) to `spawn`, but it needs testing.

Hornwitser:
Sending ^C via the terminal on Linux may cause SIGINT to be sent twice to the factorio process, causing it to exit without completing the saving.  This happens because the shell sends the interupt to all members of the process group for the job and this includes both the node client process and the factorio server process.

Avoid this situation by passing detached to spawn when the platform is Linux, which causes the process to be created with its own process group and therefore not targeted by a SIGINT from the shell.

This unfortunately does not work on Windows as we can't send SIGINT or ^C to a process there.  Doing `subprocces.kill()` on Windows terminates the process similarly to how SIGKILL works on Linux.

<hr>

Fixes #217

### DIFF
--- a/client.js
+++ b/client.js
@@ -359,7 +359,8 @@ write-data=${ path.resolve(config.instanceDirectory, instance) }\r\n
             '--rcon-port', Number(process.env.RCONPORT) || instconf.clientPort,
             '--rcon-password', instconf.clientPassword,
         ], {
-            'stdio': ['pipe', 'pipe', 'pipe']
+			'stdio': ['pipe', 'pipe', 'pipe'],
+			'detached': process.platform === "linux",
         }
     );
     factorio.stdout.on("data", data => {
@@ -471,7 +472,8 @@ write-data=${ path.resolve(config.instanceDirectory, instance) }\r\n
 				'--server-settings', instancedirectory + '/server-settings.json',
 				'--port', args.port || Number(process.env.FACTORIOPORT) || instanceconfig.factorioPort
 			], {
-				'stdio': ['pipe', 'pipe', 'pipe']
+				'stdio': ['pipe', 'pipe', 'pipe'],
+				'detached': process.platform === "linux",
 			}
 		);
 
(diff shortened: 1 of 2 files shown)


client.js
```diff
@@ -359,7 +359,8 @@ write-data=${ path.resolve(config.instanceDirectory, instance) }\r\n
             '--rcon-port', Number(process.env.RCONPORT) || instconf.clientPort,
             '--rcon-password', instconf.clientPassword,
         ], {
-            'stdio': ['pipe', 'pipe', 'pipe']
+            'stdio': ['pipe', 'pipe', 'pipe'],
+			'detached': process.platform === "linux",
```
psihius: Formatting issue?

Hornwitser: Nah, original code has spaces instead of tabs.  I suppose both lines could be changed to use tabs here.

Hornwitser:
I have forgotten to add an entry in the changelog for this fix.

client.js
psihius: Most of the codebase uses tabs - just reformat to tabs the parts y'r touching, we should get it to uniform state :)

Issue #201, opened 2019-08-09
[Design Proposal] Daemonize slave


Hornwitser:
I'm thinking it would be better if a slave would work more like a daemon that managed and ran multiple Factorio instances.  The advantage of this being that you could manage the instances better from the master server this way, having for example the option to implement a web panel on the master server that starts and stop individual instances.  Maybe a future feature based on this could be a button/api that moves an instance from one slave to another for ease of load balancing.

Danielv123:
I was working on this ages ago, that was the intention behind serverManager. I never got that far though. Main issues I ran into:

1. Client.js is monolithic and ugly. There is no good way of programatically starting a new slave through it. As we are working towards compartimentalization and a more testable codebase, solving this at the same time would make sense.

2. Thats about it I guess.

Godmave:
This would also take over some/all of the functionality of the bird panel. You would only set up the daemon which connects to the master and allows to spawn and manage instances from there. You could still have the whole thing inside a docker container for easy deployment.

psihius:
I actually have an idea since there have been some developments on the front of a management server thingie that can do all we need. See Discord :)

Hornwitser:
Please don't make vague references to discussions that has happened on Discord in issues.  It's impossible to find the discussion at a later point in time.  Discord lets you [link directly to the relevant message](https://discordapp.com/channels/450361298220351489/450361706426925066/614414110561665075) either use that or don't mention the discussion at all.

psihius:
https://discordapp.com/channels/450361298220351489/450361706426925066/614414110561665075

Hornwitser:
This was implemented in #261.

PR #242, opened 2019-10-02
Rewrite factorio server spawning and communication

Hornwitser:
This is a do over of the server spawning and communication logic, the idea is to get some structure on this instead of being a complete mess in client.js.  It's partly necessary for #200 as it proved to be problematic to adopt the new configuration system with all the clutter present.  It's also a hard requirement for #201.

I've not yet completed this, but I wanted to put this out in case anyone has some thoughts they would like to share on this.


client.js
```diff
@@ -326,34 +229,84 @@ async function downloadMod() {
 	}
 }
 
+/**
```
psihius: Question, maybe this should go into the lib folder as utility methods? So we don't have big fat client.js with all the code and something more manageable?

client.js
```diff
@@ -326,34 +229,84 @@ async function downloadMod() {
...
+				logger.warning(`Warning: ignoring file '${entry.name}' in sharedMods`);
+			}
+
+		} else {
+			logger.warning(`Warning: ignoring non-file '${entry.name}' in sharedMods`);
+		}
+	}
+}
+
 async function createInstance(config, args, instance) {
```
psihius: Same goes here. Also, this function is somewhat big, ideally you split it into smaller chunks and that also lends itself to go into a file of its own or something :)

client.js
```diff
@@ -326,34 +229,84 @@ async function downloadMod() {
...
+	console.log(`Creating ${instance.path()}`);
+	await fs.ensureDir(instance.path());
+	await fs.ensureDir(instance.path("script-output"));
+
+	let target = path.join("lib", "scenarios");
+	let link = instance.path("scenarios");
+	await fs.symlink(path.relative(path.dirname(link), target), link, 'junction');
+
+	await symlinkMods(instance, "sharedMods", console);
 	let instconf = {
```
psihius: Maybe use actual files with default values, so they are not defined in the code like this and loaded via require?

client.js
```diff
@@ -362,11 +315,16 @@ write-data=${ instance.path() }\r\n
...
+	await server.init();
+
+	let serverSettings = await server.exampleSettings();
 	let name = "Clusterio instance: " + instance.name;
 	if (config.username) {
 		name = config.username + "'s clusterio " + instance.name;
 	}
-	let serversettings = {
+
+	let overrides = {
```
psihius: Same here - load it from file?

client.js
```diff
@@ -375,65 +333,55 @@ write-data=${ instance.path() }\r\n
...
-			process.exit(0);
+
+	for (let [name, value] of Object.entries(overrides)) {
+		if (!Object.hasOwnProperty.call(serverSettings, name)) {
+			throw Error(`Expected server settings to have a ${name} property`);
 		}
+		serverSettings[name] = value;
+	}
+
+	await fs.writeFile(instance.path("server-settings.json"), JSON.stringify(serverSettings, null, 4));
```
psihius: filename should be a `const server_config = "server-settings.json";`  or something like that. We have been working on removing hard-coded file names from the codebase.

client.js
```diff
@@ -796,7 +680,7 @@ async function instanceManagement(config, instance, instanceconfig) {
 		setInterval(getID, 10000);
 		getID();
 		function getID() {
-			messageInterface("/silent-command rcon.print(#game.connected_players)", function(playerCount) {
+			server.sendRcon("/silent-command rcon.print(#game.connected_players)").then((playerCount) => {
```
psihius: But are we using only RCON going forward? stdin/stdout is a valid communication layer too and for some things probably is even better.

client.js
Hornwitser: Unless you have some grand architectural structure pre planned, I feel like it's the wrong time to be moving these functions around.

Hornwitser:
I've completed most of the code now.  Summary of changes:
- Refactored fileOps.getNewestFile as it was not compatible with the idea that you pass the name of the file as it appears in the saves directory.
- Renamed factorioPort to gamePort in the class.
- Removed some commented out code.
- Refactored getFactorioLocale and moved it to lib/factorio/locale
- Used the ini module to create the config.ini for Factorio
- Introduced a modular outputHeuristics filter which triggers actions based on outputs of Factorio.
- Changed the start and startScenario to wait until the server is ready to accept clients before returning.
- Added the disableAchievements method.
- Added integration tests.

Hornwitser:
I've completed this now.  Latest changes are that I've removed the deriving of instance id from the RCON password, removed stringUtils, and fixed some issues with shutdown and rcon on Windows.

Danielv123:
Your changelog is missing the fact that instanceconfig.unique is now deprecated in favour of instanceconfig.id - also, other places in the code its called instanceID, isn't it? Maybe consider changing that if it makes sense

Hornwitser:
I'm pretty sure the config rewrite is going to have change this again so keeping it as unique should be fine for now.  I didn't want to name it unique in the config file and calling it instanceconfig.instanceID seems a little redundant.

client.js
psihius: The thing is - you define the file names somewhere at the top of the file, preferably the root file and you just use that variable throughout the code. And it makes it easier to also add the ability to pass the file name as a command-line parameter. While we use instance-per-dir approach, you might want to use more than one server-settings.json file or even store them somewhere else and pass them at launch as paths.

PR #245, opened 2019-10-12
Ease installation

Hornwitser:
These are some targeted changes to make installation easier.  I tried installing clusterio on a clean install of Ubuntu Desktop 18.04 and the problems I ran into were -q missing from wget line making the sudo password prompt difficult to see, listening on 443 requiring elevated privileges (ssl is broken, no point having it enabled by default), pm2 being confusing and restarting the master over and over when there's an error at startup, and forgetting to put in the masterAuthToken which the client warns about but the message quickly drowns in other log output.

### DIFF
--- a/config.json.dist
+++ b/config.json.dist
@@ -14,7 +14,7 @@
 	"masterPort": 8080,
 	
 	"__commentsslPort": "HTTPS port for the master to listen to. Set to 0 or an empty string to not listen for HTTPS.",
-	"sslPort": 443,
+	"sslPort": 0,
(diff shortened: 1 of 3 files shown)


config.json.dist
```diff
@@ -14,7 +14,7 @@
 	"masterPort": 8080,
 	
 	"__commentsslPort": "HTTPS port for the master to listen to. Set to 0 or an empty string to not listen for HTTPS.",
-	"sslPort": 443,
+	"sslPort": 0,
```
Danielv123: Yep, this should probably have been done before.

PR #251, opened 2019-10-28
Improve handling of FactorioServer.stop()

Hornwitser:
Update rcon-client for the updated connection handling which waits for the connection to end when calling .end() on it and also wait for it connect if currently trying to connect in FactorioServer.stop().

And add hang detection to FactorioServer.stop(), should hopefully prevent the the issue observed by travis-ci.

### DIFF
--- a/lib/factorio/server.js
+++ b/lib/factorio/server.js
@@ -620,6 +620,13 @@ class FactorioServer extends events.EventEmitter {
 
 		this._state = "stopping";
 		if (this._rconClient) {
+
+			// If RCON is not yet fully connected that operation needs to
+			// complete before the RCON connection can be closed cleanly.
+			if (!this._rconReady) {
+				await events.once(this, 'rcon-ready');
+			}
+
 			await this._rconClient.end();
 		}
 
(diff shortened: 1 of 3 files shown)


lib/factorio/server.js
```diff
@@ -638,7 +647,33 @@ class FactorioServer extends events.EventEmitter {
...
+			let hanged = true;
+			function setAlive(output) { hanged = false; }
+			this.on('_quitting', setAlive);
+
+			let timeoutId = setTimeout(() => {
+				if (hanged) {
+					console.error("Factorio appears to have hanged, sending second SIGTERM");
+					this._server.kill();
+				}
+			}, 5000);
```
psihius: I don't think 5 seconds is really enough. Bigger maps can easily take much more to save, we have seen up to 20-25 seconds.

Hornwitser: This timeout is from sending SIGTERM until receiving the "Quitting: signal" message on stdout, which normally take no more that a millisecond.  Map saving happens after that so this code will wait indefinitely for saving to complete.

PR #252, opened 2019-11-09
Fix inventoryImports

Hornwitser:
Apparently this plugin hasn't worked for months.  I wonder if people are going to start complaining that this batshit op plugin is enabled by default.


sharedPlugins/inventoryImports/interfacingFunctions.js
```diff
@@ -141,7 +141,10 @@ function pollInventories(outputFile){
...
+
+	// Warning: This command is not in sync with lua commands.lua
+	return `/silent-command local a='{"players":{'local b=false;for c,d in pairs(game.players)do if d.connected and d.character then if b then a=a..','else b=true end;local e=game.players[c].get_inventory(defines.inventory.character_main).get_contents()local g={}for h=1,game.players[c].force.character_logistic_slot_count do g[h]=game.players[c].character.get_request_slot(h)end;a=a..'"'..c..'":{"inventory":{'local i=false;for j,k in pairs(e)do if i then a=a..','else i=true end;a=a..'"'..j..'":'..k end;a=a..'},"requestSlots":{'i=false;for j,k in pairs(g)do if i then a=a..','else i=true end;a=a..'"'..k['name']..'":'..k['count']end;a=a..'}}'end end;a=a.."}}"game.write_file("${outputFile}",a,true,0)`
```
psihius: Is backquote/backtick a proper string encasing symbol in Lua?

psihius: Okay, it's defining a template with the ability to insert variables for substitution. Works then I guess :)

Hornwitser: The backticks form a template literal.  It's a JavaScript string, like `"double quoted"` and `'single quoted'` with the difference being you can include newlines, and there's a substitution syntax of the form `${expression}`.  See [the MDN reference on them](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

PR #262, opened 2019-11-21
early config.uploadModsToMaster check

survaaki:
Move check to start of function for config.uploadModsToMaster. Write one log line instead of one log line per mod when disabled. No change to functionality when enabled.


Danielv123:
I should mention that this feature was recently removed from 2.0, but I see no reason not to merge this for the 1.2.x branch

Hornwitser:
You should add an entry to the changelog for this.

survaaki:
Entry to changelog has been added

PR #272, opened 2019-12-23
Updated NodeJs version to 11.13+ due to usage of Events.once 

IBlackI:
Using a version lower than 11.13 will crash due to usage of Events.once https://nodejs.org/api/events.html#events_events_once_emitter_name which has been added in node v11.13.0. Event.once is used in /lib/link/connectors.js:78

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Clusterio can also do a few other neat things, such as giving you access to epoc
 **Warning**: These instructions are for the unstable master version and is not
 recommended for use, see [the 1.2.x branch][1.2.x] for how to install the stable version.
 
-JS does not support EOL ubuntu releases. Make sure you are on the most recent LTS release or newer.
+NodeJS does not support EOL ubuntu releases. Make sure you are on the most recent LTS release or newer.
 
 Master and all slaves:
 
--- a/README.md
+++ b/README.md
@@ -101,11 +101,11 @@ Clusterio can also do a few other neat things, such as giving you access to epoc
 **Warning**: These instructions are for the unstable master version and is not
 recommended for use, see [the 1.2.x branch][1.2.x] for how to install the stable version.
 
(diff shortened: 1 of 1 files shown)


Hornwitser:
How does the events.once usage pass on Travis?  It's used all over the place, yet all the tests pass on Node.js v10.18.0.

IBlackI:
No idea. I was using node 10.14 and it crashed on the Events.once. The docs say it was introduced in 11.13 so that is why I updated the requirements to 11.13.

IBlackI:
Actually. They backported it. https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#2019-05-28-version-10160-dubnium-lts-bethgriggs

That explains why it is working on newer versions. But still a good idea to specify the version to use instead of only the major version.

Hornwitser:
The docs [also say it was added in v10.16.0](https://nodejs.org/docs/latest-v10.x/api/events.html#events_events_once_emitter_name).  Which was released in May this year.

psihius:
Well, what do you want to do here?
Bump the "engine" in package,json to `>=11.x` or bump the 10th version to `>=10.16`

Hornwitser:
Both?  `>=10.16 <11 || >=11.13` should work fine.  Though it doesn't actually do anything more than print a warning on npm install.

I was thinking of checking the version at startup, and refuse to run with an error if it's too low.

IBlackI:
No need to error out, it will crash anyway. As long as it is clearly
described in the requirements it is fine. But right now I was running a
version of node 10 (as docs said) which was too low, making it harder to
point to the version of node being the problem.

Op wo 25 dec. 2019 13:26 schreef Hornwitser <notifications@github.com>:

Hornwitser:
Implemented through #274 as it's my intention to keep supporting Node.js v10.16+.

PR #277, opened 2020-03-09
Initial config system rewrite

Hornwitser:
This is a scaled down version of the proposed configuration system in #200.  The main things that have not been implemented from the proposal is versioning, migrations, and basing a config on another config.  These things have turned out to be a major headache to implement and the utility of them seems small.  I might implement config bases if the need for it is there in the future.

The replacement for versioning and migration of configs is lenient handling of config fields.  So instead of trying to ensure that the config never ever gets an invalid value in it, invalid and missing values are replaced with the default value for that field.  I also made it so that unknown fields are preserved so that uninstalling a plugin doesn't magically erase all the config it had next time the server is started.

There's also some changes to how the config is handled from what was proposed.  I found a simpler way to implement what was needed, which is just a way to config this after all.  It is based on config groups, so instances for example comes with two built-in groups: factorio and instance.  The factorio group contains things like which port to run on, server-settings.json entries to override, etc,, and the instance group contains the id and name of the instance.  Plugins get to define their own groups of config fields that apply either to the master server or per instance, which comes with an enabled field pre defined for them.

### DIFF
--- a/master.js
+++ b/master.js
@@ -13,23 +13,21 @@ node master
 // const updater = require("./updater");
 // updater.update().then(console.log);
 
-// argument parsing
-const args = require('minimist')(process.argv.slice(2));
-
 const deepmerge = require("deepmerge");
 const path = require("path");
 const fs = require("fs-extra");
 const jwt = require('jsonwebtoken');
 const crypto = require('crypto');
-const base64url = require('base64url');
 const moment = require("moment");
 const request = require("request");
 const setBlocking = require("set-blocking");
 const events = require("events");
+const yargs = require("yargs");
 const version = require("./package").version;
 
-// constants
-let config = {};
+// ugly globals
+let masterConfig;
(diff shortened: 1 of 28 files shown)


master.js
```diff
@@ -339,37 +380,92 @@ class ControlConnection extends BaseConnection {
...
+		}
+	}
+
+	async assignInstanceCommandRequestHandler(message, request) {
+		let instanceConfig = db.instances.get(message.data.instance_id);
+		if (!instanceConfig) {
+			throw new errors.RequestError(`Instance with ID ${message.data.instance_id} does not exist`);
+		}
+
+		// XXX: Should the instance be stopped if it's running on another slave?
```
psihius: For now, I would say that the user has to stop the instance himself and handle the migration. We might think about developing migration later, cause it's a complexity multiplier. Don't want to feature-creep at this moment.

PR #282, opened 2020-04-04
Small things

Hornwitser:
This is a collection of small improvements and fixes that I have put off doing for quite some time now.


slave.js
```diff
@@ -807,22 +816,22 @@ async function startClient() {
...
+	// If we get here the command was run
 
 	await fs.ensureDir(slaveConfig.get("slave.instances_directory"));
 	await fs.ensureDir("sharedMods");
 	await fs.ensureDir("modules");
 
 	// Set the process title, shows up as the title of the CMD window on windows
 	// and as the process name in ps/top on linux.
-	process.title = "clusterioClient";
+	process.title = "clusterioSlave";
```
psihius, line 827: I wonder, can we also add the actual name of the instance to the process title so it is actually easy to see in the process list what is running what? :)

Hornwitser: Well, a slave runs instances, however many or few you want so putting the instance names there is probably no good.  Putting the slave name there should not be an issue though.

psihius: Well, as long as I can do a `ps -AF` and identify which process runs which instance, all I wanted here. Yeah, you right, the slave name fits better cause instance name might not be console friendly :)

PR #283, opened 2020-04-08
Fix overflow check in subspace storage

TheAIBot:
Hex was missing an f in the overflow check.
Previous and new value:

|   | Hex | Dec |
|---|---|---|
| Old | 0x7ffffff  | 134.217.727 |
| New | 0x7fffffff | 2.147.483.647|


Hornwitser:
I was sure I double checked and counted that I had the right number of Fs...

PR #289, opened 2020-04-16
Storage img fix

antt1995:


Hornwitser:
This is incorrect.   The input to the replace, `factorioLocale["<category>"][name]`, is the English Locale string for that item.  In other words, for the `steel-chest` item it would be accesses as `factorioLocale["entity-name"]["steel-chest"]` and the value for it would be `Steel chest`.  Replacing spaces with underscores is therefore correct here.

PR #293, opened 2020-04-24
Alpha bugfixes 2

Hornwitser:
Fix bad path for stylesheet and research sync breaking on technologies with a level higher than the prototype max level.


plugins/research_sync/module/sync.lua
```diff
@@ -141,6 +141,10 @@ function research_sync.sync_technologies(data)
     for _, tech_data in pairs(game.json_to_table(data)) do
         local tech = force.technologies[tech_data[nameIndex]]
         if tech and tech.enabled and tech.level <= tech_data[levelIndex] then
+            if tech_data[levelIndex] > tech.prototype.max_level then
+                tech_data[levelIndex] = tech.prototype.max_level
+            end
+
             tech.level = tech_data[levelIndex]
```
TheAIBot, line 148: Could also be implemented with math.min.
`tech.level = math.min(tech_data[levelIndex], tech.prototype.max_level)`

PR #294, opened 2020-05-11
Integrate plugins closer with connections

Hornwitser:
Adds tracking of plugins available in master-slave connections, preventing plugins not on the master server from being loaded on slaves.  Adds plugin events for when the connection is connected, dropped, closed and preparing to disconnect.  Adds plugin event for instance status updates.  And fixes a heap of bugs with connections ands the API documentation.

### DIFF
--- a/lib/link/connectors.js
+++ b/lib/link/connectors.js
@@ -115,6 +115,20 @@ class WebSocketBaseConnector extends events.EventEmitter {
 	setClosing() {
 		this._state = "closing";
 	}
+
+	/**
+	 * True if the connection is established and active
+	 */
+	get connected() {
+		return this._state === "connected";
+	}
+
+	/**
+	 * True if the connection is closing down or closed
+	 */
+	get closing() {
+		return ["closing", "closed"].includes(this._state);
+	}
 }
 
 /**
--- a/lib/link/connectors.js
+++ b/lib/link/connectors.js
@@ -155,13 +155,6 @@ class WebSocketClientConnector extends WebSocketBaseConnector {
 		this._reconnectDelay = reconnectDelay;
 		this._timeout = 15 * 60;
 		this._startedReconnect = null;
-
-		/**
-		 * Mapping of plugin name to plugin version loaded on the master
-		 * server.  Is `null` before handshake has taken place.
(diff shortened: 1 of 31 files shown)


lib/link/connectors.js
```diff
@@ -36,7 +43,7 @@ class WebSocketBaseConnector extends events.EventEmitter {
 	}
 
 	_doHeartbeat() {
-		this._check(["connected"]);
+		this._check(["connected", "closing"]);
 		if (Date.now() - this._lastHeartbeat > 2000 * this._heartbeatInterval) {
```
psihius, line 47: Hm, thinking about this, maybe the heartbeat multiplier - 2000 - used here should be a config option so you can actually configure what the heartbeat timeout you want to use? Some people might prefer having a way smaller timeout interval than a multiplier of 2000 retries...

Hornwitser: I think you need to pay a bit more attention to the units here.  Date.now() is unix time in milliseconds, this._heartbeatInterval is in seconds.  The logic is saying if the time since the last heartbeat is greater than 2x the heartbeat interval.

psihius: Can we, maybe, then homogenise the values here to a single base and not mix seconds and microseconds? It either should be microseconds and interval should be defined in those or hide the microseconds and do calculations in seconds as the base.
Personally, I rather just use microseconds overall. Mixing value bases for times is always leading to some kind of WTF later.

Hornwitser: Would `(Date.now() - this._lastHeartbeat) / 1000 > 2 * this._heartbeatInverval` be acceptable?  This isn't really relevant to this PR though and we can't change JavaScript functions like Date.now(), SetInterval/SetTimeout operating in milliseconds.  Clusterio code like the protocol and config options are in seconds, Prometheus metrics are _strongly_ recommended to be in seconds and I for one do not like configuring things like autosave interval in milliseconds.

psihius: That works, but to be honest it does read even more convoluted.

Provided we use microseconds for _heartbeatInterval, could be `(Date.now() * 1000) - this.lastHeartbeat > 2 * this._heartbeatInterval` ?

PR #298, opened 2020-05-31
Config events

Hornwitser:
Implementation of config update event, plugin hook for config update, and live update of factorio.settings to running instances using the `/config set` command.  This allows updating things like the name, description, tags, max players, visibility, upload slots, etc, via the Clusterio interface without having to restart the instance for it to take effect.

I also added a set-instance-config-prop command for updating a single property of an object config without having to give the whole object.  This makes the testing simple and allows updating for example only the name of the factorio.settings object config.

### DIFF
--- a/clusterctl.js
+++ b/clusterctl.js
@@ -224,6 +224,26 @@ commands.push(new Command({
 }));
 
 commands.push(new Command({
+	definition: ["set-instance-config-prop", "Set property of field in instance config", (yargs) => {
+		yargs.option({
+			"instance": { describe: "Instance to set config on", nargs: 1, type: "string", demandOption: true },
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -2,7 +2,8 @@ language: node_js
 node_js:
   - 10
 before_install:
-  - curl -o factorio.tar.gz -L https://www.factorio.com/get-download/latest/headless/linux64
+  # For some reason certificate verification just randomly decided to fail :/
+  - curl --insecure -o factorio.tar.gz -L https://www.factorio.com/get-download/latest/headless/linux64
```
psihius, line 6: They had a certificate expire, they updated them. It should be fine now. Or is it a random recurring thing that makes no sense?

PR #307, opened 2020-08-26
master.js: fix missing response for /api/getID

mlyszczek:
According to rfc2616#section-6:

> After receiving and interpreting a request message, a server
> responds with an HTTP response message.

That response was missing for /api/getID which could lead to
EMFILE, because needle module was waiting indefiniately for
response from server. Since client sends /api/getID once every 10
seconds, this piles up, ultimately reaching ulimit for opened file
descriptors.

Signed-off-by: Michał Łyszczek <michal.lyszczek@bofc.pl>


Hornwitser:
`res.sendStatus` should be done regardless of whether there is a `req.body` or not .

mlyszczek:
Is getID call valid without body? I mean, should return code for empty body be also 200, or something like 400?

Hornwitser:
I don't think it matters what you reply with, since it didn't reply with anything before, as long as it doesn't break more.  1.2.x is legacy and more or less all the code here has been axed in 2.0.

PR #312, opened 2020-08-30
Add status (stopped, running, initialized) to listInstances

Danielv123:
as discussed in discord.

### DIFF
--- a/lib/link/messages.js
+++ b/lib/link/messages.js
@@ -294,6 +294,7 @@ messages.listInstances = new Request({
 					"name": { type: "string" },
 					"id": { type: "integer" },
 					"assigned_slave": { type: ["null", "integer"] },
+					"status": { enum: ["stopped", "initialized", "running"] },
 				},
 			},
 		},
--- a/master.js
+++ b/master.js
@@ -616,6 +616,7 @@ class ControlConnection extends BaseConnection {
 				id: instance.config.get("instance.id"),
 				name: instance.config.get("instance.name"),
 				assigned_slave: instance.config.get("instance.assigned_slave"),
+				status: instance.status,
 			});
 		}
 		return { list };


psihius (approved):

lib/link/messages.js
```diff
@@ -294,6 +294,7 @@ messages.listInstances = new Request({
 					"name": { type: "string" },
 					"id": { type: "integer" },
 					"assigned_slave": { type: ["null", "integer"] },
+					"status": { enum: ["stopped", "initialized", "running"] },
```
Hornwitser, line 297: This should be added to the required list a few lines up in the schema as well.  I've usually added the type parameter for enum validated items, though I see the updateInstances event has it without `type: "string"` as well.

Hornwitser (changes requested):
There seems to be a missing component here.  If you start up a master and list its instances before the slave the instances are on connects, the status is undefined.  That should not happen, the status should be well defined in all cases.

I suggest adding two more statuses to be possible for instances: `unknown` and `unassigned`.  Additionally: 
- In `updateInstancesEventHandler` there's a case where an instance is set with db.instances.set which doesn't include the status, it should be set to the status received from the slave.
- In `createInstanceRequestHandler` the newly created instance should have status `unassigned`.
- In `loadInstances` the loaded instance should have status `unknown` if `instance.config.get("instance.assigned_slave")` is non null, otherwise `unassigned`.

psihius:
@Danielv123 So what's the status on this one then?

Danielv123:
I believe this PR might have broken due to the reorganization PR being merged, but it’s a 2 line change. It is important as the web interface doesn’t make much sense without this functionality (having to press start instance to get an error about instance already started is just horrible UX), but I don’t believe we ever came to an agreement about how to define all the undefined behavior.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

Hornwitser:
I can redo this with the changes I've suggested later today.  It would make the status much more sensible than the code here.

PR #313, opened 2020-09-01
List management

Hornwitser:
Add integration of the server adminlist, whitelist and banlist into Clusterio.  Membership is stored as booleans in the user accounts and changes are synchronized live to slaves which pass it on to instances which in turn will update the running server's lists with RCON commands.

Test coverage is more than likely poor on this.  I seem to be in the can't be bothered to write good coverage mood for the time being.

### DIFF
--- a/master.js
+++ b/master.js
@@ -504,7 +504,7 @@ class BaseConnection extends link.Link {
 		event.send(connection, message.data);
 	}
 
-	async broadcastEventToInstance(message, event) {
+	async broadcastEventToSlaves(message, event) {
 		for (let slaveConnection of slaveConnections.values()) {
 			// Do not broadcast back to the source
 			if (slaveConnection === this) { continue; }
@@ -513,6 +513,10 @@ class BaseConnection extends link.Link {
 
 			event.send(slaveConnection, message.data);
 		}
+	}
+
+	async broadcastEventToInstance(message, event) {
+		await this.broadcastEventToSlaves(message, event);
 	}
 
 	async prepareDisconnectRequestHandler(message, request) {
@@ -803,6 +807,9 @@ class ControlConnection extends BaseConnection {
 			list.push({
 				name: user.name,
 				roles: [...user.roles].map(role => role.id),
+				is_admin: user.isAdmin,
+				is_banned: user.isBanned,
+				is_whitelisted: user.isWhitelisted,
 				instances: [...user.instances],
 			});
 		}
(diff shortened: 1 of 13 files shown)


psihius:
Maybe someone else has the time and will to work on tests for this?

RMuskovets (changes requested):
I wrote some tests for `adminlistUpdateEventHandler`, more to come for both other functions and this one:
```js
describe(".adminlistUpdateEventHandler()", function() {
    it("should throw when `message` not given", function() {
        assert.throws(() => instance.adminlistUpdateEventHandler(), "`message` must be a object");
    });
    it("should throw when `message` isn't a object", function() {
        assert.throws(() => instance.adminlistUpdateEventHandler(null), "`message` must be a object");
        assert.throws(() => instance.adminlistUpdateEventHandler(""), "`message` must be a object");
        assert.throws(() => instance.adminlistUpdateEventHandler([]), "`message` must be a object");
        assert.throws(() => instance.adminlistUpdateEventHandler(0), "`message` must be a object");
        assert.throws(() => instance.adminlistUpdateEventHandler(false), "`message` must be a object");
    });
    it("should throw on empty `message`", function() {
        assert.throws(() => instance.adminlistUpdateEventHandler({}), "`message` must not be empty");
    });
    it("should throw on non-string `name`", function() {
        assert.throws(() => instance.adminlistUpdateEventHandler({name: undefined}), "`name` must be a string");
        assert.throws(() => instance.adminlistUpdateEventHandler({name: null}), "`name` must be a string");
        assert.throws(() => instance.adminlistUpdateEventHandler({name: {}}), "`name` must be a string");
        assert.throws(() => instance.adminlistUpdateEventHandler({name: []}), "`name` must be a string");
        assert.throws(() => instance.adminlistUpdateEventHandler({name: 0}), "`name` must be a string");
        assert.throws(() => instance.adminlistUpdateEventHandler({name: false}), "`name` must be a string");
    });
    it("should throw on non-boolean `admin`", function() {
        assert.throws(() => instance.adminlistUpdateEventHandler({admin: undefined}), "`admin` must be boolean");
        assert.throws(() => instance.adminlistUpdateEventHandler({admin: null}), "`admin` must be boolean");
        assert.throws(() => instance.adminlistUpdateEventHandler({admin: {}}), "`admin` must be boolean");
        assert.throws(() => instance.adminlistUpdateEventHandler({admin: []}), "`admin` must be boolean");
        assert.throws(() => instance.adminlistUpdateEventHandler({admin: 0}), "`admin` must be boolean");
        assert.throws(() => instance.adminlistUpdateEventHandler({admin: ""}), "`admin` must be boolean");
    });
});
```

Hornwitser:
Event handlers are invoked by lib/link and the input message is validated before the message is invoked.  See [processMessage of the Link class](https://github.com/clusterio/factorioClusterio/blob/a8b8232d730bab5f85edd74183d791ffaa209859/lib/link/link.js#L80), and the adminlistUpdate message definition in lib/link/messages.js.  All of these tests are for inputs that the validation would fail for the message and therefore not invoke the handler, so they don't seem to be much of use.

The things that I would like to have tested, but which is hard to test is the logic for resyncing the lists in the case where the connection from the slave to the master is dropped, the list is changed and then the connection is re-established.

psihius (approved):

master.js
```diff
@@ -832,10 +839,71 @@ class ControlConnection extends BaseConnection {
...
 		}
+
+		if (user.is_admin) {
+			this.broadcastEventToSlaves({ data: { name, admin: false }}, link.messages.adminlistUpdate);
+		}
+		if (user.is_whitelisted) {
+			this.broadcastEventToSlaves({ data: { name, whitelisted: false }}, link.messages.whitelistUpdate);
+		}
+		if (user.is_banned) {
+			this.broadcastEventToSlaves({ data: { name, banned: false, reason: "" }}, link.messages.banlistUpdate);
```
Danielv123, line 904: Does this not support ban reason? Seems like something users would expect to have.

Hornwitser: Ban with ban reason is supported.  What you see here is the notification of removal from the ban list due to the user account being deleted.

Danielv123: I see

Issue #316, opened 2020-09-17
[Design Proposal] Restructure and repack code


Hornwitser:
I've been thinking about this for some time now.  If we publish Clusterio and its plugins as packages on NPM it would solve the whole installing and updating process for both Clusterio and plugins.  The specifics that I had in mind was to use [Lerna](https://lerna.js.org/) in order to split the code into the packages: lib, master, slave, and ctl, under the npm org clusterio.

The basic idea is that these will be proper packages hosted on npm and that installation of clusterio will go through npm.  I.e., a working master server would be possible to bring up with `npm init && npm install @clusterio/master` followed by `npx clusterio-master run`.  This presuming of course that the master package exports a `clusterio-master` script.

File structure of the repository after this change would be similar to #214.  Something like this:

```
factorioClusterio/
+- CHANGELOG.md
+- README.md
+- LICENSE.md
+- lerna.json
+- packages/
|  +- lib/ - Code shared between master and slaves
|  |  +- package.json - lib dependencies
|  |  +- prometheus.js
|  |  +- link/
|  |  +- ...
|  +- master/
|  |  +- package.json - master dependencies
|  |  +- master.js
|  |  +- static/
|  |  +- routes.js
|  +- slave/
|  |  +- package.json - slave dependencies
|  |  +- config.json
|  |  +- slave.js
|  +- ctl/
|  |  +- package.json - clusterctl dependencies
|  |  +- clusterctl.js
|  +- plugin-global_chat/
|     +- package.json - global_chat's dependencies
|     +- info.js - Plugin definiton
|     +- ... plugin files.
+- sharedMods/
```

Plugins turned into npm packages will get usdependency resolution and installation for free which would also resolve #196.  This does necessitate the introduction of some kind of plugin array in the config that says which plugins have been installed as there's no easy way to figure out which npm packages are installed and compatible plugins from npm.  I am thinking plugin installation would be done with some kind of `add-plugin` command line tool which installs the package with npm and adds it to the necessary config array.

The separate web interface project would be able to pull `@clusterio/lib` directly as a regular npm dependency after this change.

-----

[...]

PR #318, opened 2020-09-21
Restructure project

Hornwitser:
Implementation of the package structure proposed in #316 with the exception of plugins and support for loading plugins from node packages.  To be consistent with the other binaries clusterctl is renamed to clusterioctl.


psihius (approved):

.travis.yml
```diff
@@ -6,7 +6,8 @@ before_install:
   - tar -xf factorio.tar.gz
   - npm install -g codecov
 script:
-  - node build-mod --bump-patch --factorio-version 0.18 --output-dir sharedMods
+  - npx lerna bootstrap --no-ci
```
Danielv123, line 9: Why would you use --no-ci on CI?

README.md
```diff
@@ -105,12 +105,13 @@ supported on EOL Ubuntu releases so make sure you're on a recent release of Ubun
...
 
     wget -qO - https://deb.nodesource.com/setup_12.x | sudo -E bash -
-    sudo apt install -y nodejs python-dev git build-essential
-    git clone -b master https://github.com/clusterio/factorioClusterio.git
-    cd factorioClusterio
+    sudo apt install -y nodejs
+    mkdir clusterio
+    cd clusterio
+    npm init -y
+    npm install @clusterio/master @clusterio/slave @clusterio/ctl
```
Danielv123, line 112: Possibly use npm init @clusterio ?

README.md
```diff
@@ -136,30 +137,19 @@ The -v flag is used to specify the instance directory. Your instances (save file
...
 
 1. Open PowerShell or Command prompt in the directory you want to install to and run the following commands.
 
-        git clone -b master https://github.com/clusterio/factorioClusterio
-        cd factorioClusterio
-        npm install --only=production
+        mkdir clusterio
+        cd clusterio
+        npm init -y
+        npm install @clusterio/master @clusterio/slave @clusterio/ctl
```
Danielv123, line 152: Same here ^

docs/communication.md
```diff
@@ -49,9 +49,10 @@ Message
 
 There's two types of messages currently implemented apart from the
 Link builtin close message and that's Request and Event.  The builtin
-messages are defined in [lib/link/messages.js](../lib/link/messages.js)
-and are attached to links via the attachAllMessages function, which is
-usally called in the Link subclass constructor.
+messages are defined in
+[lib/link/messages.js](/package/lib/link/messages.js) and are attached
```
Danielv123, line 53: Path is updated, text is not?

docs/config-system.md
```diff
@@ -6,15 +6,15 @@ configuration system in Clusterio, and unless you're developing
...
+the master server, instances, slaves, and the clusterioctl utility.
 Plugins get their own config group for the master server and instances
 which they can add their own fields to.  The fields are defined early in
 the startup, and the groups and their fields for disabled plugins will
 still be created.
 
 The built-in config groups and their fields are defined in
-[/lib/config/definitions.js](lib/config/definitions.js) while plugins
-can add their own config groups to the `MasterConfigGroup` and
+[lib/config/definitions.js](/packages/lib/config/definitions.js) while
```
Danielv123, line 16: Same here. This is developer documentation, I don't think its a given that a new developer would know to look in /packages

Danielv123 (changes requested):

README.md
Danielv123: You will have to decide if you want to add that here or later. Either makes sense.

.travis.yml
Hornwitser: As noted in the commit that does that it reqires package-lock.json files to be commited to the repository and we do not for the time being do this.

README.md
Hornwitser: Can't do that until the `@clusterio/create` exists, which it doesn't.  This is a method that can be used until such a package is created.

Hornwitser: To reiterate, the documentation should reflect the current state of the project.  Not some future idea that may or may not have been implemented.  If the create package is implemented then the documentation here will be changed to reflect that.

.travis.yml
Danielv123: Makes sense, we should maybe consider doing that sometime, but that doesn't belong here.

Danielv123 (approved):

PR #320, opened 2020-09-28
Import new web interface

Hornwitser:
This the ground work for having the web interface code imported into the main repository.  The motivation for importing it here is that coordinating changes to the codebase is simpler when it's in one repository.  The development and deployment also becomes much simpler, the master server can integrate directly with the webpack build process and have changes to the web code trigger an automatic rebuild during development.  When publishing the `@clusterio/master` package the production build of the interface can be included such that simply running the master server will have the web interface working.


psihius (approved):
No objections from me

PR #321, opened 2020-09-28
Rework instance status

Hornwitser:
Extended implementation of #312 which adds unassigned and unknown statuses and replaces the initialized status with starting, creating_save, and exporting_data.


psihius (approved):
Looks reasonable

Danielv123 (approved):

PR #323, opened 2020-10-21
Convert plugins to node packages

Hornwitser:
Introduces a plugin-list.json config with a mapping of plugin name to require path.  This replaces the old behaviour of loading everything found in the plugins directory.  Now you have to add plugins after installing them with `clusterio<side> plugin add <require_path>`.  This gives the benefit of being able to publish plugins as packages on npm, and then install them via npm.

For development of plugins providing a path to where the plugin is installed is also supported.  This also means you no longer need to put the repository of the plugin inside the Clusterio directory when doing plugin development.

Also fixes the lua code being located outside the the package tree and changes integration tests to make sure paths are correctly handled.  After this it should be possible to publish this to npm.


psihius (approved):
Does look good, having the ability to stick plugins anywhere on the filesystem does sound like a decent idea.

PR #325, opened 2020-10-23
Lint fixes 2

Hornwitser:
Fixes to bring lint warnings down from 547 to 229.  Notable changes are prefixing imports from lib with lib, e.g. `const libPlugin = require("@clusterio/lib/plugin")` and removing the defunct inventoryImports plugin.  The rationale behind prefixing lib imports is that the names for the modules in it are very generic and shadows otherwise useful variable names.  The lib prefix also helps with distinguishing where these come from, `libPlugin` is much clearer than `plugin`.


packages/lib/prometheus.js
```diff
@@ -93,29 +93,48 @@ function labelsToKey(labels, metricLabels) {
...
-		labels.set(name, value
-			.slice(1, -1)
-			.replace(/\\"/g, "\"")
-			.replace(/\\n/g, "\n")
-			.replace(/\\\\/g, "\\")
-		);
+	if (key !== "") {
+		for (let pair of key.split(",")) {
+			let [name, value] = pair.split("=", 2);
+			labels.set(name, value
```
psihius, line 99: I wonder if this can be done as a single RegExp or just run multiple replace commands without the regexp engine?

psihius (approved):
Looks good, really improved readability at a glance.

packages/lib/prometheus.js
Hornwitser: If you use a string instead of a RegExp with the global flag only the first occurrence is replaced, so that's not an option.  It's possible to use a single RegExp with a replacement function to select the appropriate replacement instead of a string though.

psihius: Well, it's up to you if you wanna replace it with single RegExp. It's just initialising and running even a single RegExp is quite a heavy operation and we have 3 here.

PR #326, opened 2020-10-26
Publish fixes

Hornwitser:
This should fix the web interface build being missing from the master package.  Not entirely certain, but that's what alpha releases are for.


psihius (approved):
Ship it

PR #327, opened 2020-10-26
Add package readmes

Hornwitser:
Adds README.md files for all the published packages and a description of the available configuration settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,8 +248,8 @@ the config for a remote slave you will need to have set the
 reached on.
 
 You can list the config of a slave on the slave itself with the `npx
-clusterioslave config list` command.  Use `npx clusterioslave config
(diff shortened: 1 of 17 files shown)


psihius:
I was thinking that these probably should be linked to from the main readme file, otherwise you kind'a have to know to search for these.

Hornwitser:
Hmm, well yes.  But this sort of applies to all of the files in docs/.  The project readme needs a clean up too.

Hornwitser:
I added links in the main readme file, updated the description of Clusterio in it to be a lot less out of date, added a section of configuring the master server port, setting it up behind a proxy, and removed claims of supporting Node.js v11 as that's EOL.

psihius (approved):

PR #336, opened 2020-11-08
Fix crash on incorrect authentication token

Hornwitser:
Fix client.js exiting abruptly when the getID endpoint returns something other than a string as the body, this which happens on authentication failure.


Danielv123 (approved):
Makes sense.

PR #337, opened 2020-11-08
Alpha 2 preparations

Hornwitser:
Lots of tiny things in preparation for alpha.2.  See commits for detailed descriptions but resolves mostly around docs fixes, bug fix in generate-token command, lint fixes, some moving in lib to avoid warnings and imports not available in web build, and responsive design fixes for the web UI.


packages/lib/factorio/patch.js
```diff
@@ -152,7 +152,7 @@ function reorderDependencies(modules) {
 
 		// No unmet dependencies, record and continue
 		present.set(module.name, module.version);
-		index++;
+		index += 1;
```
psihius, line 155: Is there like a lint rule or something that mandate changing it from increment to this form?

packages/lib/link/connectors.js
```diff
@@ -102,12 +102,14 @@ class WebSocketBaseConnector extends events.EventEmitter {
 		if (!["handshake", "connected", "closing"].includes(this._state)) {
 			throw new libErrors.SessionLost("No session");
 		}
-		let message = { seq: this._seq, type, data };
+		let seq = this._seq;
```
psihius, line 105: I'm curious - race conditions or just making code more readable and understandable? :)

Danielv123 (approved):
Looks good, I love how clean the tests are

packages/lib/link/connectors.js
Danielv123: Readability - he increments this._seq, previously it was done in the return statement further down which was unclear.

packages/lib/factorio/patch.js
Danielv123: There was a warn for it, he made it an error instead. See .eslintrc.js https://github.com/clusterio/factorioClusterio/pull/337/commits/5337280eec4426873e11e6aedbc6602ecc904a94

psihius (approved):

PR #338, opened 2020-11-12
Assorted improvements

Hornwitser:
Small fixes and features accumulated from the experience with testing alpha.2 and developing edge transports.  See individual commit descriptions for what has changed.

### DIFF
--- a/packages/ctl/ctl.js
+++ b/packages/ctl/ctl.js
@@ -43,7 +43,11 @@ function formatOutputColored(output) {
 	let info = "";
 	if (output.type === "log") {
 		let level = output.level;
-		if (level === "Info") {
+		if (level === "Script") {
+			level = chalk.bold.greenBright(level);
+		} else if (level === "Verbose") {
+			level = chalk.bold.gray(level);
+		} else if (level === "Info") {
 			level = chalk.bold.blueBright(level);
 		} else if (output.level === "Warning") {
 			level = chalk.bold.yellowBright(level);
--- a/packages/lib/factorio/server.js
+++ b/packages/lib/factorio/server.js
@@ -318,13 +318,15 @@ function parseOutput(line, source) {
 		// A seconds output has two different kinds of formats: The first is a
 		// log level and source location and has a format of
 		// "Level File.cpp:123: message"
-		const secLogRegex = /^(\w+) (\w+\.cpp:\d+): (.*)$/;
+		// "Script @/path/to/file.lua:123: message"
+		// "Script =(command):123: message"
+		const secLogRegex = /^(\w+) ((\w+\.cpp:\d+)|([@=].*?:\d+)): (.*)$/;
 		let secLogMatch = secLogRegex.exec(secContent);
 		if (secLogMatch) {
 			output.type = "log";
 			output.level = secLogMatch[1];
 			output.file = secLogMatch[2];
-			output.message = secLogMatch[3];
+			output.message = secLogMatch[5];
 
 		// The other possibility is that the content is a generic message
 		} else {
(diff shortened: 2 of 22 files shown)


packages/ctl/ctl.js
```diff
@@ -43,7 +43,11 @@ function formatOutputColored(output) {
 	let info = "";
 	if (output.type === "log") {
 		let level = output.level;
-		if (level === "Info") {
+		if (level === "Script") {
```
psihius, line 46: The logging levels are better done as constants. You know how it goes, you think it's gonna be used only once and suddenly strings are all over the code :)

packages/lib/factorio/server.js
```diff
@@ -517,12 +538,27 @@ class FactorioServer extends events.EventEmitter {
 		this.rconPassword = options.rconPassword;
 		/** Enable player whitelist */
 		this.enableWhitelist = options.enableWhitelist;
+		/** Enable verbose logging */
+		this.verboseLogging = options.verboseLogging;
 		this._state = "new";
```
psihius, line 543: For fixed things like `_state` here probably also a good idea to make it actual constants and use those.
In PHP I would make it part of the class that's responsible for it, idk how best to do it in JS though.

packages/slave/slave.js
```diff
@@ -218,6 +233,7 @@ class Instance extends libLink.Link{
 	}
 
 	async init(pluginInfos) {
+		this.notifyStatus("starting");
```
psihius, line 236: Maybe it's my state of being at 5am talking here, and I probably am overthinking this, but we could use something like `this.notifyStatus(new StartingMessage())`. But this s not PHP, loading classes is a manual job and in general I think I'm getting ahead of myself :D

packages/web_ui/src/components/InstanceViewPage.jsx
```diff
@@ -83,7 +83,11 @@ function formatOutputColored(output, key) {
 	let info = "";
 	if (output.type === "log") {
 		let level = output.level;
-		if (level === "Info") {
+		if (level === "Script") {
```
psihius, line 86: The levels already are used twice in the code. One typo and you are racking your brain why something does not work for hours. Let's just say I have wasted A LOT of time to this specific error.... :D

Hornwitser: I get your sentiment here.  I've banged my head on similar issues before.  But I don't think putting it as constants is going to help much.  You see in JavaScript everything is objects, so the way it would have to be exposed would be through properties of an object.  Like for example

```js
const levels = {
    INFO: "Info",
    SCRIPT: "Script,
};
```

And then you'd access it like `libFactorio.levels.INFO`, or some appropriate place.  The problem is that accessing a property that doesn't exist is not an error, it just returns `undefined`.  So if you fat finger it like `if (level === libFactorio.levels.IINFO)` it wouldn't become an error, it would become equivalent to `if (level === undefined)`  which works more or less just the same as `if (level === "IInfo")` does.

If we were using TypeScript we could easily leverage the type system to prevent these kinds of mistakes, and I'd totally be for taking advantage of such things.  But we're not, so for now I would say it's another reason to use TypeScript.

psihius: The idea here is that this would be caught on the syntax highlighting and by code analysis tooling. 
And code auto-complete makes it easier to not make a mistake. 

It's true - for runtime, it will make no difference if it's one or the other type of error. We are talking about human here :)

Danielv123: I don't think it makes much of a difference for him if he still uses Vim.

Danielv123 (approved):
Love the changes to logging and error handling, personally don't mind the strings as long as its easy to figure out all the possible options when working with the endpoints. They seemed to be documented well enough in comments.

psihius (approved):

PR #341, opened 2020-11-30
Implement logging system

Hornwitser:
Replaces scattered console logs with an integrated cluster logging system based on windston. Slaves send log messages from themselves and instances to the master server which stores it in a single big flat file. Resolves the following issues:
- Lack of reporting back to the source. Now if an instance plugin logs a warning or error it shows up in the instance console.
- Spammy console, the level output to stdout can be configured and defaults to info, which suppresses a lot of verbose messages.
- Lack of instance log. The instance output is sent to the cluster log which can be queried now. This PR implements loading the last 400 entries in the instance log in the web interface (and fixes scroll to bottom, yay!)
- Lots of irrelevant junk in the test log. With verbose being ignored by default most of it is gone now.

The performance is not ideal, right now it's taking 100ms to query the cluster log with 17k entries in it. The scaling is linear so that means a log with a million entries would take 5 seconds to query. I have some ideas for how to improve this performance, but that'll have to come later.


psihius (commented):
The amount of test coverage warnings is kind'a high on this one. 
Probably a good idea to update the tests and try to keep the test coverage, especially since we are getting to a point of things actually starting to work.

psihius (approved):

PR #347, opened 2020-12-23
Expose slave public_address

Danielv123:
Exposes the public_address field from the slave config in messages.listSlaves and in the table in the web interface as requested by L1n3 on discord. The field was made optional as exposing the slave to a public address shouldn't be mandatory IMO.

I think I caught all the places that need updating for this, not entirely sure though.


Hornwitser:
This doesn't actually make it available though. The slave objects on the master are created from the register_slave message and that doesn't include the public address.

PR #348, opened 2020-12-25
Expose some column

yw9381:
Web UI can show public_address, game_port,assigned_slave_name
slave can send slave.public_address
master can recv and save slave.public_address

### DIFF
--- a/packages/lib/link/messages.js
+++ b/packages/lib/link/messages.js
@@ -293,6 +293,7 @@ messages.listSlaves = new Request({
 					"agent": { type: "string" },
 					"version": { type: "string" },
 					"name": { type: "string" },
+					"public_address": { type: "string" },
 					"id": { type: "integer" },
 					"connected": { type: "boolean" },
 				},
--- a/packages/lib/link/messages.js
+++ b/packages/lib/link/messages.js
@@ -315,6 +315,9 @@ messages.listInstances = new Request({
 					"name": { type: "string" },
 					"id": { type: "integer" },
 					"assigned_slave": { type: ["null", "integer"] },
+					"assigned_slave_name": { type: ["null", "string"] },
+					"public_address": { type: ["null", "string"] },
+					"game_port": { type: ["null", "integer"] },
 					"status": { enum: [
 						"unknown", "unassigned", "stopped", "starting", "running", "creating_save", "exporting_data",
 					]},
--- a/packages/lib/link/messages.js
+++ b/packages/lib/link/messages.js
@@ -311,11 +311,10 @@ messages.listInstances = new Request({
 			type: "array",
 			items: {
 				additionalProperties: false,
-				required: ["name", "id", "assigned_slave", "status"],
+				required: ["name", "id", "assigned_slave_name", "public_address", "game_port", "status"],
 				properties: {
 					"name": { type: "string" },
 					"id": { type: "integer" },
-					"assigned_slave": { type: ["null", "integer"] },
 					"assigned_slave_name": { type: ["null", "string"] },
 					"public_address": { type: ["null", "string"] },
 					"game_port": { type: ["null", "integer"] },
--- a/packages/master/master.js
+++ b/packages/master/master.js
@@ -70,6 +70,7 @@ const cookieParser = require("cookie-parser");
 // Required for express post requests
 const bodyParser = require("body-parser");
 const fileUpload = require("express-fileupload");
+const { push } = require("./src/routes");
 const app = express();
 let httpServer;
 let httpsServer;
@@ -257,6 +258,14 @@ function validateSlaveToken(req, res, next) {
 	}
 
 	next();
+}
+
+function findInstancesAssignedSlave(assigned_slave_id){
+	for (let slave of db.slaves.values()) {
+		if (assigned_slave_id == slave.id) {
+			return slave;
+		}
+	}
 }
 
 app.get("/api/plugins", (req, res) => {
@@ -680,22 +689,34 @@ class ControlConnection extends BaseConnection {
 		if (message.data.name !== null) {
 			slaveConfig.set("slave.name", message.data.name);
 		}
+		if (message.data.public_address !== null) {
+			slaveConfig.set("slave.public_address", message.data.public_address);
+		}
 		if (message.data.generate_token) {
 			this.user.checkPermission("core.slave.generate_token");
 			slaveConfig.set("slave.master_token", this.generateSlaveToken(slaveConfig.get("slave.id")));
 		}
 		return { serialized_config: slaveConfig.serialize() };
 	}
-
 	async listInstancesRequestHandler(message) {
 		let list = [];
 		for (let instance of db.instances.values()) {
-			list.push({
+			let push_info = {
 				id: instance.config.get("instance.id"),
 				name: instance.config.get("instance.name"),
 				assigned_slave: instance.config.get("instance.assigned_slave"),
+				assigned_slave_name: null,
+				game_port: instance.config.get("factorio.game_port"),
+				public_address: null,
 				status: instance.status,
-			});
+			}
+			// get assigned slave info
+			if (instance.config.get("instance.assigned_slave") != null) {
+				let assigned_slave_info = findInstancesAssignedSlave(instance.config.get("instance.assigned_slave"));
+				push_info.assigned_slave_name = assigned_slave_info.name;
+				push_info.public_address = assigned_slave_info.public_address;
+			}
+			list.push(push_info);
 		}
 		return { list };
 	}
@@ -1089,12 +1110,14 @@ class SlaveConnection extends BaseConnection {
 		this._id = registerData.id;
 		this._name = registerData.name;
 		this._version = registerData.version;
+		this._public_address = registerData.public_address;
 		this.plugins = new Map(Object.entries(registerData.plugins));
 
 		db.slaves.set(this._id, {
 			agent: this._agent,
 			id: this._id,
 			name: this._name,
+			public_address: this._public_address,
 			version: this._version,
 			plugins: registerData.plugins,
 		});
--- a/packages/master/master.js
+++ b/packages/master/master.js
@@ -260,12 +260,15 @@ function validateSlaveToken(req, res, next) {
 	next();
 }
 
-function findInstancesAssignedSlave(assigned_slave_id){
+function findInstancesAssignedSlave(assigned_slave_id) {
+	let ret = null;
 	for (let slave of db.slaves.values()) {
-		if (assigned_slave_id == slave.id) {
-			return slave;
+		if (assigned_slave_id === slave.id) {
+			ret = slave;
+			break;
 		}
 	}
+	return ret;
 }
 
 app.get("/api/plugins", (req, res) => {
@@ -695,6 +698,7 @@ class ControlConnection extends BaseConnection {
 		}
 		return { serialized_config: slaveConfig.serialize() };
 	}
+
 	async listInstancesRequestHandler(message) {
 		let list = [];
 		for (let instance of db.instances.values()) {
--- a/packages/master/master.js
+++ b/packages/master/master.js
@@ -689,9 +689,6 @@ class ControlConnection extends BaseConnection {
 		if (message.data.name !== null) {
 			slaveConfig.set("slave.name", message.data.name);
 		}
-		if (message.data.public_address !== null) {
-			slaveConfig.set("slave.public_address", message.data.public_address);
-		}
 		if (message.data.generate_token) {
 			this.user.checkPermission("core.slave.generate_token");
 			slaveConfig.set("slave.master_token", this.generateSlaveToken(slaveConfig.get("slave.id")));
--- a/packages/master/master.js
+++ b/packages/master/master.js
@@ -709,9 +709,9 @@ class ControlConnection extends BaseConnection {
 				game_port: instance.config.get("factorio.game_port"),
 				public_address: null,
 				status: instance.status,
-			}
+			};
 			// get assigned slave info
-			if (instance.config.get("instance.assigned_slave") != null) {
+			if (instance.config.get("instance.assigned_slave") !== null) {
 				let assigned_slave_info = findInstancesAssignedSlave(instance.config.get("instance.assigned_slave"));
 				push_info.assigned_slave_name = assigned_slave_info.name;
 				push_info.public_address = assigned_slave_info.public_address;
--- a/packages/master/master.js
+++ b/packages/master/master.js
@@ -260,17 +260,6 @@ function validateSlaveToken(req, res, next) {
 	next();
 }
 
-function findInstancesAssignedSlave(assigned_slave_id) {
-	let ret = null;
-	for (let slave of db.slaves.values()) {
-		if (assigned_slave_id === slave.id) {
-			ret = slave;
-			break;
-		}
-	}
-	return ret;
-}
-
 app.get("/api/plugins", (req, res) => {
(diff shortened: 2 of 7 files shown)


yw9381:
WebUI
image
Slave send
image

packages/lib/link/messages.js
```diff
@@ -315,6 +316,9 @@ messages.listInstances = new Request({
 					"name": { type: "string" },
 					"id": { type: "integer" },
 					"assigned_slave": { type: ["null", "integer"] },
+					"assigned_slave_name": { type: ["null", "string"] },
```
Danielv123, line 319: Currently slave IDs are mapped to slave name independently in the places where it is required. Always providing it with the request is useful and does cut down on one request in many cases while consuming minimal bandwidth, but it is still a calculated parameter. @Hornwitser probably has some thoughts on this. Also, if we are doing it this way we should probably use this field on the instance view page instead of using an additional request.

packages/lib/link/messages.js
```diff
@@ -315,6 +316,9 @@ messages.listInstances = new Request({
...
+					"public_address": { type: ["null", "string"] },
+					"game_port": { type: ["null", "integer"] },
```
Danielv123, line 321: Placing this right on the instance is useful

packages/web_ui/src/components/InstancesPage.jsx
```diff
@@ -18,7 +18,9 @@ export default function InstancesPage() {
 		return result["list"].map(item => ({
 			key: item["id"],
 			"Name": item["name"],
-			"Assigned Slave": item["assigned_slave"],
+			"Public Address": item["public_address"] + ":" + item["game_port"],
+			"Assigned Slave": item["assigned_slave_name"],
+			"Assigned Slave ID": item["assigned_slave"],
```
Danielv123, line 23: We probably don't need to show assigned_slave when we have assigned_slave_name

test/integration/index.js
```diff
@@ -134,7 +134,7 @@ before(async function() {
 	await exec("node ../../packages/master config set master.auth_secret TestSecretDoNotUse");
 	await exec("node ../../packages/master config set master.http_port 8880");
 	await exec("node ../../packages/master config set master.https_port 4443");
-	await exec("node ../../packages/master config set master.tls_bits 1024");
+	await exec("node ../../packages/master config set master.tls_bits 2048");
```
Danielv123, line 137: Is there any reasoning behind this?

Danielv123 (approved):

test/integration/index.js
yw9381: in my local server, run build script, use TLS 1024bit key and after that i got some error
```
node ../../packages/ctl slave create-config --id 4 --name slave --generate-token
master: [info] Listening for HTTP on port 8880
master: [fatal] 
master: +---------------------------------------------------------------+
master: | Unexpected error occured while starting master, please report |
master: | it to https://github.com/clusterio/factorioClusterio/issues   |
master: +---------------------------------------------------------------+
master: Error: error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small
master:     at Object.createSecureContext (_tls_common.js:129:17)
master:     at Server.setSecureContext (_tls_wrap.js:1328:27)
master:     at Server (_tls_wrap.js:1186:8)
master:     at new Server (https.js:70:14)
master:     at Object.createServer (https.js:94:10)
master:     at startServer (/root/factorioClusterio/packages/master/master.js:193:738)
```
build script
```bash
#!/bin/bash
npx lerna bootstrap --no-ci
node packages/lib/build_mod --source-dir packages/slave/lua/clusterio_lib --output-dir temp/test/sharedMods
proxychains -f /etc/proxychains.conf curl -o temp/test/sharedMods/subspace_storage_1.99.8.zip -L https://github.com/clusterio/factorioClusterioMod/releases/download/1.99.6/subspace_storage_1.99.8.zip
npx lerna run build
npm run-script ci-cover
codecov -f coverage/lcov.info
npm run-script lint
```

packages/web_ui/src/components/InstancesPage.jsx
yw9381: yep, but I think `assigned_slave` should still be exposed to facilitate troubleshooting in some cases

Hornwitser: It's already exposed in the instance view so I see no reason to expose it here. The web interface doesn't generally show IDs because they are long random numbers.

packages/lib/link/messages.js
Hornwitser: The IDs are resolved to names on the control client by querying the slave list and looking it up. The reasoning is that I eventually want to support pushing instance and slave status updates to the  control client (which includes name changes) so it would be easier convert the system to this if the control client did the ID to name translation itself.

packages/master/master.js
```diff
@@ -259,6 +260,17 @@ function validateSlaveToken(req, res, next) {
 	next();
 }
 
+function findInstancesAssignedSlave(assigned_slave_id) {
```
psihius, line 263: While on 5-10-20 slave instances this is fast and even 50 probably would be barely noticeable, if this is called in a lot of places (and other places in the code do the pattern), this will start to pile up as latency.
This should be a hash-map (associative array) lookup, not just looping over an array.

packages/master/master.js
```diff
@@ -690,12 +702,22 @@ class ControlConnection extends BaseConnection {
...
 				assigned_slave: instance.config.get("instance.assigned_slave"),
+				assigned_slave_name: null,
+				game_port: instance.config.get("factorio.game_port"),
+				public_address: null,
 				status: instance.status,
-			});
+			};
+			// get assigned slave info
+			if (instance.config.get("instance.assigned_slave") !== null) {
+				let assigned_slave_info = findInstancesAssignedSlave(instance.config.get("instance.assigned_slave"));
```
psihius, line 716: And here we are calling this in a loop - so you get a loop inside a loop lookup, which makes it worse - this is an (O)^2 complexity issue.

test/integration/index.js
psihius, line 137: Probably a good idea to have it as a config value overall.

psihius (changes requested):

packages/master/master.js
Hornwitser: I think you should be more careful about looking in the mapping here it's not for certain that the slave an instance is assigned to exists just because it's not null.

Also the variable name `assigned_slave_info` does not follow the camelCase convention.

packages/master/master.js
Hornwitser: This function is entirely redundant. `db.slaves` is a mapping of slave id to slave info objects. A call to `findInstancesAssignedSlave(...)` can be replaced with `db.slaves.get(...)`.

test/integration/index.js
Hornwitser: I don't like this as a solution, 2048 bit keys took a really long time to generate last I checked. After a lot of digging I've found that setting the OPENSSL_CONF environment variable to the absolute path to a .cnf file with the following content
```
openssl_conf = openssl_init

[openssl_init]
ssl_conf = ssl_sect

[ssl_sect]
system_default = system_default_sect

[system_default_sect]
CipherString = DEFAULT:@SECLEVEL=0
``` 
should allow certificates with 512 bit RSA keys to be used. Either way this change doesn't belong in this PR.

Hornwitser (changes requested):

test/integration/index.js
yw9381: in my VM Debian 10, the ssl default conf is `DEFAULT@SECLEVEL=2`, i think change `openssl.cnf` the system configuration is not a good solution, Because modifying the ssl configuration may affect some other aspects of the system
```ini
[system_default_sect]
MinProtocol = TLSv1.2
CipherString = DEFAULT@SECLEVEL=2
```
i write ssl generate script to test need time，the tested virtual machine configuration is as follows, use SATA SSD hard driver
```bash
➜  cat /proc/cpuinfo
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 63
model name      : Intel(R) Xeon(R) CPU E5-2678 v3 @ 2.50GHz
stepping        : 2
microcode       : 0x43
cpu MHz         : 2499.998
cache size      : 30720 KB
physical id     : 0
siblings        : 1
core id         : 0
cpu cores       : 1
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 15
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon nopl xtopology tsc_reliable nonstop_tsc cpuid pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm cpuid_fault invpcid_single pti ssbd ibrs ibpb stibp fsgsbase tsc_adjust bmi1 avx2 smep bmi2 invpcid xsaveopt arat md_clear flush_l1d arch_capabilities
bugs            : cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs itlb_multihit
bogomips        : 4999.99
clflush size    : 64
cache_alignment : 64
address sizes   : 43 bits physical, 48 bits virtual
power management:

➜  uname -a
Linux debian 4.19.0-8-amd64 #1 SMP Debian 4.19.98-1+deb10u1 (2020-04-27) x86_64 GNU/Linux

➜  free -m
              total        used        free      shared  buff/cache   available
Mem:            986          83         420           4         482         753
Swap:          1021           0        1021

➜  df -h
文件系统        容量  已用  可用 已用% 挂载点
udev            479M     0  479M    0% /dev
tmpfs            99M  4.3M   95M    5% /run
/dev/sda1        39G  1.7G   35G    5% /
tmpfs           494M     0  494M    0% /dev/shm
tmpfs           5.0M     0  5.0M    0% /run/lock
tmpfs           494M     0  494M    0% /sys/fs/cgroup
tmpfs            99M     0   99M    0% /run/user/0

➜  dd if=/dev/urandom of=/root/1g.bin bs=1M count=1024
记录了1024+0 的读入
记录了1024+0 的写出
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 5.67209 s, 189 MB/s

➜  dd if=/root/1g.bin of=/dev/null
记录了2097152+0 的读入
记录了2097152+0 的写出
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 3.74378 s, 287 MB/s

➜  node -v
v12.20.0
```
test script， generate_ssl_cert is `packages/master/src/generate_ssl_cert.js`:
```javascript
const generateSslCert = require("./generate_ssl_cert");

let tls_cert = "./cert.crt";
let tls_key = "./cert.key";

generateSslCert({
	bits: 2048,
	sslCertPath: tls_cert,
	sslPrivKeyPath: tls_key,
	doLogging: true,
});
```
this script output info is 
```bash
➜  time node index.js
[winston] Attempt to write logs with no transports {"message":"Generating SSL certificate with 2048 bits at ./cert.crt","level":"info"}
[winston] Attempt to write logs with no transports {"message":"Generated 2048bit keypair in: 135.286294 ms","level":"info"}
[winston] Attempt to write logs with no transports {"message":"Generated certificate in: 201.285105 ms","level":"info"}
node index.js  0.48s user 0.04s system 99% cpu 0.519 total
```
I think it takes about 0.4 seconds to generate the certificate is not very slow, and only need to generate the certificate once

packages/master/master.js
yw9381: i will be change to `db.slaves.get(...)`

packages/master/master.js
yw9381: click Add instance button and this instance unassigned any slaves, `instance.config.get("instance.assigned_slave")` return `null`, so I use `instance.config.get("instance.assigned_slave") !== null` to judge
11f09008-0506-4f77-96a2-2a15c25a6414

packages/web_ui/src/components/InstancesPage.jsx
yw9381: so do you think that only the `slave_name` should be displayed in the web interface instead of the `slave_id` again?

test/integration/index.js
Hornwitser: How weird. When I changed the keysize down from 2048 it made the build close a minute shorter and now it doesn't? What changed hmm?

Note that I didn't mention touching the system's openssl.cnf file. You would place that into a separate file and run node with OPENSSL_CONFIG=/path/to/my.cnf

packages/web_ui/src/components/InstancesPage.jsx
Hornwitser: Yes, long random numbers don't belong in the web interface. It could be shown in a details page for each slave, but that's a later problem.

yw9381: ok, i deleted `assigned_slave`  from WebUI column and push commit

test/integration/index.js
yw9381: i think is possible that the computing resources provided by the container used for construction are limiting the CPU usage. So it takes more time when the key bit is 2048

Hornwitser: This change will conflict with the removal of the automatic creation of a TLS certificate in #352.

packages/lib/link/messages.js
Hornwitser: `assigned_slave` is used in the InstanceViewPage so you can't just remove it from here without breaking something.

packages/master/master.js
```diff
@@ -690,12 +691,22 @@ class ControlConnection extends BaseConnection {
...
+				game_port: instance.config.get("factorio.game_port"),
+				public_address: null,
 				status: instance.status,
-			});
+			};
+			// get assigned slave info
+			if (instance.config.get("instance.assigned_slave") !== null) {
+				let assigned_slave_info = db.slaves.get(instance.config.get("instance.assigned_slave"));
+				push_info.assigned_slave_name = assigned_slave_info.name;
+				push_info.public_address = assigned_slave_info.public_address;
```
Hornwitser, line 707: `assigned_slave_info` may be `undefined` here if for example `database/slaves.json` is deleted, rendering the cluster unusable.

Hornwitser (changes requested):
The tests are failing, and the code changes has made the application unusable. You should run `npm run test` before you push changes to make sure it at least functions.

packages/master/master.js
yw9381: i think `assigned_slave_info` defalut is `undefined` is not problem, if player create new instance and don't assgin any slave, the `assigned_slave_info` is`undefined`, in WebUI, `public_address` and `assigned_slave_name` is display `undefined`, but instance must be assigned a slave, after the assign to any slave, `assigned_slave_info` can get slave data

Hornwitser: If `assigned_slave_info` is `undefined` then `assigned_slave_info.name` throws `TypeError: Cannot read property 'name' of undefined`.

yw9381: `if (instance.config.get("instance.assigned_slave") !== null) {` determine whether the instance is assigned to a slave, so  don't worry about this problem

Hornwitser: There is no guarantee that the id returned by `instance.config.get("instance.assigned_slave")` is valid, and it's easy enough to get into a situation where it's invalid by deleting database/slave.json.

Hornwitser:
Closed for going stale.

Issue #333, opened 2020-10-28
Add option for bind address


Hornwitser:
There's no option for selecting which address to bind to on the master server which is useful for setting it up behind a reverse proxy.  Should be fairly simple to add an optional `master.bind_address` config option, pass that as the third argument to `listen()`, and document setting it to localhost as recommended when proxying in README.md.

PR #349, opened 2020-12-27
This adds a bind_address option 

tovernaar123:
the bind_address was request by #333

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -31,6 +31,11 @@ of this and master.http_port needs to be set to a port.
 
 Defaults to 8443.
 
+### master.address_bind
+Address to bind the master server on. If set to null the master server will be bound to 
+all interfaces. Has to be set to a valid interface. 
+
+Defaults to null.
 
 ### master.external_address
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -31,11 +31,12 @@ of this and master.http_port needs to be set to a port.
 
 Defaults to 8443.
 
-### master.address_bind
-Address to bind the master server on. If set to null the master server will be bound to 
-all interfaces. Has to be set to a valid interface. 
-
-Defaults to null.
+### master.bind_address
+IP address to bind the HTTP and HTTPS ports on.  Useful to limit which
+interface to accept connections from.  If set to null the unspecified
+IPv6 address will be used, which on most systems means it will also
+listen on the unspecified IPv4 address and accept connections from all
+interfaces.
 
 ### master.external_address
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -38,6 +38,9 @@ IPv6 address will be used, which on most systems means it will also
 listen on the unspecified IPv4 address and accept connections from all
 interfaces.
 
+Defaults to null.
+
+
 ### master.external_address
 
 External address the master server is accessible on.  Currently only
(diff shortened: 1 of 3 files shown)


docs/configuration.md
```diff
@@ -1,401 +1,406 @@
-Configuration
-=============
-
-
-Master Configuration
---------------------
-
-### master.database_directory
```
Hornwitser, line 8: Please don't change the line endings for files. Go into your editor and set the EOL or end of line convention of this file to Unix/LF and save it.

docs/configuration.md
```diff
@@ -1,401 +1,406 @@
...
+server.  If set to null no HTTPS server will be exposed.  At least one
+of this and master.http_port needs to be set to a port.
+
+Defaults to 8443.
+
+### master.address_bind
+Address to bind the master server on. If set to null the master server will be bound to 
+all interfaces. Has to be set to a valid interface. 
+
+Defaults to null.
```
Hornwitser, line 38: The spacings is not consistent with all of the other options in the file. There's should be two blank spaces before the headings and one blank space after it. There should also be two spaces after periods and the lines are wrapped to 72 characters.

I don't think the description here is saying the right thing. I suggest changing it to:
```
IP address to bind the HTTP and HTTPS ports on.  Useful to limit which
interface to accept connections from.  If set to null the unspecified
IPv6 address will be used, which on most systems means it will also
listen on the unspecified IPv4 address and accept connections from all
interfaces.
```

packages/lib/config/definitions.js
```diff
@@ -33,6 +33,13 @@ MasterGroup.define({
 	optional: true,
 	initial_value: 8443,
 });
+MasterGroup.define({
+	name: "address_bind",
+	title: "Address bind",
+	description: "The address to bind the master https/http server to.",
```
Hornwitser, line 39: Suggested rewording: `IP address to bind the HTTP and HTTPS ports on.`

packages/master/master.js
```diff
@@ -1943,6 +1943,7 @@ async function startServer() {
 	// Make sure we're actually going to listen on a port
 	let httpPort = masterConfig.get("master.http_port");
 	let httpsPort = masterConfig.get("master.https_port");
+	let bindIp = masterConfig.get("master.address_bind");
```
Hornwitser, line 1946: If you add `|| ""` after the `.get()` it becomes an empty string when null and you don't need the if statements further down. The variable should also be named the same as the option (like the ones above it).

Hornwitser (changes requested):
Why name the option `address_bind` instead of `bind_address`? Looking at other server software "bind address" seems more common.

docs/configuration.md
tovernaar123: yeah sorry

docs/configuration.md
tovernaar123: :P that is better yes

packages/master/master.js
tovernaar123: okey

packages/lib/config/definitions.js
tovernaar123: done

Hornwitser (approved):

Danielv123 (approved):

psihius (approved):
Would be nice to have tests cover this.
I know, it's trivial, but I would like us to get into the habit of making sure new stuff is covered by tests. You never know, a change down the line in 2 years might introduce a bug and test will catch it.
I'm not saying it's a must here but would be nice.

PR #355, opened 2021-01-29
Adds live updates to slave 

tovernaar123:
live updates from adding slaves to the webgui


packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -1,28 +1,46 @@
-import React, { useContext } from "react";
+import React, { useContext, useState } from "react";
 
 import libLink from "@clusterio/lib/link";
 
 import DataTable from "./data-table";
 import ControlContext from "./ControlContext";
 import PageLayout from "./PageLayout";
-
+import { copySync } from "fs-extra";
```
Hornwitser, line 8: This is nonsense. It's not used, and in the web code `fs-extra` is an empty package.

packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -1,28 +1,46 @@
...
-		return result["list"].map((item) => ({
+		result = result["list"].map((item) => ({
 			key: item["id"],
 			"Name": item["name"],
 			"Agent": item["agent"],
 			"Version": item["version"],
 			"Connected": item["connected"] && "Yes",
 		}));
+		return result;
+	}
```
Hornwitser, line 36: Is there a reason for this change?

packages/lib/link/messages.js
```diff
@@ -916,6 +916,22 @@ messages.debugWsMessage = new Event({
 	},
 });
 
+messages.liveUpdateSlaves = new Event({
+	type: "live_update_slaves",
+	links: ["master-control"],
+	eventProperties: {
+		item: {
```
Hornwitser, line 923: The convention in this file is that properties are put in `"quotes"` when they are present in the payload, and are not quoted when they are JSON Schema keywords. So item should be quoted here.

packages/master/master.js
```diff
@@ -1610,7 +1625,9 @@ async function handleHandshake(message, socket, req, attachHandler) {
 		}
 
 		logger.verbose(`SOCKET | registered slave ${data.id} version ${data.version}`);
-		slaveConnections.set(data.id, new SlaveConnection(data, connector));
+		let slave = new SlaveConnection(data, connector);
+		slaveConnections.set(data.id, slave);
+		db.slaves.updateGui(data.id);
```
Hornwitser, line 1630: I don't see the reason for changing this line. You don't use `slave` in a new place that necessitate it being store in a variable. The name is also misleading.

packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -1,28 +1,46 @@
-import React, { useContext } from "react";
+import React, { useContext, useState } from "react";
```
Hornwitser, line 1: The import added here is unused.

Hornwitser (changes requested):
There's several issues with this implementation. In no particular order:
- The permissions system is ignored. Pretty much every other piece of information you can get out of the control connection requires some kind of permission to be granted to the user trying to get the information for it to work. For example to list the slaves in the cluster with the listSlaves request the `core.slave.list` permission needs to have been granted. There's no permission controlling the liveUpdatesSlaves event, it's sent to everyone.
- The liveUpdatesSlaves event is sent even if the receiving party doesn't want it. It's not a lot of traffic by itself, but if every piece of data that can update in real time is sent to every connection it's going to add up quickly. Especially in the sort of 100 player+ clusters that Clusterio 2.0 aims to handle.
- The update logic implemented in the web interface is not usable by multiple components at the same time. This is problematic if you want to say have the slave list available in two places in the web interface at the same time. This will happen in the InstanceViewPage if the assigned slave name and the list of slaves to assign are both to be updated in real time.

If you take a look at how log lines are added to the InstanceConsole component in real time you'll notice that it doesn't have any of these issues. It registers a handler with the Control class which triggers on a new log line, the Control class then sends a request to the master server to subscribe to log messages from the given instance, this request will be denied if the user doesn't have the permission to subscribe to log messages. It ensures log messages aren't sent unless they are wanted, and multiple components can subscribe to log messages and they will all receive updates.

Finally `update_slaves`, `live_update_caller`, `add_row`, `save_table_ref`, `new_table` does not follow the `camelCase` naming convention.

tovernaar123:
It shood now only send the data to the people with the correct permisions and only send it to the gui

psihius (approved):
Changes seem fine, but I don't see tests :)

packages/master/master.js
```diff
@@ -695,6 +715,7 @@ class ControlConnection extends BaseConnection {
 	async setMasterConfigPropRequestHandler(message) {
 		let { field, prop, value } = message.data;
 		masterConfig.setProp(field, prop, value, "control");
+
```
Hornwitser, line 718: Another random newline.

packages/master/master.js
```diff
@@ -2006,6 +2029,15 @@ async function startServer(args) {
...
 
 	db.slaves = await loadMap(masterConfig.get("master.database_directory"), "slaves.json");
+	// live updating
+	db.slaves.updateGui = function(key) {
+		for (let control of controlConnections) {
+			if (control.doLiveUpdateSlaves) {
+				control.updateSlaves(key);
+			}
+		}
+	};
```
Hornwitser, line 2039: Once again you're modifying an instance of a built-in Map class. The type of `db.slaves` is `Map`, not `MapWhichMayHaveAnUpdateGuiMethod`. This should be a stand alone function.

Also this is inefficient. You do the work of retrieving the slave record and constructing the update for every control connection instead of doing the work once and sending the data to each control connection. The code in ControlConnection.updateSlaves could be moved outside the for loop here and then the message with the computed item sent to each connection inside the loop.

packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -1,15 +1,27 @@
-import React, { useContext } from "react";
+import React, {useContext} from "react";
```
Hornwitser, line 1: There's no linting rule for this, but every other import in the project has spaces inside the `{ brackets }`. Also this is a random unrelated change.

packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -1,15 +1,27 @@
...
+			key: item["id"],
+			"Name": item["name"],
+			"Agent": item["agent"],
+			"Version": item["version"],
+			"Connected": item["connected"] && "Yes",
+		};
+		if (table) {
+			table.addRow(row);
+		}
+	});
```
Hornwitser, line 24: Modifying state directly inside a component function is a strict no-no in React. The rules for component functions is that the body may not have any side effects, here you modify the control instance in the body of the SlavesPage component. Side effects must be put inside a `useEffect` hook.

This also leaks the handler updating the table (it's not removed when the component is unmounted), and does not permit multiple instances of the component to exist at the same time as the "slaves_page" identifier is a singleton, which is bad.

packages/web_ui/src/util/websocket.jsx
```diff
@@ -119,6 +123,26 @@ export class Control extends libLink.Link {
...
+		}
+		this.liveUpdateSlaveHandlers.set(id, handler);
+	}
+
+	async offLiveSlaveAdded(id) {
+		this.liveUpdateSlaveHandlers.delete(id);
+		if (this.liveUpdateSlaveHandlers.size === 0) {
+			libLink.messages.setLiveSlaveSubscription.send(this, { connect: false });
+		}
+	}
```
Hornwitser, line 138: Why do you identify the handler with an arbitrary `id`? This is different from how the onInstanceLog and offInstanceLog works and makes it more difficult to make components properly add and remove their handlers from this event.

Also logic is missing to re-establish the slaveUpdateSubscription after a reconnect event like there is for the log handlers (see the connect event handler in the constructor).

packages/web_ui/src/util/websocket.jsx
```diff
@@ -119,6 +123,26 @@ export class Control extends libLink.Link {
...
+
+	async liveUpdateSlavesEventHandler(message) {
+		for (let [_, handler] of this.liveUpdateSlaveHandlers.entries()) {
+			handler(message);
```
Hornwitser, line 142: Use the `.values()`/`.keys()` methods of a map if you only need to iterate the value/key.

test/integration/clusterio.js
```diff
@@ -1,14 +1,39 @@
+/* eslint-disable no-console */
```
Hornwitser, line 1: With a handful of exceptions there should be no use of console in the project, this is one of the files it shouldn't be used in.

test/integration/clusterio.js
```diff
@@ -1,14 +1,39 @@
...
+			console.log(name, line);
+		});
+		let stderr = new server._LineSplitter((line) => { console.log(name, line.toString("utf8")); });
+		process.stdout.on("data", chunk => { stdout.data(chunk); });
+		process.stdout.on("close", () => { stdout.end(); });
+		process.stderr.on("data", chunk => { stderr.data(chunk); });
+		process.stderr.on("close", () => { stderr.end(); });
+	});
+}
+
```
Hornwitser, line 36: Do not copy and paste code like this. Just import it from `./index`. 😞

test/integration/clusterio.js
```diff
@@ -32,6 +57,35 @@ describe("Integration of Clusterio", function() {
...
+				slowTest(this);
+				let testPromise = new Promise((resolve, reject) => {
+					let control = getControl();
+					control.onLiveSlaveAdded("testing", async function() {
+						resolve();
+					});
+				});
+				if (await fs.exists("./temp/test/liveUpdates/slave-config.json")) {
+					await fs.unlink("./temp/test/liveUpdates/slave-config.json");
+				}
```
Hornwitser, line 72: Use `fs.remove` instead.

test/integration/clusterio.js
```diff
@@ -32,6 +57,35 @@ describe("Integration of Clusterio", function() {
...
+				/* eslint-disable max-len */
+				await execCtl("slave create-config --id 5 --name slave2 --generate-token --output ./liveUpdates/slave-config.json");
+				await exec("node ../../packages/slave config set slave.tls_ca ../../test/file/tls/cert.pem --config ./liveUpdates/slave-config.json");
+				slaveProcess = await spawn("slave2:", "node ../../packages/slave run --config ./liveUpdates/slave-config.json", /SOCKET \| registering slave/);
+				/* eslint-enable max-len */
```
Hornwitser, line 77: The max line length rule is there for a reason. Break up these lines into shorter ones.

test/integration/index.js
```diff
@@ -27,6 +28,19 @@ class TestControl extends libLink.Link {
...
+		if (this.liveUpdateSlaveHandlers.size === 0) {
+			libLink.messages.setLiveSlaveSubscription.send(this, { connect: true });
+		}
+		this.liveUpdateSlaveHandlers.set(id, handler);
+	}
+
+	async liveUpdateSlavesEventHandler(message) {
+		for (let [_, handler] of this.liveUpdateSlaveHandlers.entries()) {
+			handler(message);
+		}
```
Hornwitser, line 41: Same as previous, use `.values()` if you don't need the keys.

Hornwitser (changes requested):
I am not a fan of the names in use here. Is this event only about the slave list updating? If so the name should reflect that:
`setLiveSlaveSubscription` -> `setSlaveListSubscription`
`liveUpdateSlave` -> `slaveListUpdate`
`doLiveUpdatesSlaves` -> `sendSlaveListUpdates`
`updateSlaves` -> `updateSlaveList`
`updateGui` -> `updateSlaveList`
`onLiveSlaveAdded` -> `onSlaveListUpdate`
`liveUpdatesSlaveHandles` -> slaveListUpdateHandlers`
`offLiveSlaveAdded` -> `offSlaveListUpdate`

test/integration/clusterio.js
tovernaar123: i am using the spawn method from index.js which uses it

test/integration/clusterio.js
tovernaar123: yeah that would be better that also fixses the console.log thing kinda forgot i had this

Hornwitser:
Made my own implementation of this feature based on this PR.

Issue #356, opened 2021-02-02
[2.0] Stopping docker container doesn't trigger shutdown


furgin:
When stopping a container with 'docker stop' it sends a TERM signal which is not handled by the master process.

This results in not being shut down correctly and not saving changes to the configuration files.

PR #357, opened 2021-02-02
Added TERM signal handler

furgin:
Added a signal handler for SIGTERM

I'm not sure if needs the special handling for the second signal of this type.


furgin:
fixes #356

Hornwitser:
Pretty sure you'll run into the same problem with the slave process.

furgin:
Good call, ill make the change and test it out locally then update the PR

Danielv123 (approved):
Makes sense.

psihius (approved):

PR #360, opened 2021-02-24
Remote master config management

Hornwitser:
Adds commands and a web interface for modifying the master config remotely. Includes a necessary access control addition to the config system to avoid leaking the authentication secret which is stored in the master config.


psihius (commented):
Is it possible to use constants for the `master`, `slave` and `control` strings? Their usage is wide and having those as constants to avoid spelling errors and having the ability to easily search and refactor using IDE tools would be a great help.

Hornwitser:
Well, the issue is that would just make it it worse. If it were to be written like 

```
masterConfig.set(field, value, libConfig.CONTROL);
```

And you fat finger `CONTROL` it doesn't throw an error, it becomes `undefined`.  And with `set` as it is now, `undefined` means use the location of the config, which would grant full access to it. If you were to pass `"cntrol"` to `set` as the location instead it would throw on attempting to set any value since no field has that location in its access definition.

psihius (approved):

PR #362, opened 2021-03-01
Add plugin list

Hornwitser:
Implementation of a plugins page on the master server which displays which plugins are enabled, their versions and potential problems with them along with a details page containing some fields from package.json.

Also adds a basic version of the list as a command to clusterioctl.


psihius (approved):

Hornwitser:
psihius, you're not supposed to approve it if the Travis CI build fails.

psihius (changes requested):
Failing tests/builds need to be fixed :)

PR #363, opened 2021-03-05
MacOS compatibility

drouarb:
Hey,

I wanted to try to create a plugin for clusterio, unfortunately, there was no MacOS support, so I made one :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,19 +190,19 @@ up, and v10.16.0+.
 
 1. Open Terminal or Command prompt in the directory you want to install to and run the following commands.
 
-```bash
-mkdir clusterio
-cd clusterio
-npm init -y
-npm install @clusterio/master @clusterio/slave @clusterio/ctl
-```
+    ```bash
+    mkdir clusterio
+    cd clusterio
+    npm init -y
+    npm install @clusterio/master @clusterio/slave @clusterio/ctl
+    ```
 
 2. Obtain Factorio:
 
-Install Factorio and copy the factorio.app in your clusterio install folder as factorio.
+Install Factorio and copy the factorio.app/Contents in your clusterio install folder as factorio.
 If you installed factorio on your mac you can run
 ```bash
-cp /Applications/factorio.app/ factorio
+    cp -r /Applications/factorio.app/Contents/ factorio
 ```
 
 ## Installing Plugins
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ the [1.2.x branch][1.2.x] for instructions on how to install the stable version.
 * [Plugins](#plugins)
 * [Ubuntu setup](#ubuntu-setup)
 * [Windows setup](#windows-setup)
+* [MacOS setup](#macos-setup)
 * [Installing Plugins](#installing-plugins)
 * [Configure Master Server](#configure-master-server)
   * [Hosting Behind Proxy](#hosting-behind-proxy)
@@ -180,6 +181,29 @@ up, and v10.16.0+.
         3.  Go to the parent folder of the folder that Steam opened and copy
             the Factorio folder into the factorio folder created in step 1.
 
+## MacOS Setup
+
+**Requirements**
+
+download and install nodeJS 12 from http://nodejs.org or brew (`brew install node`).  Clusterio runs on Node.js v12 and
+up, and v10.16.0+.
+
+1. Open Terminal or Command prompt in the directory you want to install to and run the following commands.
+
+```bash
+mkdir clusterio
+cd clusterio
+npm init -y
+npm install @clusterio/master @clusterio/slave @clusterio/ctl
+```
+
+2. Obtain Factorio:
+
+Install Factorio and copy the factorio.app in your clusterio install folder as factorio.
+If you installed factorio on your mac you can run
+```bash
+cp /Applications/factorio.app/ factorio
+```
 
 ## Installing Plugins
 
--- a/packages/lib/factorio/server.js
+++ b/packages/lib/factorio/server.js
@@ -832,7 +832,7 @@ class FactorioServer extends events.EventEmitter {
 
 		await this._writeConfigIni();
 		this._server = child_process.spawn(
-			this.dataPath("..", "bin", "x64", "factorio"),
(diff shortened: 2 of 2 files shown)


Danielv123:
Well documented and clean code. But I thought macos had similar headless distributions to linux? Why does it have to use different paths?

drouarb:
There is no headless distribution on MacOS (or i didn't found it ?), it uses the full factorio game as windows but since we are on MacOS the binary path is standard

Hornwitser:
> Well documented and clean code. But I thought macos had similar headless distributions to linux? Why does it have to use different paths?

No, headless build has only ever been for Linux.

Danielv123 (approved):
Looks good to me then.

packages/lib/factorio/server.js
```diff
@@ -857,6 +857,20 @@ class FactorioServer extends events.EventEmitter {
...
+	 *
+	 * @return {string} path to factorio binary
+	 */
+	getBinaryPath() {
+		if (process.platform === "darwin") {
+			return this.dataPath("..", "MacOS", "factorio");
+		}
+		return this.dataPath("..", "bin", "x64", "factorio");
+	}
+
```
Hornwitser, line 873: Is there a rationale for placing this method here in the class, seems rather random in the middle, it's related to the dataPath so shouldn't it be after that? Also the documentation is missing a mention of it not being valid before `init()` is called.  Note that this class doesn't use the getThing setThing naming of methods, it should either be `binaryPath()`, a property getter for `binaryPath`, or set as a property during `init()`.

psihius (approved):
Seems good to me too

Hornwitser (changes requested):
This doesn't quite follow the code style of the class it's modifying. Does this pass all the tests on Mac? If this is all that's needed to support Mac that's pretty nice.

packages/lib/factorio/server.js
drouarb: I think there is no point setting an internal class variable for this, so if you prefer I can rename it to binaryPath(). Also, placement is rather random, I don't know how do you order the functions.

README.md
```diff
@@ -180,6 +181,29 @@ up, and v10.16.0+.
...
+npm install @clusterio/master @clusterio/slave @clusterio/ctl
+```
+
+2. Obtain Factorio:
+
+Install Factorio and copy the factorio.app in your clusterio install folder as factorio.
+If you installed factorio on your mac you can run
+```bash
+cp /Applications/factorio.app/ factorio
+```
```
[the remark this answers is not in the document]
drouarb: My bad, you are right.

drouarb:
Yes, tests are running on mac, also I am successfully running a small cluster on my mac

packages/lib/factorio/server.js
Hornwitser: The order is things that are logically related are loosely grouped together. I would put this function after `dataPath`.

PS: every other function in the whole project uses `@returns` instead of `@return`.

Hornwitser:
The subspace errors are likely from missing the subspace_storage_1.99.8.zip and clusterio_lib_0.1.2.zip mods in temp/test/sharedMods. The travis build copies them there, but you can safely ignore the errors.

test/integration/clusterio.js
```diff
@@ -158,7 +158,9 @@ describe("Integration of Clusterio", function() {
 					],
 					[
 						"allow_commands", true,
-						"allow-commands", "Allow Lua commands: Yes.",
+						"allow-commands", (process.platform === "darwin"
+							? "Yes"
+							: "Allow Lua commands: Yes."),
```
Hornwitser, line 163: Are you sure you're running these tests with the latest released version of Factorio? The output of this command recently changed and it makes no sense for the output to differ based on which platform Factorio was built for.

drouarb: I'm running in 1.1.19, though it was an MacOS stuff

Hornwitser:
I notice the Factorio distribution for Mac puts everything in a subfolder called Contents. This is different from how Windows and Linux are packaged and will break support for multiple installations. The following two directory layouts are supported for the configured `slave.factorio_directory`.

```
Single installation
-------------------
factorio/
 +- data/
 |  +- base/
 |  +- core/
 |  +- changelog.txt
 |  +- ...
 +- bin/ (or MacOS/)

Multi installation
------------------
factorio/
 +- 0.17.69/ (folder name doesn't matter)
 |  +- data/
 |  |  +- base/
 |  |  +- core/
 |  |  +- changelog.txt
 |  |  +- ...
 |  +- bin/ (or MacOS/)
 +- 1.0.0/
 |  +- data/
 |  |  +- base/
 |  |  +- core/
 |  |  +- changelog.txt
 |  |  +- ...
 |  +- bin/ (or MacOS/)
 +- 1.1.25/
    +- data/
    |  +- base/
    |  +- core/
    |  +- changelog.txt
    |  +- ...
    +- bin/ (or MacOS/)
```

Either you will have to update the detection logic to go into the `Contents` folder on Mac, or copy `/Applications/factorio.app/Contents` to `factorio` (so that all platforms gets a consistent layout and future changes doesn't break it).

Hornwitser:
Would you mind testing if setting `slave.factorio_directory` to `/Applications/factorio.app/Contents` works on MacOS? Just modify the line at around the 200 line mark in test/integration/index.js that says:
```js
await exec(`node ../../packages/slave config set slave.factorio_directory ${path.join("..", "..", "factorio")}`);
```
into
```js
await exec(`node ../../packages/slave config set slave.factorio_directory /Applications/factorio.app/Contents`);
```
And run the tests.

drouarb:
Just tried, everything still works fine :)

Hornwitser:
In that case I'd like the instructions to say copy /Applications/factorio.app/Contents to factorio/ and then have the changes to the integration tests removed. It would minimize the amount of things that are different between the platforms.

psihius:
@drouarb could we do the requested changes to this PR so we can wrap this up and merge? Alpha5 release is coming closer and closer :)

Hornwitser (approved):

psihius (approved):

PR #364, opened 2021-03-06
Fix windows test failure

Hornwitser:
Another set of fixes for failure of the tests on Windows. This involves the node-forge dependency being in the wrong place, and the exit logic not making sure the control connector is closed before exiting which was masked by the use of the `--exit` flag to mocha.


psihius (approved):
Seems logical, merge it

PR #370, opened 2021-03-17
Use a diff and author friendly markdown formatting

Hornwitser:
Simplify the the way markdown files are formatted to make them easier to author without editors that do auto line breaking, and make the diffs from them simpler to read.

---

This is something I've thought about for some time now. Basically having markdown files that look pretty to read makes them more difficult to edit and makes the diffs harder to read. So the idea here is make the format easy to edit and author and stop caring about the source looking pretty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,64 +1,48 @@
-Changelog
-=========
+# Changelog
 
-Version 2.0.0
--------------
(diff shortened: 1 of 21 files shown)


psihius:
Looks nice, should help with editing how you said. Merging it.

PR #379, opened 2021-05-17
Small things 4

Hornwitser:
Small fixes an improvement that doesn't make sense to put into separate PRs:
- Tiny change to factorio lib which makes it able to start and stop 0.16 servers.
- Removal of log rotation which is not needed any more.
- Switch over from deprecated linter rules to the plugin they were moved to.
- Fix spidertron serialization of equipment grids
- Support symbolic links in modules folder
- Fix log query broken after the master refactor
- Remove PluginsContext as the plugins are already available in the ControlContext.


packages/slave/slave.js
```diff
@@ -467,7 +453,7 @@ class Instance extends libLink.Link {
 		// Find stand alone modules to load
 		// XXX for now only the included clusterio module is loaded
 		for (let entry of await fs.readdir(path.join(__dirname, "modules"), { withFileTypes: true })) {
-			if (entry.isDirectory()) {
+			if (!entry.isFile()) {
```
psihius, line 456: Is the reversed check due to symlinks and stuff like that? :)

psihius (approved):

packages/slave/slave.js
Hornwitser: Yes.

psihius:
Merge it

PR #380, opened 2021-06-01
Save management

Hornwitser:
Implementation of save management. The following features are provided
- Create save with map options.
- Load scenario with map options.
- List saves.
- Start server with a given save.
- Upload save.
- Download save.
- Delete save.
- Copy save on start if it's an autosave.
- Update subscriptions to instances status changes for the web interface.
- Update subscriptions to save list changes for the web interface.


psihius (approved):
Looks good, didn't find anything that jumped out at me.

PR #382, opened 2021-06-02
Add slave update subscriptions

Hornwitser:
Allow the web interface to listen for updates to the slave list and apply it live to the places where slaves are shown in the web interface.

This is a working implementation of what #355 set out to do.


Danielv123 (approved):
Looks good. My experience is that this will be very useful for the new interface I created for the slaves to make the lua testing a bit easier :)

PR #384, opened 2021-06-09
Fix tests on Windows and remove internet requirement to pass

Hornwitser:
Fix for HttpCloser test on Windows and the requirement of having internet to for the tests to pass.


Danielv123 (approved):
Thanks!

PR #390, opened 2021-06-15
Alpha 7 preparations

Hornwitser:
The last necessary bit for making the planned installer flow work with the web and ctl interface is a way to generate a slave token for a random slave id. This implements that and some minor bug fixes. After this I'll push to npm and do testing on the installer.


Danielv123 (approved):

Danielv123:
Looks all good to me. Not having to specify an ID is useful as it seemed like a needlessly complicated step before.

PR #395, opened 2021-06-18
Rename repository

Hornwitser:
Having the repositories be named factorioClusterio and factorioClusterioMod is something that's irked me since I started on this project. It's Clusterio, no more no less. The mod was also rebranded to subspace_storage long ago so it's due for a rename too.


psihius (approved):

Danielv123 (approved):
Merge time?

PR #408, opened 2021-08-31
Updated Service Info

antt1995:
Included Instructions for service

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,7 @@ Otherwise see below for OS specific instructions.
 
     To specify a version of Factorio to download replace "latest" in the URL with a version number like "1.0.0".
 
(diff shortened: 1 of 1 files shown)


Hornwitser (changes requested):
It was intentional that no instructions were given on how to run via systemd before the install plugins and configure master server steps, please move the instructions for running via systemd to to the Running Clusterio section.

Also the extensions on `run-master` and `run-slave` are intentionally left out because they differ between Linux (.sh) and Windows (.cmd).

antt1995:
Ok Updated as requested

Hornwitser (approved):

psihius (approved):

PR #409, opened 2021-09-04
Allow automatic Factorio install during create on Linux

### LINKED ISSUE #330: Automatically download Factorio headless on Linux
This would be useful for Linux installations.  Download the headless version of Factorio and use that if the selected version is not installed on the slave.

justchen1369:
related to #330

### DIFF
--- a/packages/create/create.js
+++ b/packages/create/create.js
@@ -2,6 +2,7 @@
 
 "use strict";
 const child_process = require("child_process");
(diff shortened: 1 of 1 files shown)


psihius (approved):

psihius:
Adding a test for this functionality would be a welcome addition. Otherwise we add and add and then wonder why things break due to absence of tests :D

PR #411, opened 2021-09-07
Rework inventory_sync

Hornwitser:
Complete rework of the inventory synchronization to reduce the number of times the inventory has to be downloaded and the possibilities for item loss/duplication.

The basic concept is to acquire exclusive access to the player data before downloading the data.  This prevents the player from logging in to two servers at the same time and getting the synced inventory twice. The acquisition will in most cases execute faster than the player can provide input to the game on login, and if the player data on the master isn't what was previously uploaded then the player inventory is deleted and the player is turned into a spectator while the updated inventory is being transferred into the game.  This model allows for the downloading of the synced inventory to be skipped in the case the player joins the same server they left from.

The synced data is expanded to include more than just the character and inventory to properly handle sync of dead players.  If the player leaves dead on one server this state will be saved and when the player joins another server they continue off as dead.  The ghost, god, and spectator controllers are supported, while the cutscene and editor controllers are not.  For players in a cutscene when joining the acquisition will be attempted immediately, but the scenario waits until the player is out of the cutscene before acting upon the response to the acquisition.  For players in the editor mode the scenario toggles them out of the mode before syncing data.

Since uploads to the master can fail handling of this is also added. The scenario will retry uploading the player data at exponentially increasing intervals until it gets the message back that the upload succeeded.  Since it's possible for servers to go down and stay down for a long time the exclusive access held by these servers are released after a timeout.  This means that a player could be playing and or have uploaded newer data by the time the instance comes back online and tries to upload the inventory, in this case the master will report the upload succeeded but discard the data.  Instances may also upload old data if they load back to an earlier save, which will also be discarded by the master.

Other small improvements made by this rework:
- Add a retry button to the failed download dialog
- Update the existing progress dialog instead of destroying and recreating it every time.
- Automatically stuff the items from the temporary inventory into the synced inventory if they fit.
- Move player back to the position they started the download from if they were turned into a spectator for the download.
- Ensure the player has a character if the stored data is wiped.
- Add error checking of the RCON commands sent by the instance.

---

Known bugs:
- You lose your crafting queue if you leave and join the same server.
- You lose your crafting queue if the first upload fails.


plugins/inventory_sync/module/define_player_inventories.lua
```diff
@@ -1,10 +1,11 @@
 local inventories = {
-	defines.inventory.character_main,
-	defines.inventory.character_guns,
-	defines.inventory.character_ammo,
-	defines.inventory.character_armor,
-	defines.inventory.character_vehicle,
```
psihius, line 6: Why the vehicle gone?

plugins/inventory_sync/module/define_player_stat_keys.lua
```diff
@@ -14,10 +14,4 @@ return {
 	"character_maximum_following_robot_count_bonus",
 	"character_health_bonus",
 	"character_personal_logistic_requests_enabled",
-	"build_distance",
```
psihius, line 17: wh not sync these any more? Admins can modify the stats of the users via lua, so they might be different than calculated. Also, RPG mods? Or is this somehow synced another way?

Hornwitser: If you read the docs the reason is obvious [these properties are read only](https://lua-api.factorio.com/latest/LuaControl.html#LuaControl.build_distance), you cannot set them, and trying to do so throws an error.

plugins/inventory_sync/module/define_player_inventories.lua
Hornwitser: The "character_vehicle" inventory doesn't actually exist, so there's no point in trying to sync it. If it did exist though, what would it be?

psihius: Well, technically if the user is in a vehicle? :D But okay, then the question would be why it was there in the first place :D

Danielv123: It was there because it was late at night and I just copied over everything that seemed relevant from the API 🤣

psihius (approved):

plugins/inventory_sync/module/define_player_inventories.lua
Hornwitser: Well, if the user is in a vehicle we don't want to copy the vehicle inventory, as the vehicle stays behind. But even if you're in a vehicle getting the `charecter_vehicle` inventory still returns nil.

PR #420, opened 2021-10-31
Slave page

Hornwitser:
Add per slave page to the web UI displaying the console for the slave and the instances assigned to it. Also adds console view to the master page.

### DIFF
--- a/packages/web_ui/src/components/SlaveViewPage.jsx
+++ b/packages/web_ui/src/components/SlaveViewPage.jsx
@@ -3,6 +3,7 @@ import { Descriptions, PageHeader, Spin, Typography } from "antd";
 import { useParams } from "react-router-dom";
 
 import InstanceList from "./InstanceList";
+import LogConsole from "./LogConsole";
 import { useAccount } from "../model/account";
 import { useInstanceList } from "../model/instance";
 import { useSlave } from "../model/slave";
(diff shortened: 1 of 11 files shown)


psihius (approved):

packages/web_ui/src/components/SlaveViewPage.jsx
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	return <PageLayout nav={nav}>
+		<PageHeader
+			className="site-page-header"
+			title={slave.name || slaveId}
+		/>
+
+		<Descriptions bordered size="small" column={{ xs: 1, sm: 2, xl: 4 }}>
+			<Descriptions.Item label="Name">{slave["name"]}</Descriptions.Item>
```
Danielv123, line 46: Is there a reason why you use bracket syntax here instead of . syntax?

Danielv123 (approved):
Pretty much the exact same functionality as I have in my local branch. Can't believe I have forgotten to make a PR for it for like 3 months now... This was cleaner anyways though.

psihius:
> Pretty much the exact same functionality as I have in my local branch. Can't believe I have forgotten to make a PR for it for like 3 months now... This was cleaner anyways though.

This is why you make "in progress" pull requests :D

packages/web_ui/src/components/SlaveViewPage.jsx
Hornwitser: It's a hint that it's a plain object, like deserialized JSON.

Danielv123: Ah, I see. I will try to keep to the same pattern then.

PR #421, opened 2021-11-03
Use safe file writing pattern

### LINKED ISSUE #416: Use the safe file writing pattern
The current method for writing data to files in Clusterio is to open the target file for writing (truncating it) and then writing the new data to this file. This is unsafe and can easily lead to complete loss of the content of the target file if the Node.js process crashes or is killed before the data is fully written to the target file. This has happened multiple times, and continues to be a major risk to data stored by Clusterio.

The common pattern for writing new data for a file to disk safely is to write all of the new data to a temporary file and once flushed to disk rename the temporary file into the target file. The replacement via rename is an atomic operation and means the target file is either fully written to disk or not modified at all.

Replace all locations Clusterio truncates and writes directly to a target file with the safe writing pattern. The files of importance includes (but are not limited to) all files in `master.database_directory`, all config files, settings written to instance directories, save patcher, and upload of locale/icon export.

Hornwitser:
Write new file contents to a temporary file and then rename over the
target file to avoid data loss in the unlikely event there's an error
writing the new file contents or the process is interrupted.

Resolves #416.

### DIFF
--- a/packages/create/create.js
+++ b/packages/create/create.js
@@ -85,6 +85,12 @@ class LineSplitter extends stream.Transform {
 class InstallError extends Error { }
 
 
+async function safeOutputFile(file, data, options={}) {
+	let temporary = `${file}.tmp`;
+	await fs.outputFile(temporary, data, options);
+	await fs.rename(temporary, file);
+}
+
 async function execFile(cmd, args) {
 	logger.verbose(`executing ${cmd} ${args.join(" ")}`);
 	const asyncExec = util.promisify(child_process.execFile);
@@ -165,7 +171,7 @@ async function validateInstallDir() {
 
 async function installClusterio(mode, plugins) {
 	try {
-		await fs.writeFile("package.json", JSON.stringify({
+		await safeOutputFile("package.json", JSON.stringify({
 			name: "clusterio-install",
 			private: true,
 		}, null, 2), { flag: "wx" });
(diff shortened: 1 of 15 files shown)


packages/create/create.js
```diff
@@ -85,6 +85,12 @@ class LineSplitter extends stream.Transform {
 class InstallError extends Error { }
 
 
+async function safeOutputFile(file, data, options={}) {
```
psihius, line 88: Why is it defined here, can't it be just imported from the file_ops like it is for the rest of the files?

packages/lib/file_ops.js
```diff
@@ -69,7 +69,29 @@ async function findUnusedName(directory, name, extension = "") {
...
+ *
+ * If the operation fails it may leave behind the temporary file.  This
+ * should not be too much of an issue as the next time the same file is
+ * written the temporary will be overwritten and renamed to the target file.
+ *
+ * @param {string} file - Path to file to write.
+ * @param {string|Buffer} data - Content to write.
+ * @param {object|string} options - see fs.writeFile, `flag` must not be set.
+ */
+async function safeOutputFile(file, data, options={}) {
```
psihius, line 87: If the temp file already exists, will this overwrite it or will it error out?
How about handling errors more gracefully? Cause you might be able to write the temp file, but not replace/overwrite the existing with the rename?

psihius (changes requested):

packages/create/create.js
Hornwitser: create runs by npm downloading all of its dependencies, unpacking them into a temporary directory and then invoking it from there. It doesn't have `@clusterio/lib` as a dependency because that's way too big to recursively download all dependencies of and run on the fly.

packages/lib/file_ops.js
Hornwitser: If the temp file already exists its content is discarded (truncated) before the new data is written to it.

I'm not sure what graceful handling of errors there is to do. If the rename fails then the operation as a whole failed an it must throw to prevent the calling code from continuing without having written the data it expected.

psihius (approved):
Merge it

PR #422, opened 2021-11-03
Ignore SessionLost from updateInstances

### LINKED ISSUE #418: Don't exit slave process if session is lost updating instances
It's possible though very unlikely that the connection to the master server is lost while the updateInstances call is being performed as a response to the connection be established in packages/slave/slave.js. In this case the current code shuts down due to it being an unexpected error. The session lost error should be expected and ignored instead.

Hornwitser:
Don't shutdown the slave if the connection session is lost during
updateInstances.  Resolves #418.

### DIFF
--- a/packages/slave/slave.js
+++ b/packages/slave/slave.js
@@ -1097,6 +1097,10 @@ class Slave extends libLink.Link {
 			}
 
 			this.updateInstances().catch((err) => {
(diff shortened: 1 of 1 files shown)


psihius (dismissed):

psihius:
Docker pull request test failed, needs looking into/

psihius (approved):

Issue #381, opened 2021-06-01
Display address and slave name in instance list


Hornwitser:
Resolve the slave ids to names in the instance list and add a column showing address to connect to it.

Previously attempted in #347.

PR #423, opened 2021-11-04
Slave instance UI updates

bbassie:
Implemented #381 partially because from what I see only the public_address was still missing.

Also added/changed the slave and instance status to use the Ant Design [Tag](https://ant.design/components/tag) component with some colors (I just picked some colors from the [colors list](https://ant.design/docs/spec/colors))

image

image

image

It also still only gets the game_port from the config, need to figure out what's the best way to get it if the config value is empty.

Feedback is welcome 😄

### DIFF
--- a/packages/lib/link/messages.js
+++ b/packages/lib/link/messages.js
@@ -1152,6 +1152,7 @@ messages.instanceStatusChanged = new Event({
 		"status": { enum: [
 			"stopped", "starting", "running", "stopping", "creating_save", "exporting_data",
 		]},
+		"game_port": { type: ["null", "integer"] },
 	},
 });
 
--- a/packages/lib/link/messages.js
+++ b/packages/lib/link/messages.js
@@ -337,6 +337,7 @@ let slaveProperties = {
 	"name": { type: "string" },
 	"id": { type: "integer" },
 	"connected": { type: "boolean" },
+	"public_address": { type: ["null", "string"] },
 };
 
 messages.listSlaves = new Request({
@@ -372,6 +373,7 @@ let instanceProperties = {
 	"name": { type: "string" },
 	"id": { type: "integer" },
 	"assigned_slave": { type: ["null", "integer"] },
+	"game_port": { type: ["null", "integer"] },
 	"status": { enum: [
 		"unknown", "unassigned", "stopped", "starting", "running", "stopping",
 		"creating_save", "exporting_data", "deleted",
--- a/packages/master/src/ControlConnection.js
+++ b/packages/master/src/ControlConnection.js
@@ -170,7 +170,7 @@ class ControlConnection extends BaseConnection {
 			id,
 			name: instance.config.get("instance.name"),
 			assigned_slave: instance.config.get("instance.assigned_slave"),
-			game_port: instance.game_port,
+			game_port: instance.game_port || null,
 			status: instance.status,
 		};
 	}
@@ -182,7 +182,7 @@ class ControlConnection extends BaseConnection {
 				id: instance.config.get("instance.id"),
 				name: instance.config.get("instance.name"),
 				assigned_slave: instance.config.get("instance.assigned_slave"),
-				game_port: instance.game_port,
+				game_port: instance.game_port || null,
 				status: instance.status,
 			});
 		}
@@ -202,7 +202,7 @@ class ControlConnection extends BaseConnection {
 				id: instance.config.get("instance.id"),
 				name: instance.config.get("instance.name"),
 				assigned_slave: instance.config.get("instance.assigned_slave"),
-				game_port: instance.game_port,
+				game_port: instance.game_port || null,
 				status: instance.status,
 			});
 		}
--- a/packages/master/src/ControlConnection.js
+++ b/packages/master/src/ControlConnection.js
@@ -106,6 +106,7 @@ class ControlConnection extends BaseConnection {
 				version: slave.version,
 				id: slave.id,
 				name: slave.name,
+				public_address: slave.public_address,
 				connected: this._master.wsServer.slaveConnections.has(slave.id),
 			});
 		}
@@ -169,6 +170,7 @@ class ControlConnection extends BaseConnection {
 			id,
 			name: instance.config.get("instance.name"),
 			assigned_slave: instance.config.get("instance.assigned_slave"),
+			game_port: instance.config.get("factorio.game_port"),
 			status: instance.status,
 		};
 	}
@@ -180,6 +182,7 @@ class ControlConnection extends BaseConnection {
 				id: instance.config.get("instance.id"),
 				name: instance.config.get("instance.name"),
 				assigned_slave: instance.config.get("instance.assigned_slave"),
+				game_port: instance.config.get("factorio.game_port"),
 				status: instance.status,
 			});
 		}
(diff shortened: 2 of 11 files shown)


packages/web_ui/src/components/InstanceStatusTag.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+const statusColors = {
+	unknown: "#d46b08",
+	unassigned: "#d4380d",
+	stopped: "#cf1322",
+	starting: "#7cb305",
+	running: "#389e0d",
+	stopping: "#d48806",
+	creating_save: "#096dd9",
+	exporting_data: "#08979c",
```
jake-87, line 12: IMO creating a save and exporting data are both very linked operations ( Working with data, writing to disk, etc ); perhaps they should be closer in color? #096dd9 ( dark blue-ish ) looks fine to me for creating_save, but I would suggest #69c0ff ( Lighter blue ) for exporting_data, keeping them both shades of blue as opposed to the contrasting #08979c ( teal-ish ) which feels like it's something completely different.

Danielv123:
That looks nice, I have had some issues distinguishing stopping/running on large amounts of instances. For getting the port, check how the server switcher plugin does it - I believe it is by listening to the factorio logs though.On Nov 4, 2021 05:06, J87 ***@***.***> wrote:
@jake-87 commented on this pull request.



In packages/web_ui/src/components/InstanceStatusTag.jsx:
> @@ -0,0 +1,21 @@
+import { Tag } from "antd";
+import React from "react";
+
+const statusColors = {
+	unknown: "#d46b08",
+	unassigned: "#d4380d",
+	stopped: "#cf1322",
+	starting: "#7cb305",
+	running: "#389e0d",
+	stopping: "#d48806",
+	creating_save: "#096dd9",
+	exporting_data: "#08979c",

IMO creating a save and exporting data are both very linked operations ( Working with data, writing to disk, etc ); perhaps they should be closer in color? #096dd9 ( dark blue-ish ) looks fine to me for creating_save, but I would suggest #69c0ff ( Lighter blue ) for exporting_data, keeping them both shades of blue as opposed to the contrasting #08979c ( teal-ish ) which feels like it's something completely different.

packages/web_ui/src/components/InstanceStatusTag.jsx
Danielv123: In this case they are entirely different operations though. Exporting_data scrapes lua prototype data and images and is more of a clusterwide task, while create_save is more of an instance setup/startup task. I think this is fine.

jake-87 (approved):
At least looking at the colors of the new section, LGTM.

Danielv123:
We also had some talk on the discord regarding gameports. Currently we use factorio to automatically assign ports when its not set in the config, but this has issues, mainly that it doesn't check for conflicts or assign to a specific allowed area. We would like to set a range of ports on each slave to assign, and then pick a port from that range when an instance is started without a fixed port in the config. We want ports to be sticky, so it will try to pick the same port each time it is started, but we do not want it to be persistent as it is when using the config field. This means the "randomly" selected instance port will have to be persisted in a new field somewhere. That also seems like it would also be a reasonable place to fetch the port for this feature.

Danielv123 (approved):
Looks all good to me. The tag was missing on the slave details page added in #420 so I added it there as well.

packages/master/src/ControlConnection.js
```diff
@@ -106,6 +106,7 @@ class ControlConnection extends BaseConnection {
 				version: slave.version,
 				id: slave.id,
 				name: slave.name,
+				public_address: slave.public_address,
```
Hornwitser, line 109: This field may be undefined for older versions.

packages/master/src/Master.js
```diff
@@ -368,6 +368,7 @@ class Master {
 			version: slave.version,
 			id: slave.id,
 			name: slave.name,
+			public_address: slave.public_address,
```
Hornwitser, line 371: This field may be undefined for older versions causing validation to fail.

packages/slave/slave.js
```diff
@@ -692,6 +694,8 @@ class Instance extends libLink.Link {
 	}
 
 	async masterConnectionEventEventHandler(message) {
+		this.notifyStatus(this._status);
+
```
Hornwitser, line 698: Why are you doing this? It'll send updates when the connection is down as well as unnecessary updates when the connection is dropped and resumed.

packages/slave/slave.js
```diff
@@ -1296,6 +1301,7 @@ class Slave extends libLink.Link {
 		libLink.messages.instanceStatusChanged.send(this, {
 			instance_id,
 			status: instanceConnection ? instanceConnection.status : "stopped",
+			game_port: null,
```
Hornwitser, line 1304: This will cause the game_port to always be notified to the master as null after a reconnect, which is definitively not right.

packages/web_ui/src/components/InstanceList.jsx
```diff
@@ -1,12 +1,13 @@
 import React from "react";
 import { useHistory } from "react-router-dom";
-import { Table } from "antd";
+import { notification, Button, Table } from "antd";
+import { CopyOutlined } from "@ant-design/icons";
```
Hornwitser, line 4: Do no import from ant-design/icons, it's got 3000 sub-modules and breaks development tools. Use
```js
import CopyOutlined from "@ant-design/icons/CopyOutlined";
```

packages/web_ui/src/components/InstanceList.jsx
```diff
@@ -18,7 +19,13 @@ export default function InstanceList(props) {
 		}
 		return String(slaveId);
 	}
-
+	function slavePublicAddress(slaveId) {
+		let slave = slaveList.find(s => s.id === slaveId);
+		if (slave) {
+			return slave.public_address;
```
Hornwitser, line 25: This may be undefined on old versions.

packages/web_ui/src/components/InstanceList.jsx
```diff
@@ -32,9 +39,34 @@ export default function InstanceList(props) {
 			render: instance => slaveName(instance["assigned_slave"]),
 			sorter: (a, b) => strcmp(slaveName(a["assigned_slave"]), slaveName(b["assigned_slave"])),
 		},
+		{
+			title: "Public address",
+			key: "public_address",
+			render: instance => {
+				let address = slavePublicAddress(instance["assigned_slave"]);
+				let full_address = address + (instance["game_port"] !== null ? `:${instance["game_port"]}` : "");
```
Hornwitser, line 47: Address should not render as `undefined` or `null` if it's not known.

packages/lib/link/messages.js
```diff
@@ -1150,6 +1152,7 @@ messages.instanceStatusChanged = new Event({
 		"status": { enum: [
 			"stopped", "starting", "running", "stopping", "creating_save", "exporting_data",
 		]},
+		"game_port": { type: ["null", "integer"] },
```
Hornwitser, line 1155: Properties added to `eventProperties` are by default required, which will cause slaves connecting from the previous version to fail. You need to add an `eventRequired` array with a list of the previous properties and handle game_port missing in the event handler.

Hornwitser (changes requested):
You need to correctly handle a slave from the previous version connecting which isn't going send game_port or public_address. Otherwise updating becomes too painful.

PR #425, opened 2021-11-14
Add timeout to invokeHook on plugins

Danielv123:
If a plugin fails to resolve its promise in a hook handler the slave would get stuck, preventing other plugins from processing their hooks. The only way to recover (in the case of a slave failing to stop properly) was a slave restart.

This PR adds a 15 second max processing time. After that time, the error will be logged and the next hook will be processed.

image

### DIFF
--- a/packages/lib/plugin.js
+++ b/packages/lib/plugin.js
@@ -4,6 +4,7 @@
  */
 "use strict";
 
+const helpers = require("./helpers");
 
 /**
  * Conceptual base for master and instance plugins.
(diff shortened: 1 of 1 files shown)


Hornwitser (changes requested):
This leaks the timer, which means on shutdown you're going to end up waiting 15 seconds after last instance stopping for the timer to expire. You should probably be using lib/helpers.timeout here.

psihius (approved):

Hornwitser (changes requested):
Now it's broken, returning the error instead of throwing it. Also creating an error instance when you don't use it isn't free since it has to create a stack traceback.

Danielv123:
🤦‍♂️ I'll fix that and test it properly when I get home.

Danielv123:
Not the prettiest solution, but it seems right. Try/catch needs to stay because hook can still throw. I used a string object to see if the returned value is an error or a response object of some kind. Afaik this cannot be bypassed when using a strict comparison, although I doubt anyone will intentionally try to make it throw.

psihius (approved):
Well, it seems you went through multiple iterations and applied some dark magic here, and since it's so dark like it's from DC universe, I don't have any reason to doubt it's what is needed here.

:rofl:

Hornwitser:
Seems a bit odd of a way to do it for me. You can use an empty object to check if the action timed out like this
```js
const timedOut = {}
let result = await helpers.timeout(plugin[hook](...args), 15000, timedOut);
if (result === timedOut) {
    throw new Error(`Invoking hook ${hook} timed out for plugin ${name}`);
}
```

Danielv123:
That is essentially what I am doing, but I wanted the code to be more descriptive. `let x = {}` is 40% faster than `let x = new String("Error message")` but I doubt the 20ns makes much of a difference here. Should I change it?

Hornwitser (approved):

PR #426, opened 2021-11-14
Start all, stop all and delete all instance buttons

Danielv123:
When playing around with my gridworld plugin I frequently found myself needing to spin my whole cluster up and down a lot, thought this might be useful for someone else as well.

image

image

### DIFF
--- a/packages/web_ui/src/components/InstancesPage.jsx
+++ b/packages/web_ui/src/components/InstancesPage.jsx
@@ -11,7 +11,6 @@ import { useInstanceList } from "../model/instance";
 import InstanceList from "./InstanceList";
 import { notifyErrorHandler } from "../util/notify";
 
-
 function CreateInstanceButton(props) {
 	let control = useContext(ControlContext);
 	let history = useHistory();
@@ -51,7 +50,7 @@ function CreateInstanceButton(props) {
 		>
 			<Form form={form}>
 				<Form.Item name="instanceName" label="Name">
-					<Input/>
+					<Input />
 				</Form.Item>
 			</Form>
 		</Modal>
@@ -59,6 +58,7 @@ function CreateInstanceButton(props) {
 }
 
 export default function InstancesPage() {
+	let control = useContext(ControlContext);
 	let account = useAccount();
 	let [instanceList] = useInstanceList();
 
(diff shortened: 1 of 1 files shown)


Hornwitser:
I am opposed to the idea of adding a delete absolutely everything with no recovery button. It might be useful for development, but in production that button is a liability.

Danielv123:
Thinking about it, I agree. There is a *lot* of work to be lost clicking it on a long running cluster, and no real reason to. It does save a few hundred clicks a day during development though, so I want it to stick around. How about only showing it when the devserver is detected? I believe react has some environment variable for that.

Starting up all instances at the same time works fine for me with ~100 instances on one host, they just take a bit longer to start. I found it to be preferable to awaiting the start, which took ages and also introduces state management challenges if you want to cancel the starting process halfway through. Maybe a pool of starting 10 at a time or something? Have to remember that in a large cluster there are only going to be a few instances per slave anyways.

I am thinking that should maybe be implemented with a general queue on the master or slave with a setting as well.  There is no reason why the web interface should avoid starting instances quickly to avoid causing issues but tasks scripted through other means shouldn't.

What part are you worried about? Master load, race conditions, slave load or filesystem load from opening so many instances of factorio?

Oh, and if we want a queue we have to figure out a good way to deal with some instances not starting. Having to manually start 50 instances because one gets stuck on "starting" feels wrong.

psihius:
"Delete all" - sometimes it is useful, but in this case maybe a "select all" with checkboxes so you can still mass delete - one way or the other it might be needed for mass operations. For example when you want to clear a cluster and setup a new one (maybe even have a script that generates all the servers).
Deleting 1 by 1 is always something I run into eventually. Just need to make sure there are safeguards from accidental clicking :D

Hornwitser:
I am also against breaking the permission system like that, admin means admin. The fact that you think the delete all button is so dangerous it shouldn't be granted by default to cluster admins is very telling. This "permission" is also enforced client side which goes against the whole point of the permission system.

If you need to delete everything often enough that you can't be bothered to click them one by one we already have a solution to this problem. It's called clusterioctl. Some tweaks to the output and it should be easy to pipe `clusterioctl instance list` to `clusterioctl instance delete`.

Danielv123:
Since its basically only needed with easy access to mass instance creation I removed it from here for now. I will instead implement it in my plugin where its actually required.

psihius (approved):

psihius:
Merge it

PR #427, opened 2021-11-15
Gh actions

Danielv123:
Got annoyed at builds failing for no reason. Thoughts?


Hornwitser:
It should run lint check as part of the testing.

psihius (approved):
Seems good to me.

Hornwitser (approved):
I'm not sure why it needs to test on node version 14 and 16, that seems like a waste of resources to me. Other than that it looks good to me.

PR #428, opened 2021-11-15
Fix error when localhost resolves to ::1

Hornwitser:
Some systems may resolve localhost to ::1 while Factorio doesn't respond
to RCON connection over IPv6 loopback, this causes instances startup to
fail.  Fix by using 127.0.0.1 instead of localhost.

### DIFF
--- a/packages/lib/factorio/server.js
+++ b/packages/lib/factorio/server.js
@@ -690,7 +690,7 @@ class FactorioServer extends events.EventEmitter {
 		}
 
 		let config = {
-			host: "localhost",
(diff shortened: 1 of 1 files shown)


Danielv123 (approved):
Maybe the docs should specify under the requirements somewhere that factorio requires ipv4 support?

psihius (approved):

Hornwitser:
Wha? Who runs Factorio without a working IPv4 stack?

psihius:
I think @Danielv123 probably meant to note that IPv6 does not work with RCON, so needs an IPv4 address

Danielv123:
Crazy people I assume.

PR #435, opened 2021-11-23
Add gridworld and edge_transports to create

Danielv123:
Edge transports has been around for some time, is there a reason why it wasn't included? Gridworld will likely still have a lot of development, but it is in a playable state now, especially since its basically just a configuration tool.


Hornwitser:
Too difficult to set up and use, and being incomplete.

psihius:
In my view we should include it at some point, maybe yeah spend a bit more time polishing and testing (and ask our users to do testing too). I vote for keeping this as a WIP PR with the intention to merge it when we feel it's good enough to call it a beta release?

Hornwitser:
This can be reopened when both these plugins are ready for public use. Or ideally have a new PR created to only add whichever is ready first.

PR #436, opened 2021-12-07
Wrong capitalisation on `process.env`

jake-87:


jake-87:
Apologies that these slipped through, as far as I can tell, they were misspelled while doing final bugfixes in the Github online editor, hence why they were not caught. I am working on unit tests for these systems, but it is more complicated due to the interactive nature of the system.

packages/ctl/ctl.js
```diff
@@ -70,7 +70,7 @@ async function configToKeyVal(data) {
 
 async function serializedConfigToString(serlializedConfig, configGroup, disallowedList) {
 	let allConfigElements = "";
-	for (let group of serializedConfig.groups) {
+	for (let group of serlializedConfig.groups) {
```
Danielv123, line 73: Why did you add a new typo?

Danielv123 (changes requested):

Danielv123 (approved):
Should be squash-merged

psihius (approved):
And this is why I run an IDE with a spell checker :100:

PR #439, opened 2021-12-08
Use strict parsing in installer

Hornwitser:
Bail out if arguments that aren't understood are passed to the
installer script instead of silently ignoring them.

### DIFF
--- a/packages/create/create.js
+++ b/packages/create/create.js
@@ -527,7 +527,9 @@ async function main() {
 		})
 		.option("plugins", {
 			array: true, describe: "Plugins to install", type: "string",
-		});
+		})
+		.strict()
+	;
 
 	if (process.platform === "linux") {
 		args = args


packages/create/create.js
```diff
@@ -527,7 +527,9 @@ async function main() {
 		})
 		.option("plugins", {
 			array: true, describe: "Plugins to install", type: "string",
-		});
+		})
+		.strict()
+	;
```
Danielv123, line 532: The semicolon on a seperate line looks really weird to me - is that something we want in our guidelines/linter config?

Danielv123 (approved):

packages/create/create.js
Danielv123: Personally I think of it as a single line that is split for organization. There is no closure, so making an ending indent doesn't make sense. Not to mention
```js
		})
		.strict();
	;
```

also being valid JS.

Hornwitser: Consider for example
```js
foo(
    arg1,
    arg2,
);
```
I think putting the semi colon at the end of the last indented line is less consistent with the above style. And I firmly believe that having the semi colon on its own line here makes it easier to read than the alternative because it's consistent with all the other multiline continuation of statements.

Danielv123: Only other argument I can find for it is [here](https://github.com/airbnb/javascript/issues/1717). It seems he was interested in doing it for cleaner diffs, but everyone who are interested in clean diffs just drop the semicolon entirely. I am not particularly against it, I just don't like it much because I haven't ever seen it anywhere else. I am good with this pull request as is though.

Issue #438, opened 2021-12-08
Specify options on installscript programatically


ElGauchooooo:
Is there currently a way to specify all the options asked by the installscript programatically by command line?
I've had a small desktop application that lets you manage multiple clusterio instances, which boots up new Amazon AWS EC2 instances as needed and initializes them - without the user having to setting up all clients / master by themself. 

What i'd like to is run something like `npm init @clusterio -admin ElGauchooooo -mode Master -plugins ...` (i guess you get the idea).

Danielv123:
There is no such functionality for the init script afaik, but everything done by the init script can be done through the clusteriomaster/clusterioslave/clusterioctl CLI tools. Only things missing from the CLI is stuff like save uploading (which you can still do by navigating into the correct folder).

Very interested in that application you got, might be something we can learn from it. Cluster management is a big focus for 2.0.

Danielv123:
Here is the install guide from before the init script was added, it will works the same way: https://github.com/clusterio/clusterio/tree/1c4010a744ae0a2c9c56fb61b0a34be83ffca4d5

jake-87:
I don't know about how npm handles init scripts + command line toggles, but `create.js` does seem to have yarg options for the needed parameters, so perhaps that could be adapted at some point?

https://github.com/clusterio/clusterio/blob/master/packages/create/create.js#L498

ElGauchooooo:
> There is no such functionality for the init script afaik, but everything done by the init script can be done through the clusteriomaster/clusterioslave/clusterioctl CLI tools. Only things missing from the CLI is stuff like save uploading (which you can still do by navigating into the correct folder).

I'll give it a try once i get back home, thank you. 

> Very interested in that application you got, might be something we can learn from it. Cluster management is a big focus for 2.0.

Sure thing, for the old clusterio version, it was just a bit hacky, but got the things done to start, persist and re-start all servers, including booting and shutting down all aws instances + backing up in S3 to save money. I'll upload a public version to one of my repos once it reaches a bit more polished state and fits everyone needs. I'll keep you updated.

Danielv123:
create.js is the init script, so if args are supported there then they are supported.On Dec 8, 2021 12:48, ElGauchooooo ***@***.***> wrote:

There is no such functionality for the init script afaik, but everything done by the init script can be done through the clusteriomaster/clusterioslave/clusterioctl CLI tools. Only things missing from the CLI is stuff like save uploading (which you can still do by navigating into the correct folder).

[...]

PR #440, opened 2021-12-08
Fix installer ignoring --plugins argument

Hornwitser:
Fix issue passing --plugins to installer which was discovered in #438.


psihius (approved):
Merge it, this one is obvious and easy fix :)

PR #441, opened 2021-12-10
Added bulk import of whitelist, admin and banlist

jake-87:


jake-87:
At the moment I have the ban reason as "bulk import"; should it also be imported from the file? It would complicate the creation of a ban list, but could be useful to a banned user.

Hornwitser:
Usernames don't have spaces in them so you could separate the name and reason with that.

jake-87:
@Hornwitser Should be ready for review.

Hornwitser:
This is missing a way to do it through the web interface.

Danielv123 (approved):
This looks good to me. We should absolutely have the same functionality available in the web interface, whether in this PR or its own.

packages/ctl/ctl.js
```diff
@@ -903,6 +903,29 @@ userCommands.add(new libCommand.Command({
...
+		yargs.positional("file", { describe: "File to import from", type: "string" });
+		yargs.options({
+			"create": { describe: "Create user if it does not exist", nargs: 0, type: "boolean", default: false },
+		});
+	}],
+	handler: async function(args, control) {
+		let data = await fs.readFile(args.file, "utf8");
+		let names = data.split("\n");
+		for (let name in names) {
+			if (name in names) {
```
Danielv123, line 917: What is the point of this though?

jake-87:
I’d be happy add the webui in this PR, however it may be a few days.

packages/ctl/ctl.js
jake-87: Not sure myself. eslint has spit out errors in the past if it’s not there, so I preemptively added it. Unfortunately I can’t test if it’s needed right now, but if it’s not, i’d be happy to remove it.

Hornwitser: `for (... in ...)` loops are an error more often than not, you shouldn't use them.

jake-87: OK. I’ll replace it next commit.

psihius:
@jake-87 converted it to draft, press ze button when it's ready for review :)

jake-87:
All this needs is a way to refresh the user table so the bulk imports show up, and it's done, as far as I can tell.

psihius:
>

psihius:
> All this needs is a way to refresh the user table so the bulk imports show up, and it's done, as far as I can tell.

You probably need an event/message for that to send out, so the systems grab it from the websocket message bus.
Also, make sure tests don't fail :)

Bonus would be adding tests for this functionality.

jake-87:
Apologies, forgot to run final tests after removing personal stuff. I'll see what I can do in terms of the event.

packages/ctl/ctl.js
```diff
@@ -903,6 +903,26 @@ userCommands.add(new libCommand.Command({
...
+	handler: async function(args, control) {
+		let data = await fs.readFile(args.file, "utf8");
+		let names = data.split("\n");
+		await names.forEach(async (name) => {
+			if (name !== "") {
+				await libLink.messages.setUserAdmin.send(control, {
+					name: name, create: args.create, admin: true,
+				});
+			}
+		});
```
Hornwitser, line 922: This does not do what you think it does. If you're using a function you're unsure of how it works [consult a reference for it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach). `forEach` returns `undefined` awaiting it is a logical error here. The callback is not awaited which means all of the requests will be started concurrently and that if any one of them throws Node.js will abort with an unhandled promise rejection. A regular `for (... of ...)` loop would be more appropriate here.

packages/ctl/ctl.js
```diff
@@ -913,16 +913,13 @@ userCommands.add(new libCommand.Command({
 	handler: async function(args, control) {
 		let data = await fs.readFile(args.file, "utf8");
 		let names = data.split("\n");
-		for (let name in names) {
-			if (name in names) {
-				if (names[name] === "") {
-					continue;
-				}
+		await names.forEach(async (name) => {
```
Danielv123, line 916: This is wrong. `Array.forEach` returns undefined, which isn't a Promise. You want to do `await Promise.all(names.map(async name => ...` instead

Danielv123: This goes for all the places you are using forEach here I see. Hornwitser suggested a for loop. The difference is that promise.all executes in parallel while `for` does sequential. I guess for a long whitelist you'd need to go sequential or batched parallel.

jake-87: Doing it in parallel was my original intent; unfortunately I did mess up the use of forEach. I’ll correct this next commit.

jake-87:
I am still working on this. Unfortunately the documentation for the link system is less than optimal, and I am having a hard time understanding the code, so I can't give a decent estimate on when the web interface will be finished.

psihius:
@jake-87 ask in Discord? We can also improve the docs this way :)

jake-87:
I'm not going to be able to finish this. Apologies. Feel free to take & merge the CLI code, but the websocket interface is not well documented enough for me to be able to understand it, and while I appreciate the discord's existence, in my mind it is not a good enough replacement for docs.

PR #445, opened 2021-12-19
Fix logs

Hornwitser:
Implements automatic splitting logs by date and better querying of the logs. When querying by descending date order the log files are read line by line in reverse and stops after the limit is reached instead of reading all of the logs from the start. Combined with a simple file based query index the log queries now finish much faster, and do not use 100s of MB of memory like the old placeholder implementation did.

This is the long awaited fix for log queries being completely broken. The logs being split by date also allows for an easy way to manually delete old logs.


psihius (approved):
Looks good, merge it

PR #447, opened 2022-01-03
Fix issues detected by lgtm.com

Hornwitser:
Fixes for issues detected by static code analysis. See https://lgtm.com/projects/g/clusterio/clusterio/ for the alerts notified about.

### DIFF
--- a/packages/lib/factorio/server.js
+++ b/packages/lib/factorio/server.js
@@ -791,6 +791,7 @@ class FactorioServer extends events.EventEmitter {
 
 		try {
 			let [code, signal] = await events.once(this._server, "exit");
+			void signal;
 			if (code !== 0) {
 				throw new Error(`Factorio exited with status ${code}`);
 			}
--- a/packages/master/master.js
+++ b/packages/master/master.js
@@ -43,18 +43,18 @@ const UserManager = require("./src/UserManager");
 let master;
 
 
-const slaveMappingGauge = new libPrometheus.Gauge(
+void new libPrometheus.Gauge(
 	"clusterio_master_slave_mapping",
 	"Mapping of Slave ID to name",
 	{
 		labels: ["slave_id", "slave_name"],
-		callback: function() {
-			slaveMappingGauge.clear();
+		callback: function(gauge) {
+			gauge.clear();
 			if (!master || !master.slaves) {
 				return;
 			}
 			for (let [id, slave] of master.slaves) {
-				slaveMappingGauge.labels({
+				gauge.labels({
 					slave_id: String(id),
 					slave_name: slave.name,
 				}).set(1);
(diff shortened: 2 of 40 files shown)


packages/lib/factorio/server.js
```diff
@@ -791,6 +791,7 @@ class FactorioServer extends events.EventEmitter {
 
 		try {
 			let [code, signal] = await events.once(this._server, "exit");
+			void signal;
```
Danielv123, line 794: I don't understand what `void` is supposed to do here? I can't find a reference to it in the lgtm link either.

packages/master/master.js
```diff
@@ -89,12 +89,14 @@ const wsActiveConnectionsGauge = new libPrometheus.Gauge(
 	"How many WebSocket connections are currently open to the master server",
 	{ callback: function(gauge) { gauge.set(master.wsServer.activeConnectors.size); }}
 );
+void wsActiveConnectionsGauge;
```
Danielv123, line 92: Also don't understand the point of this. `void [object]` simply evaluates the object to an object and returns `undefined`. This looks like a noop?

packages/web_ui/src/model/instance.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useContext, useState } from "react";
```
Danielv123, line 1: Don't you still need to import React in jsx files to get babel to behave properly? Which is why the `Import React from "react";` still exists in the other files. Or does destructuring work for the same purpose?

packages/lib/factorio/server.js
Hornwitser: It's a common way to mark unused parameters or variables as intentionally not used.

packages/master/master.js
Hornwitser: Same with this. `void wsActiveConnectionsGauge;` doesn't do anything but looks like a usage of the variable for static analysis tools.

packages/web_ui/src/model/instance.jsx
Hornwitser: As far as I understand React is only referenced if there's JSX syntax like `<div>` in the file. The model files don't have any such objects in them.

packages/master/master.js
Danielv123: If the variable really is unused, why not just do `void new libPrometheus.Gauge(...)`?

Hornwitser: That is an option but seems a bit weird to create anonymous metrics to me.

Danielv123: I mean, it is though. Putting it in a variable indicates that the variable is passed to be used somewhere. In this case, the variable isn't used and can't really be used since  the metric is updated through the callback.

Hornwitser: It can be used to retrieve the value, e.g. wsActiveConnectionsGauge.get() for a status report. But the variable can be added when that status report is added.

Danielv123 (approved):
This looks good to me.

PR #452, opened 2022-02-01
Added setup guide for slaves from web interface

Danielv123:
Adds a slave setup guide to the generate slave token dialog.

Screenshot from 2022-02-02 00-07-41

This also exposes the resolvedPath of plugins manually installed from folders to the web. I do this to automatically select all plugins installed from npm on the master server. If anyone has anything against that we can either remove the feature or filter out plugins installed from folders.

If no plugins are found it will default to the normal interactive menu from `packages/create`

The copy buttons now use the browser clipboard API which is supported basically everywhere https://caniuse.com/async-clipboard

### DIFF
--- a/packages/master/src/routes.js
+++ b/packages/master/src/routes.js
@@ -117,7 +117,7 @@ function getPlugins(req, res) {
 		if (web.main === "remoteEntry.js") {
 			web.error = "Incompatible old remoteEntry.js entrypoint.";
 		}
-		plugins.push({ name, version: pluginInfo.version, enabled, loaded, web });
+		plugins.push({ name, version: pluginInfo.version, enabled, loaded, web, requirePath: pluginInfo.requirePath });
 	}
 	res.send(plugins);
 }
--- a/packages/web_ui/src/components/SlavesPage.jsx
+++ b/packages/web_ui/src/components/SlavesPage.jsx
@@ -1,6 +1,7 @@
-import React, { useContext, useRef, useState } from "react";
+import React, { useContext, useRef, useState, useEffect } from "react";
 import { useHistory } from "react-router-dom";
-import { Button, Form, Input, Modal, PageHeader, Table, Tag } from "antd";
+import { Button, Form, Input, Modal, PageHeader, Table, Tag, Typography } from "antd";
+import CopyOutlined from "@ant-design/icons/lib/icons/CopyOutlined";
 
 import { libLink } from "@clusterio/lib";
 
@@ -9,7 +10,7 @@ import ControlContext from "./ControlContext";
 import PageLayout from "./PageLayout";
 import PluginExtra from "./PluginExtra";
 import { useSlaveList } from "../model/slave";
-import { notifyErrorHandler } from "../util/notify";
+import notify, { notifyErrorHandler } from "../util/notify";
 
 const strcmp = new Intl.Collator(undefined, { numerice: "true", sensitivity: "base" }).compare;
 
@@ -18,12 +19,24 @@ function GenerateSlaveTokenButton(props) {
 	let control = useContext(ControlContext);
 	let [visible, setVisible] = useState(false);
 	let [token, setToken] = useState(null);
+	let [slaveId, setSlaveId] = useState(null);
 	let [form] = Form.useForm();
 	let tokenTextAreaRef = useRef(null);
+	let [pluginList, setPluginList] = useState([]);
+	useEffect(() => {
+		(async () => {
+			let response = await fetch(`${webRoot}api/plugins`);
+			if (response.ok) {
+				const plugins = await response.json();
+				setPluginList(plugins);
+			} else {
+				notify("Failed to load plugin list");
+			}
+		})();
+	}, []);
 
 	async function generateToken() {
 		let values = form.getFieldsValue();
-		let slaveId = null;
 		if (values.slaveId) {
 			slaveId = Number.parseInt(values.slaveId, 10);
 			if (Number.isNaN(slaveId)) {
@@ -35,8 +48,11 @@ function GenerateSlaveTokenButton(props) {
 
 		let result = await libLink.messages.generateSlaveToken.send(control, { slave_id: slaveId });
 		setToken(result.token);
+		setSlaveId(slaveId);
 	}
 
+	// Only install plugins that aren't filesystem paths. Npm modules have max 1 forward slash in their name.
+	const pluginString = pluginList.map(p => `"${p.requirePath}"`).filter(x => x.split("/") <= 1).join(" ");
 	return <>
 		<Button
 			onClick={() => { setVisible(true); }}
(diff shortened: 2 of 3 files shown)


psihius (approved):

psihius:
@Danielv123 @Hornwitser is this one still relevant? If it is, I see no reason not to sync it and merge it.

Danielv123:
Afaik this is still relevant as slave setup hasn't changed. Haven't tested it with an upstream master branch in a few months though.

psihius:
> Afaik this is still relevant as slave setup hasn't changed. Haven't tested it with an upstream master branch in a few months though.

Would you be able to rebase it then and let's knock this one out then? :)

Danielv123:
Sure, soon™️

packages/master/src/routes.js
```diff
@@ -117,7 +117,7 @@ function getPlugins(req, res) {
 		if (web.main === "remoteEntry.js") {
 			web.error = "Incompatible old remoteEntry.js entrypoint.";
 		}
-		plugins.push({ name, version: pluginInfo.version, enabled, loaded, web });
+		plugins.push({ name, version: pluginInfo.version, enabled, loaded, web, requirePath: pluginInfo.requirePath });
```
Hornwitser, line 120: I don't like that you leak the full require path in a unauthenticated endpoint when what you actually want is is to include package.name if it equals requirePath and package.private is not true. This should also use the `snake_case` convention like everything else that goes over the network.

packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -50,30 +66,105 @@ function GenerateSlaveTokenButton(props) {
...
-				</Form.Item>
-				<Form.Item>
-					<Button
-						onClick={() => { generateToken().catch(notifyErrorHandler("Error generating token")); }}
-					>Generate</Button>
-				</Form.Item>
-				<Form.Item label="Token" required>
-					<Input.TextArea value={token} autoSize ref={tokenTextAreaRef} />
+					<Input onChange={() => { generateToken().catch(notifyErrorHandler("Error generating token")); }} />
 				</Form.Item>
```
Hornwitser, line 74: Slave ID is marked optional yet there's no way to generate an input without inputting an ID here.

image

packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -18,12 +19,24 @@ function GenerateSlaveTokenButton(props) {
...
+				notify("Failed to load plugin list");
+			}
+		})();
+	}, []);
 
 	async function generateToken() {
 		let values = form.getFieldsValue();
-		let slaveId = null;
 		if (values.slaveId) {
 			slaveId = Number.parseInt(values.slaveId, 10);
```
Hornwitser, line 41: `slaveId` is changed to a react state variable but still being set directly here, this leads to ill formed behaviour: inputting "a" into the text field then removing it with backspace causes a generateSlaveToken request with slade_id: NaN being sent.

packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -50,30 +66,105 @@ function GenerateSlaveTokenButton(props) {
...
+		onClick={async () => {
+			if (await checkClipboardPermission()) {
+				navigator.clipboard.writeText(text);
+				notify(message);
+			}
+		}}
+	>
+		<CopyOutlined />
+	</Button>;
+}
```
Hornwitser, line 166: This new copy button just shows up as red in my browser and doesn't do anything when I click on it?

image

Also the spacing of the text is wrong, there should be more space from the code block to the unrelated text below it than there is from the code block to the related text above it.

packages/web_ui/src/components/SlavesPage.jsx
```diff
@@ -18,12 +19,24 @@ function GenerateSlaveTokenButton(props) {
 	let control = useContext(ControlContext);
 	let [visible, setVisible] = useState(false);
 	let [token, setToken] = useState(null);
+	let [slaveId, setSlaveId] = useState(null);
 	let [form] = Form.useForm();
 	let tokenTextAreaRef = useRef(null);
```
Hornwitser, line 24: This variable is no longer used.

Hornwitser (commented):
I didn't review this before now, but now that it's merged I had a look at it and found some issues with the way it was implemented.

Danielv123:
I'll get that fixed. The copy button is red because you haven't given it the permission to access the clipboard. It should ask in the top left corner by the URL bar. I am thinking I should add a hovertext explaining it maybe, and make it request it again on clicking if that is possible.On 16 Jul 2022 17:52, Hornwitser ***@***.***> wrote:
@Hornwitser commented on this pull request.

I didn't review this before now, but now that it's merged I had a look at it and found some issues with the way it was implemented.

In packages/master/src/routes.js:
> -		plugins.push({ name, version: pluginInfo.version, enabled, loaded, web });
+		plugins.push({ name, version: pluginInfo.version, enabled, loaded, web, requirePath: pluginInfo.requirePath });

I don't like that you leak the full require path in a unauthenticated endpoint when what you actually want is is to include package.name if it equals requirePath and package.private is not true. This should also use the snake_case convention like everything else that goes over the network.



In packages/web_ui/src/components/SlavesPage.jsx:
>  				<Form.Item name="slaveId" label="Slave ID">
-					<Input/>
-				</Form.Item>
-				<Form.Item>
-					<Button
-						onClick={() => { generateToken().catch(notifyErrorHandler("Error generating token")); }}
-					>Generate</Button>
-				</Form.Item>
-				<Form.Item label="Token" required>
-					<Input.TextArea value={token} autoSize ref={tokenTextAreaRef} />
+					<Input onChange={() => { generateToken().catch(notifyErrorHandler("Error generating token")); }} />
 				</Form.Item>

Slave ID is marked optional yet there's no way to generate an input without inputting an ID here.




In packages/web_ui/src/components/SlavesPage.jsx:
> -		let slaveId = null;
 		if (values.slaveId) {
 			slaveId = Number.parseInt(values.slaveId, 10);

slaveId is changed to a react state variable but still being set directly here, this leads to ill formed behaviour: inputting "a" into the text field then removing it with backspace causes a generateSlaveToken request with slade_id: NaN being sent.



In packages/web_ui/src/components/SlavesPage.jsx:
> +function CopyButton({ text, message }) {
+	let [clipboardPermision, setClipboardPermission] = useState("granted");
+	useEffect(() => {
+		(async () => {
+			let result = await navigator.permissions.query({ name: "clipboard-write" });
+			// result.state is "granted", "denied" or "prompt"
+			setClipboardPermission(result.state);
+			result.onchange = function () {
+				setClipboardPermission(result.state);
+			};
+		})();
+	}, []);
+
+	async function checkClipboardPermission() {
+		let result = await navigator.permissions.query({ name: "clipboard-write" });
+		// result.state is "granted", "denied" or "prompt"
+		setClipboardPermission(result.state);
+		result.onchange = function () {
+			setClipboardPermission(result.state);
+		};
+		return result.state === "granted";
+	}
+
+	return <Button
+		className="copy-button"
+		danger={clipboardPermision !== "granted"}
+		onClick={async () => {
+			if (await checkClipboardPermission()) {
+				navigator.clipboard.writeText(text);
+				notify(message);
+			}
+		}}
+	>
+		<CopyOutlined />
+	</Button>;
+}

This new copy button just shows up as red in my browser and doesn't do anything when I click on it?

Also the spacing of the text is wrong, there should be more space from the code block to the unrelated text below it than there is from the code block to the related text above it.



In packages/web_ui/src/components/SlavesPage.jsx:
> @@ -18,12 +19,24 @@ function GenerateSlaveTokenButton(props) {
 	let control = useContext(ControlContext);
 	let [visible, setVisible] = useState(false);
 	let [token, setToken] = useState(null);
+	let [slaveId, setSlaveId] = useState(null);
 	let [form] = Form.useForm();
 	let tokenTextAreaRef = useRef(null);

This variable is no longer used.

Hornwitser:
> I'll get that fixed. The copy button is red because you haven't given it the permission to access the clipboard. It should ask in the top left corner by the URL bar. I am thinking I should add a hovertext explaining it maybe, and make it request it again on clicking if that is possible.

There is no prompt. Not when loading it up, and not when clicking on it.

Also, please use common internet ettiquette and refrain from top posting.

Hornwitser:
> There is no prompt. Not when loading it up, and not when clicking on it.

I looked into this and the reason the clipboard is unavailable in my environment is that the page wasn't considered a secure context when served over HTTP. The clipboard button just being red and not working is not intuitive here, it should say why it doesn't work.

PR #453, opened 2022-03-26
Small things 7

Hornwitser:
Additions needed for the FAGC plugin to integrate nicely, and some minor QoL improvements. See commit log for changes.

### DIFF
--- a/packages/lib/plugin.js
+++ b/packages/lib/plugin.js
@@ -653,6 +653,42 @@ class BaseWebPlugin {
 		 * @type {Array<module:lib/plugin~Page>}
 		 */
 		this.pages = [];
+
+		/**
+		 * Extra react component to add to core components
+		 *
+		 * Interface to augment core components of the web UI.  Setting a
+		 * component as one of the supported properties of this object will
+		 * cause the web UI to render it when displaying that component,
+		 * usually at the end.  Each component will receive a `plugin` param
+		 * which is the instance of the web plugin that contained the
+		 * component extra.
+		 *
+		 * @type {object}
+		 * @property {React.ComponentType} MasterPage -
+		 *     Placed at the end of the master page.
+		 * @property {React.ComponentType} SlavesPage -
+		 *     Placed at the end of the slaves list page.
+		 * @property {React.ComponentType} SlaveViewPage -
+		 *     Placed at the end of each slave page.  Takes a `slave` param
+		 *     which is the slave the page is displayed for.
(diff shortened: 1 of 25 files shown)


packages/lib/plugin.js
```diff
@@ -355,6 +355,19 @@ class BaseMasterPlugin {
...
 	async onInstanceStatusChanged(instance, prev) { }
 
+	/**
+	 * Called when the value of a master config field changed.
+	 *
+	 * Invoked after the value of the config field given by `field` has
+	 * changed on the master.
+	 *
+	 * @param {module:lib/config.ConfigGroup} group -
+	 *     The group who's field got changed on.
```
Danielv123, line 365: The group who's field got changed ~~on.~~?

packages/lib/plugin.js
```diff
@@ -640,6 +656,42 @@ class BaseWebPlugin {
...
+		 * @property {React.ComponentType} UserViewPage -
+		 *     Placed at the end of each user page.  Takes a `user` param
+		 *     which is the user object the page is displayed for.
+		 * @property {React.ComponentType} RolesPage -
+		 *     Placed at the end of the roles list page.
+		 * @property {React.ComponentType} RoleViewPage -
+		 *     Placed at the end of each role page.  Takes a `role` param
+		 *     which is the role object the page is displayed for.
+		 */
+		this.componentExtra = {};
```
Danielv123, line 694: This is neat. Should also be added to the docs.

Danielv123 (approved):
I like it, definitely want to use the plugin extras functionality for my own plugins when I find time to work on them. Found what I think is a grammar error, ignore it if its not. Documenting the pluginExtras functionality would also be nice as it seems like something plugin authors are going to want to interact with.

packages/lib/plugin.js
Hornwitser: Yes, that's better.

packages/lib/plugin.js
Hornwitser: Unfortunately there isn't a document describing developing for the web interface yet. It's something I've put off writing because it was changing frequently.

Danielv123: Makes sense. If you feel like its still not set in stone I am not opposed to leaving it undocumented.

Hornwitser: I will leave it for now. There's still some things in the web plugin model I want to think through more on.

psihius (approved):

PR #456, opened 2022-03-29
Save transfer

Hornwitser:
Implement rename, copy, and cross instance/slave save transfer. Copy, rename and local transfer is implemented via simple file operations while the cross slave transfers use the proxy stream system. For organization sake the save related commands are grouped under `ctl instance save` now.


psihius (approved):
LGTM

Issue #319, opened 2020-09-21
Commit package-lock.json


Danielv123:
As per npm guidelines. We originally decided not to do this due to polluting the repo history.

Hornwitser:
The advantages as far as I can see with committing package-lock.json are
- Faster first install in dev environment.
- Faster install in ci environment.
- Reproducible package tree installed.

The disadvantages are
- A chore to update, as there's new dependencies updated weekly.
- Bloat, it's 14k lines and over half a megabyte in size.
- Not in sync with user install.  Unless we publish a shinkwrap file, users installing through npm is going to get another set of packages installed than the dev environment.
- Not solving a problem we have.  We don't care if we get updated dependencies, for the most part it just works.

chpatton013:
The reproducibility is the big advantage. Faster install in CI is nice and all, but you can get that by distributing images instead.

As for the disadvantages, I don't think it's so bad:

> A chore to update, as there's new dependencies updated weekly.

You would want to add a CI gate to verify that it gets updated before accepting PRs that change the `package.json`.

> Bloat, it's 14k lines and over half a megabyte in size.

It's ASCII, and it's structured data. Git is going to compress that remarkably well.

> Not in sync with user install. Unless we publish a shinkwrap file, users installing through npm is going to get another set of packages installed than the dev environment.

If you have a packaging process it's not hard to add the shrinkwrap command. If you don't, then all the more reason to formalize it!

> Not solving a problem we have. We don't care if we get updated dependencies, for the most part it just works.

Or preventing a problem you haven't discovered yet.

Hornwitser:
Another issue, if we move to pnpm as per #458 then commiting the lockfile becomes problematic as developing on external plugins will only work if the external plugins's dependencies are included in the lockfile (this has to do with other approaches ending up with multiple installations of webpack that gets confused with each other and break the live web build feature).

I don't know of any other way to make external plugin development work along with a development repository of Clusterio, if I did I'd like to commit the lockfile.

Danielv123:
If its required to solve the webpack instances problem then I say we go for it. We can still shrinkwrap the npm packages, right? Since people no longer install from the repo this should only affect developers which can fix dependency issues anyways.

PR #458, opened 2022-06-05
Replace npm and lerna with pnm

Hornwitser:
I'm not happy with how badly npm and lerna have been working. So let's try pnpm instead.

Unfortunately it seems like we wont be able to make live web builds work without having webpack installed in one location only, and that only happens in pnpm if plugins are part of the same workspace as the clusterio repository, so to make that happen external plugins need to be placed in the `external_plugins` folder from now on when developing. But if that is done then the lock file can't be commited because it contains a bunch of external plugins in it.


Danielv123 (approved):
Looks good to me. Resolves #319 . Hopefully we won't have any more webpack development issues now. Nice touch adding a readme to the external_plugins folder

PR #462, opened 2022-06-22
Minor refactor

Hornwitser:
Move a good deal of the code in slave.js and ctl.js out of their respective source files and into separate and more reasonably sized source files.

Also minor bugfix to start buttons in the instance views because I can't be bothered to rebase it out for this.


psihius:
Codecov is complaining, coverage is slipping. It would be nice to keep it in a more happy state ;)

psihius (approved):

PR #467, opened 2022-07-15
User UI update

Hornwitser:
Show ban reason in the user view page, hide switches for admin/whitelist when not having the permission to use them and fix some minor graphical issues on the page.


Hornwitser:
Err, I guess? It's not like it's likely to randomly stop working.

psihius:
> Err, I guess? It's not like it's likely to randomly stop working.

We are developers. We have seen weirder :D

PR #478, opened 2023-02-02
Add mod pack oriented mod management

Hornwitser:
Add mod management oriented around mod packs which contain the mod versions and mod settings to start instances using the given mod pack with. Mods and mod packs are stored and managed on the master server. The mod packs themselves can be exported as a mod pack string, [see the proposal on the forum for the format](https://forums.factorio.com/103578) making it easy to share them.

On instance start the mod pack configured for the instance is loaded from the master server and mods/ folder for that instance is entirely replaced with the mods specified in the mod pack. A mod-settings.dat file also generated from the mod settings specified in the mod pack. Exporting data from the game is tied to mod packs and on export settings prototypes is used to fill in available mod settings with their default values. This export is also used to render the mod settings dialog similarly to how it appears in the game.

The code is structured to allow slaves to download mods they are missing from the master server but this is not implemented yet. This means than only the standalone installation works properly. For remote slave you will currently have to copy the mods over manually.


psihius:
Oof, this one is gonna take a while to review :fearful:

packages/lib/data/ModInfo.js
```diff
@@ -0,0 +1,231 @@
...
+	 * @type {string}
+	 */
+	description = "";
+
+	/**
+	 * Major version of Factorio this mod supports.
+	 * Sourced from info.json.
+	 * @type {string}
+	 */
+	factorioVersion = "0.12";
```
psihius, line 84: Why 0.12? We are far far far away from the time 0.12 was last relevant :D

psihius (approved):
Merge it :+1:

packages/lib/data/ModInfo.js
Hornwitser: It's what the game interprets a mod without `factorio_version` as being for.

PR #480, opened 2023-02-21
Small things 10

Hornwitser:
More isolated tasks from my todo list. See individual commits for changes and rationale for the changes.

### DIFF
--- a/packages/lib/factorio/patch.js
+++ b/packages/lib/factorio/patch.js
@@ -251,8 +251,14 @@ async function patch(savePath, modules) {
 	// Write back the save
 	let tempSavePath = `${savePath}.tmp`;
 	let stream = zip.generateNodeStream({ compression: "DEFLATE" });
-	let pipe = stream.pipe(fs.createWriteStream(tempSavePath));
-	await events.once(pipe, "finish");
+	let fd = await fs.open(tempSavePath, "w");
+	try {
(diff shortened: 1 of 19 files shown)


packages/lib/factorio/patch.js
```diff
@@ -251,8 +251,14 @@ async function patch(savePath, modules) {
 	// Write back the save
 	let tempSavePath = `${savePath}.tmp`;
 	let stream = zip.generateNodeStream({ compression: "DEFLATE" });
-	let pipe = stream.pipe(fs.createWriteStream(tempSavePath));
-	await events.once(pipe, "finish");
+	let fd = await fs.open(tempSavePath, "w");
+	try {
+		let pipe = stream.pipe(fs.createWriteStream("", { fd, autoClose: false }));
+		await events.once(pipe, "finish");
+		await fs.fsync(fd);
```
psihius, line 258: You don't need this line - the `finally` block is always executed.

psihius (approved):
Except for that one feedback comment, looks good.

packages/lib/factorio/patch.js
Hornwitser: I am confused by what you mean with this. Which line? The fsync one?

psihius: @Hornwitser nevermind, i'm an idiot. Lack of sleep today made me thing both lines where a `.close()`

Issue #189, opened 2019-08-02
Consider renaming master/slave terminology


clarfon:
In addition to being a rather awful connection to… well, slavery, it's a bit confusing what the actual servers actually mean.

If we were to simply rename based upon the server architecture, controller/worker might be a better set of names. But honestly, I feel like calling the master servers clusters and the slave servers worlds might be a better name.

A clusterio world is the connection between a factorio server and a clusterio cluster.

Danielv123:
Sorry, but this is common terminology which is in no way confusing - https://en.wikipedia.org/wiki/Master/slave_(technology)

I disagree that they have to be renamed.

psihius:
Note: This is my personal opinion.

While I understand the general direction and this is not my first encounter with this messaging, I have to disagree strongly in this case.
What many native English speakers or people who know English on a high level do not understand, most of our community are not English natives by far.
Second - master/slave server architecture pertains to a very specific subset of implementation, that pertains to somewhat low-level implementation details and when you are working internationally, this is a very well known technical term pointing to a very specific thing that is universally understood by engineers.

Controller, primarily, is a WEB development terminology. We, actually, have those in our Master server.
Worker - is a thread/queue-based processing client.

So as far as I'm concerned, and having to wrangle a 50 man team with biggest Clusterio events in the Factorio community, where there are more English non-natives than actual English native speakers, especially when actual developers are concerned (we have a few very good devs who have somewhat shoddy English, but they do some of the more crazy things), I'm against changing terminology that is 100% on point, everybody knows, does not need explaining and there is no good replacement for it.
Unless someone can come up with replacement terminology, that is easy, concise and has precise meaning - No from me.

clarfon:
In response to both… I understand the desire to use simple terms to help make it easier to work with people whose native languages aren't English. And, I understand that this is the standard set of terms that the industry uses.

In the linked Wikipedia article, it spends an entire section talking about how more and more people are moving away from these terms due to their roots in slavery. Also, although these terms have very clear definitions in very distinct places (e.g. in the context of databases, master/slave *always* refers to primary/replica; in the context of build systems, master/slave *always* refers to controller/worker; etc.) the terms in general themselves *do not* have any distinct definitions across all of these fields.

I strongly disagree that controller *exclusively* refers to MVC in the context of model-view-controller, and that worker *exclusively* refers to queue-based processing. While these are the most common uses of these terms, there are in fact many different terms used in software because a lot of these patterns intersect.

[...]

PR #481, opened 2023-02-22
Rename master/slave terminology to controller/host

Hornwitser:
The role of the master server as the party controlling every part of the cluster and slaves as hosts for Factorio servers are not well reflected in their name and has caused confusion among users on several occasions. Rename these to controller and host respectively to better reflect their roles.

People also complained about the naming see #189 for an example.


psihius (approved):

psihius:
Some people on the internet will be happy

PR #485, opened 2023-04-22
Data class refactor

Hornwitser:
Refactor serialisation and deserialisation to use a JSON data class pattern. The basic structure of such a class is a static `jsonSchema` property with the schema for the serialised data from the class, a static `fromJSON` method that takes serialised data and turns it into an instance of the class and an optional `toJSON` method that turns an instance of the class into serialised data. See the following example:

```js
/**
 * A Data class example
 * @alias module:Data
 */
class Data {
    /** @type {string} */
    member = "foo";

    constructor(member) {
        if (member !== undefined) { this.member = member; }
    }

    static jsonSchema = {
        type: "object",
        properties: {
            "member": { type: "string" },
        },
    };

    static fromJSON(json) {
        return new this(json.member);
    }

    // toJSON is called by JSON.stringify when serialising this class
    toJSON() {
        let json = {};
        // Omit property if it's set to the default value
        if (this.member !== "foo") { json["member"] = this.member; }
        return json;
    }
}
```

The benefits of these data classes are
- Logic for serialisation, deserialisation, and the data schema is kept right next to each other.
- Using classes for data allows methods directly related to the data to be defined on it, and for it to be inspected at runtime.
- The constructor is free to take whichever arguments are most convenient for use in the code.
- The serialisation and schemas easily compose with other data classes. For example the jsonSchema of another class can directly use Data.jsonSchema, and its fromJSON can use Data.fromJSON.
- Data can be validated by running it through a schema validator
- The format is friendly for a conversion to TypeScript. If it's combined with Typebox the JSON can also be type checked.

The major downside is that it's verbose to type out.

There has been a few variations in how data is serialised, but I hope that by using this all those inconsistent approaches can be unified in a way that composes well.

A major refactor to the link system was required to make it able to fully take advantage of the data classes. The most notable changes are the reversal of the direction of the API, instead of calling send on a message with the link as the argument, you now call send on the link with an instance of the message as the argument.

Since compatibility in the link system was broken with this refactor I used this opportunity to fix a good deal of quirks with how it worked. Most notably the removal of magic registration of event handlers, and a much simpler to understand end-to-end routing system. (See the link system refactor commit it for more.)


psihius (commented):
Looks great, I like it :)

packages/host/host.js
```diff
@@ -51,15 +52,15 @@ class HostConnector extends libLink.WebSocketClientConnector {
 			plugins[pluginInfo.name] = pluginInfo.version;
 		}
 
-		this.sendHandshake("register_host", {
-			token: this.hostConfig.get("host.controller_token"),
-			agent: "Clusterio Host",
+		this.sendHandshake(new libData.MessageRegisterHost(new libData.RegisterHostData(
```
psihius, line 55: In cases like these, I prefer the arguments to be formatted so you can see what gets created/passed like this - it makes things far more readable at a glance

`this.sendHandshake(
    new libData.MessageRegisterHost(
        new libData.RegisterHostData(
			this.hostConfig.get("host.controller_token"),
			"Clusterio Host",
			version,
			this.hostConfig.get("host.name"),
			this.hostConfig.get("host.id"),
			this.hostConfig.get("host.public_address"),
			plugins,
		)
    )
);`

packages/host/src/Instance.js
```diff
@@ -284,9 +301,9 @@ class Instance extends libLink.Link {
 			instance_id: this.id,
 			type: "join",
 			name,
-			stats: stats.toJSON(),
+			stats,
 		};
-		libLink.messages.playerEvent.send(this, event);
+		this.sendTo(new libData.InstancePlayerUpdateEvent("join", name, undefined, stats), "controller");
```
psihius, line 306: Same here on formatting, the second variant in this case is not really needed - I usually chunk that when it hits the 120 characters on the line - that's when the code line just gets a bit too long
```
this.sendTo(
    new libData.InstancePlayerUpdateEvent("join", name, undefined, stats), 
    "controller"
);
```
or
```
this.sendTo(
    new libData.InstancePlayerUpdateEvent(
        "join", 
        name, 
        undefined, 
        stats
    ), 
    "controller"
);
```

packages/host/src/Instance.js
```diff
@@ -371,24 +391,24 @@ rcon.print(game.table_to_json(players))`.replace(/\r?\n/g, " ");
 				type = "special";
 			}
 
-			list.push({
-				name,
+			list.push(new libData.SaveDetails(
```
psihius, line 394: Same here - when passing a structure - the opening `{` on the same line works well. But when we are talking creation of a new object, I feel it's best to have it on its own line
```
list.push(
    new libData.SaveDetails(
        type,
		name,
		stat.size,
		stat.mtimeMs,
		name === loadedSave,
		name === defaultSave,
	)
);
```

psihius: See line 438 of this file where it's done this way already :)
Maybe there is a style linter/fixer tool that can handle stuff like this? We have CS-Fixer in PHP that handles things like this with breeze.

psihius (approved):
Overall looks good.
The node 12 tests pipeline is failing though, but since we talked about bumping the minimal version, I think we can ignore it. We need to add node 18 and node 20 versions to the test list since those are 2 current major LTS versions. 16 version also loses its security support in September 2023, so I don't know if it's even worth it to keep it around too.

packages/host/host.js
Hornwitser: I agree these nested class construction are not very readable, so I've changed the ones I could find with grep.

packages/host/src/Instance.js
Hornwitser: I haven't found any formatters that isn't going to reformat everything to what it thinks is the way to do it and ESLint does not seem to have any rules for this specific case (maybe there's a plugin for that?) This style is used in quite a few places, like the commands in ctl, so changing it I think should be a separate issue/PR.

packages/host/src/Instance.js
Hornwitser: Perhaps the argument order for sendTo should be reversed?
```js
this.sendTo("controller", new libData.InstancePlayerUpdateEvent("join", name, undefined, stats));
```

To me something like this also flows nicely when reading it.
```js
this.sendTo("controller",
    new libData.InstancePlayerUpdateEvent("join", name, undefined, stats)
);
```

packages/host/src/Instance.js
psihius: At least you set it up once and then just use it to ensure code style.
You can play with it and see what results it provides and settle on the code-style rules. Auto-formatting helps that it can be run automatically and fix most things without the user even intervening :)

PR #487, opened 2023-07-27
Bump supported Node.js to v18 and update dependencies

Hornwitser:
The LTS releases for v12 and v14 are already end of life and the v16 is end of life in under two months from now on.  Bump supported version up to v18 as it does not make sense to continue to support EOL releases. Even if we wanted to keep supporting them, our dependencies will not.

**Note:** Node.js v18+ is currently not avaliable via the NodeSource PPA on Ubuntu 19.10 and below or Debian 9 and below.


psihius (approved):
Looks good, pipeline tho failing, but looks like it's an infra problem.

PR #488, opened 2023-07-28
Refactor lib

Hornwitser:
Export all the symbols in lib in a single flat namespace. This simplifies using the package and makes reorganising it easier as consumers of it does not need to updated whenever anything moves around internally. This enabled me to move all the source files into a src/ folder like the structure of the other packages uses.

Some unnecessary code was moved out of lib to their respective packages. This helps reduce the startup time for ctl.


psihius (approved):
Merge it

PR #492, opened 2023-08-10
Update README.md

ndisalvio3:
Fix naming schema


Hornwitser:
The controller and host names are the correct names. They were renamed in #481 but this change hasn't been put in a release yet which is why the names don't match with the latest released version.

Issue #491, opened 2023-08-06
Migrate to TypeScript


Hornwitser:
A lot of coding errors result for type errors that could have been caught if TypeScript was used. Having types available would also make code completion work better. It is desirable to migrate the code over to TypeScript.

- [x] Set up a shared JavaScript/TypeScript environment where files can be ported over one at a time
- [x] Port packages/lib
- [x] port pagkages/controller
- [x] port packages/host
- [x] port packages/ctl
- [x] port packages/web_ui
- [x] port plugins/global_chat
- [x] port plugins/inventory_sync
- [x] port plugins/player_auth
- [x] port plugins/research_sync
- [x] port plugins/statistics_exporter
- [x] port plugins/subspace_storage

nitin-pandita:
Hey @Hornwitser  i would like to work on it

nitin-pandita:
> That would be very helpful. Would you mind waiting until I've ported packages/lib to TypeScript and have it merged first? Shouldn't take more than a couple of days at most, and then we can take a package each to port at a time.

sure

Hornwitser:
@nitin-pandita and @Luciole0x I have merged the porting of the lib package now so it should be fairly straight forward to pick a package and start porting it now. See the lib package for how I set up the tsconfig and prepare script step.

Luciole0x:
Ok, I will start working on the controller package.

Hornwitser:
I will work on the ctl package.

nitin-pandita:
I will work on web_ui package

Hornwitser:
There were some problems with the typings I made for the lib package which meant that Link.handle and Link.send/sendTo would result in a type error instead of correctly deducing the types. These should be fixed by #494.

Hornwitser:
I will work on the host package.

Luciole0x:
I will work on the global_chat plugin (all included plugins need the transition to, right ?).
Regarding code coverage, do you have some kind of "philosophy" on how you implement the tests ?
I may try my hand on making some but as I'm still goofing a lot, may prove not too productive... we'll see.

Hornwitser:
If you're not changing the logic of the code you do not need to concern yourself about the coverage. So if the existing tests are doing the same thing after it ports cleanly to TypeScript then that's good enough.

Luciole0x:
I will work on the inventory_sync plugin.

Hornwitser:
@nitin-pandita how is the progress with porting web_ui going?

[...]

PR #495, opened 2023-08-13
Transition controller package to TypeScript.

Luciole0x:
Part of #491

Focusing on the Controller package.

Some issues to understand parts like src/routes.ts:getMetrics()  or  how lib.BaseControllerPlugin use is metrics in src/Controller.ts:880. And some guessing here and there on the right type when not able to extrapolate the right match.

Test are broken not expecting controller to behave as a module getting "Cannot use import statement outside a module" errors.
Not sure how to handle this to be honest.


packages/controller/src/UserManager.ts
```diff
@@ -72,11 +72,9 @@ export default class UserManager {
 		}
 
 		let defaultRoleId = this._config.get("controller.default_role_id");
-		if (defaultRoleId === null) {
-			throw new Error(`controller.default_role_id is not defined.`);
-		}
+		let roles = defaultRoleId ? [defaultRoleId] : [];
```
Hornwitser, line 75: Note that defaultRoleId can be set to 0, which is the default admin role. If you intended for that value to be ignored you should add a comment saying so, otherwise compare it with null.

Hornwitser:
The tests are still failing, looks like there's a missing export in lib/src/config/classes.ts. Also keep in mind when importing a TypeScript module from JavaScript the default export becomes named `default`, so that means `const SomeClass = require("path/to/SomeClass");` needs to be changed to `const SomeClass = require("path/to/SomeClass").default;`.

Hornwitser:
The tests are still failing.

Luciole0x:
I try to figure out why the test don't show the same results when run locally (I got 26 errors).
Part of it may be running on Windows as some errors show path with a mix of \\ and /
I fell like I'm missing some default files/settings or something ?
For exemple on the test "bootstrap generate-user-token" > "work for existing user" where the test user come from, I see all the info in the MockController of "test/mock.js", but I don't see a link with the test executing "node ../../packages/controller bootstrap generate-user-token test".

Hornwitser (approved):
Looks good now. I will look into why tests are failing on Windows later.

PR #496, opened 2023-08-15
Fix tests on windows

Hornwitser:
Fix minor inconsistencies causing tests to fail on Windows.

### DIFF
--- a/packages/controller/src/Controller.ts
+++ b/packages/controller/src/Controller.ts
@@ -87,7 +87,7 @@ export default class Controller {
 	
 	logDirectory: string = "";
 	clusterLogIndex: lib.LogIndex | null = null;
-	clusterLogBuildInterval: NodeJS.Timer | null = null;
+	clusterLogBuildInterval: ReturnType<typeof setInterval> | null = null;
 
 	constructor(
 		clusterLogger: winston.Logger,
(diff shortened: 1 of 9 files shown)


Luciole0x:
Work fine for me to the exception of a few issues with end of lines \n vs \r\n :
 - test/file/hash.txt: removing the new line and keeping only the "The quick brown fox jumps over the lazy dog"
 - test/integration/patch.js:37+38 : assert.equal(..., "-- test\n"); by assert.match(..., /^-- test\r?\n$/m);
 - test/lib/hash.js:14+21 : "be417768b5c3c5c1d9bcb2e7c119196dd76b5570" par "2fd4e1c67a2d28fced849ee1bb76e7391b93eb12" (caused by removing the end of line in hash.txt)

Luciole0x:
You're right, avoiding CRLF and LF shenanigans is for the best.
But turning autocrlf off doesn't expose to mixing them ? 
I would better like to try and prevent CRLF to leek from my side in any way, something like this should work ?:
   core.eol lf
   core.autocrlf input

Starting from a fresh copy after changing the settings, tests run without issues.

Hornwitser:
Sure, you can set it to input if you think you're going to accidentally introduce the wrong line endings.

PR #497, opened 2023-08-17
Port global_chat plugin to typescript

Luciole0x:
Part of #491

I tried to change as little as possible to the logic to allow the field main of the package.json of the plugins to define the info.js location while being retro-compatible.
Fields like `instanceEntrypoint` from info.js are relative to the root of the plugin not info.js location though, don't no if this is a issue.


Hornwitser:
We don't need to be backwards compatible in the plugin entrypoint logic. Alpha 14 will already require a significant rework of existing plugins to work at all so changing the entrypoint is just another point on the migration list. You also should not be trying to resole package.main yourself; `require(pluginPath)` will load what's specified in the `"main"` field in `package.json`, and if that field isn't present it'll load `index.js`.

Hornwitser (approved):
Looks good now.

PR #499, opened 2023-08-18
Port remaining plugins to Typescript

Luciole0x:
Part of #491

Have issue with implementing the lib.Request<req, res> interface, is there an example somewhere ?


Hornwitser:
All the *Request classes in lib/data/message_*.js implements it.

Luciole0x:
I saw them, but the interface isn't enforced, no ?
As long as I don't try to explicitly call something who require an instance of the interface, Typescript don't tell if it's rightfully implemented or am I missing something ?

Hornwitser:
Yes, that is correct.

plugins/statistics_exporter/instance.ts
```diff
@@ -1,7 +1,23 @@
...
+		[key:string]: {
+			[key:string]: IpcFlowType,
+		}
+	},
+}
+
+type IpcFlowType = {
+	"input": number,
+	"output": number,
+}
```
Hornwitser, line 20: The shape of the "input" and "output" entries is an object where string keys map to numbers. i.e. `"input": Record<string, number>`.

Hornwitser (approved):
Very nice. I noticed one type error in the statistics exporter plugin.

PR #501, opened 2023-08-25
Module locale

### LINKED ISSUE #500: Allow locale files for modules
tl;dr In order to provide translations for modules a locale file is required. Currently there is no method for inserting locale files into a scenario.

## Description 
Factorio uses locale files to provide translations for mods and scenarios. Mods do not need to be considered here because they are handled directly by factorio. However, there is currently no way to include these files from a module into the final scenario because injection is controled by clusterio. Therefore, a method needs to be defined for how locale files can be injected.

## Options

1) In the root of a module there is a directory named `locale` which has its contents recursively copied into the scenario locale directory. If there is name conflict then the plugin name is added as a prefix. This prefix can be included on all files as a logic simplification because locale file names do not have any impact. This would be the most intuitive method and the simplist to impliment. Eg, `module/locale/en/foo.cfg` -> `scenario/locale/en/my-module-foo.cfg`

2) It is unlikely for a module to ever require more than one cfg file per language; therefore, the name of the cfg file could be converted to the locale name and the extra directory depth removed. This also removes the need for recersive coping and allows for a whitelist of allowed language files. This would be my chosen solution because it is clean and can enforce that only factorio supported langauges get copied (emiting a warning if an unknown one is inlcuded). Eg, `module/locale/en.cfg` -> `scenario/locale/en/my-module.cfg`

3) Same logic as 2 but all langauge files are placed in the root of the module, and only cfg files are copied. This would just be messy imo. Eg `module/en.cfg` -> `scenario/locale/en/my-module.cfg`

4) A new property named `locale` is added to `module.json` which lists the supported locales and lists the required locale files to be copied for each. This is the most complete option, as it allows multiple files for a language and allows for those files to be placed anyway the developer intends. However. I feel this is over complex and unnecessary as such support is unlikely to be used. Eg, `"locale": [ { "langauge": "en", "files": "en.cfg" } ]` + `module/en.cfg` -> `scenario/locale/en/my-module.cfg`

Hornwitser:
Refactor module patching logic to implement mapping of locale files located at locale/{lang}/{file} in a module to locale/{lang}/{module-name}-{file} in the save. This additionally support a shorthand notation by mapping locale/{lange}.cfg to locale/{lang}/{module-name}.cfg.

Resolves #500.

### DIFF
--- a/packages/controller/src/Controller.ts
+++ b/packages/controller/src/Controller.ts
@@ -164,7 +164,7 @@ export default class Controller {
 			const webpackConfigs = [];
 
 			if (args.dev) {
-				webpackConfigs.push(require("../webpack.config")({}));
+				webpackConfigs.push(require("../../../webpack.config")({}));
 			}
 			if (args.devPlugin) {
 				let devPlugins = new Map();
--- a/packages/host/src/Instance.ts
+++ b/packages/host/src/Instance.ts
@@ -9,7 +9,7 @@ import type { Static } from "@sinclair/typebox";
 import * as lib from "@clusterio/lib";
 
 import { FactorioServer } from "./server";
-import { ModuleInfo, patch } from "./patch";
+import { SaveModule, patch } from "./patch";
 import { exportData } from "./export";
 import type Host from "./Host"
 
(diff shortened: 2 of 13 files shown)


Cooldude2606 (approved):
I have reviewed the tests and they match the requirements of request #500

PR #502, opened 2023-08-25
Fix issue with research_sync plugin

Luciole0x:
Fix type issues when loading technologies and broadcasting progress.
Locally tested;
Researches sync when adding a new instance.
Was able to research on one instance and see progress from another instance while searching a second or same research.
No more errors in logs.

### DIFF
--- a/plugins/research_sync/controller.ts
+++ b/plugins/research_sync/controller.ts
@@ -1,5 +1,6 @@
 import fs from "fs-extra";
 import path from "path";
+import { Static } from "@sinclair/typebox";
 
 import * as lib from "@clusterio/lib";
 const { RateLimiter } = lib;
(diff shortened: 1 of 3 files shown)


Hornwitser:
Isn't this still broken though? If someone updates from alpha.13 the technologies.json file will contain data of the form
```ts
type TechnologiesJson = [string, { level: number, progress: number | null, researched: boolean }][]
```
But your updated code expects technologies.json to contain data of the of the form
```ts
type TechnologiesJson = [string, [string, number, number | null, boolean]][]
```

If you want to change the format of data stored on disk you need to write code that can load both the old and new format. But I don't see a good reason to change the format here.

Luciole0x:
Looking at the previous changes I made I see what I misunderstood. I was using technologies Map as `Map<string, Technology>` where it was `Map<string, {  level:number, progress:number|undefined, researched:boolean }`.
The different Technology like type across confuse me.
I will add a distinct type, to make it work like before.

psihius (commented):
The code looks good to me, I do have one request - can we try and update the tests please and cover the changes? 

I think it's time to put some effort into updating and improving the tests so we don't degrade the coverage.

Hornwitser (approved):

Hornwitser:
I think that's a separate effort to this PR. The harness needed to test research_sync properly is non-trivial to write.

PR #503, opened 2023-08-25
fix inventory_sync plugin 

Luciole0x:
Fix an issue with player_data / playerData property.

### DIFF
--- a/packages/lib/src/config/definitions.ts
+++ b/packages/lib/src/config/definitions.ts
@@ -157,6 +157,7 @@ export class ControllerConfig extends classes.Config {
 	get(name: "controller.proxy_stream_timeout"): number;
 	get(name: "controller.default_mod_pack_id"): number | null;
 	get(name: "controller.default_role_id"): number | null;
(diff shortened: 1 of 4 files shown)


psihius (approved):
Seems okay to me, @Hornwitser your turn :)

Hornwitser (approved):
As a band aid this works, but there's clearly a deeper issue with the typings of plugins that need to be fixed. I'll look into that.

Luciole0x:
Yes most of the issues comes from the `type ___ = any` in places like lib/src/plugin.ts but we cannot import the types from the other packages as it will create some circular dependency on the types and `tsc --build` fail as the output will overwrite .d.ts files used as input.
One possible solution may be to create the interfaces inside the lib package and make the implementations use those interfaces to prevent out of sync in the others packages.

Hornwitser:
I was thinking about moving the instance base plugin to the `@clusterio/host` package. That way it could make use of the types from that package, but I'm not sure if that will create other issues.

PR #505, opened 2023-08-26
Add module export detection

Cooldude2606:
Allows modules to include a `module_exports.lua` file in their root which will cause the file `save/modules/module_name.lua` to be generated pointing to the exports file.

The justification for this is to encourage code reuse between modules by explicitly stating their exports and making them easy to import in another module. If `myPlugin` in includes an export file, then other modules can include it through `require("modules/myPlugin")`.

The file name `module_exports.lua` was selected to not conflict with `api.lua` of the clustorio lib and to match the style of commonjs modules with `module.exports =`.

### DIFF
--- a/test/host/patch.js
+++ b/test/host/patch.js
@@ -28,6 +28,29 @@ describe("host/patch", function() {
 				}
 			});
 		});
+
+		describe("loadFiles()", async function() {
+			const referenceFiles = new Map([
+				["modules/test/test.lua", "-- test\n"],
+				["modules/test/module_exports.lua", "-- module_exports\n"],
+				["locale/en/test.cfg", "module-test=A Test\n"],
+				["local/en/test-locale.cfg", "module-test-locale=Test Locale\n"],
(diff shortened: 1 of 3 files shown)


test/host/patch.js
```diff
@@ -28,6 +28,29 @@ describe("host/patch", function() {
 				}
 			});
 		});
+
+		describe("loadFiles()", async function() {
```
Hornwitser, line 32: describe can't take an async function. You will need to either move the setup to an async `before(async function() { ... })` hook, or move the setup inside an async `it("...", async function() { ... })` test.

Hornwitser (changes requested):

test/host/patch.js
Cooldude2606: Typo has been fixed, async has been moved into a before hook, found and fixed error with the contents of the export file.

Hornwitser (approved):
Nicely done.

PR #509, opened 2023-09-04
make events and requests implement there respective interface

Luciole0x:
Modified Request, Event and related interfaces to allow implements.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@clusterio/ctl": "workspace:*",
     "@clusterio/host": "workspace:*",
     "@clusterio/lib": "workspace:*",
+    "@sinclair/typebox": "^0.30.4",
     "eslint": "^8.45.0",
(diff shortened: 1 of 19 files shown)


Hornwitser:
What is the point of doing this?

Luciole0x:
To allow Typescript to tell when an event or request is missing something while implementing it.
If the structure of Event or Request is changed be sure to not miss to update one of the related classes.
What the point of the interface if it's never used ?

Hornwitser:
The interface is already being used and it's already being checked when sending a message over a link and when adding a handler for a message. If you omit a required element then either typescript will tell you it's missing, or the code will throw an error on startup.

I would very much want to avoid having to add `implements Request<RepatedClassName, RepatedResponseName>` to every request and event class if it can be avoided. Same goes for having to move the definition of the Response class out of the Request class, and making the type any more complicated than it needs to be.

PR #511, opened 2023-09-06
allow to revoke host tokens from web ui

Luciole0x:
Linked to https://github.com/orgs/clusterio/projects/2?pane=issue&itemId=21054425

### DIFF
--- a/packages/lib/src/users.ts
+++ b/packages/lib/src/users.ts
@@ -376,6 +376,11 @@ definePermission({
 	description: "Generate tokens for hosts to connect to the cluster with.",
 });
 definePermission({
+	name: "core.host.update_access",
+	title: "update host access",
+	description: "Revoke or restore a host access to the cluster.",
+});
+definePermission({
 	name: "core.host.create_config",
 	title: "Create host config",
 	description: "Create host configs via the controller.",
--- a/packages/lib/src/users.ts
+++ b/packages/lib/src/users.ts
@@ -378,7 +378,7 @@ definePermission({
 definePermission({
 	name: "core.host.revoke_token",
 	title: "Revoke host token",
-	description: "Revoke access token and terminate all existing control connections.",
+	description: "Revoke access token and terminate existing connection for hosts.",
 });
 definePermission({
 	name: "core.host.create_config",
--- a/packages/lib/src/users.ts
+++ b/packages/lib/src/users.ts
@@ -376,9 +376,9 @@ definePermission({
 	description: "Generate tokens for hosts to connect to the cluster with.",
 });
 definePermission({
-	name: "core.host.update_access",
-	title: "update host access",
-	description: "Revoke or restore a host access to the cluster.",
+	name: "core.host.revoke_token",
+	title: "Revoke host token",
+	description: "Revoke access token and terminate all existing control connections.",
 });
 definePermission({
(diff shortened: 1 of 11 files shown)


Hornwitser:
This is not how revoking a token should work and this does not implement what the backlog item asked for. If you think what you implemented is a better solution _you should describe in your PR why you think so, and why you decided not to implement what was asked for._

The relevant details here are
- The backlog item asked for a "tokenValidAfter" timestamp that can only be set to the current time to irreversibly revoke all existing tokens for the host, but you have implemented a "disallow this host from connecting" flag that can be set or cleared.
- A "disallow this host from connecting" flag is not useful in the case of a token leaking because while the flag is set no tokens can be used to connect as that host, and while the flag is cleared any tokens for the host including the leaked one can be used to connect as that host. There's no way to invalidate the existing tokens and reissue a new one.
- The way this implementation terminates the existing connection to the host being revoked access to allows for a malicious host to keep the connection open indefinitely because it can simply ignore the prepare disconnect message.

Please have a look at how the UserRevokeToken functionality is implemented for revoking user tokens and terminating existing connections from that user.

Luciole0x:
Ok, my bad, I completely drifted away and miss the point.
Hopefully, these changes should make it more aligned with what requested in #512 .

Cooldude2606:
I find the term `tokenValidAfter` to be a confusing one in this context, it reads as if it is describing a single token which can not be used before a given time, matching how `nbf - Not Before` works for JWT. I would like to suggest `tokensIssuedAfter` as it is more descriptive of the token property being checked.

packages/lib/src/users.ts
```diff
@@ -375,6 +375,11 @@ definePermission({
 	title: "Generate host token",
 	description: "Generate tokens for hosts to connect to the cluster with.",
 });
+definePermission({
+	name: "core.host.revoke_token",
+	title: "Revoke host token",
+	description: "Revoke access token and terminate all existing control connections.",
```
Hornwitser, line 381: This should say "terminate existing connecting for the host"

packages/web_ui/src/components/HostViewPage.tsx
```diff
@@ -34,9 +40,31 @@ export default function HostViewPage() {
 		</PageLayout>;
 	}
 
-	return <PageLayout nav={nav}>
-		<PageHeader title={host.name || String(hostId)} />
+	let hostButtons = <Space> {
+			account.hasAnyPermission("core.host.revoke_access", "core.admin") &&
```
Hornwitser, line 44: You do not need to check for "core.admin" when checking for another permission, it is special cased and having it makes all permission checks succeed.

Hornwitser (approved):
Looks good. Thank you for your contribution and appologies for the late review.

Hornwitser:
> I would like to suggest `tokensIssuedAfter` as it is more descriptive of the token property being checked.

To be honest I am not sure I like either of `tokenValidAfter` and `tokensIssuedAfter`. It's supposed to describe "tokens issued before this point in time are invalid". Maybe `tokensInvalidBefore`?

Either way, feel free to open PR renaming the field for both hosts and users after this one is merged (though you might want to add migration logic for the user one).

Luciole0x:
> appologies for the late review
No problem.

PR #518, opened 2023-11-20
Remove channels to simplify subscription

Hornwitser:
Make subscriptions either on or off to greatly simplify the logic of subscribing to events.


Cooldude2606 (commented):
This all looks good to me. Doing the filtering within each update handler is much simpler than anything I was thinking of.

My only comment is around `EventData.subscriptions` and its properties: `all` will never be `false` using the new logic (before it could be false if there was a channel subscription). Because of this we should be able to change the type to `Map<Link, () => void>` and not worry about any other properties.

Hornwitser:
Superseded by #534

PR #519, opened 2023-11-24
Small things 11

Hornwitser:
Various small fixes, see commit log for details.


packages/lib/src/subscriptions.ts
```diff
@@ -365,7 +365,7 @@ export class EventSubscriber<T, V=T> {
 		));
 
 		if (response.eventReplay) {
-			await this._handle(response.eventReplay);
+			await this._handle(response.eventReplay as any);
```
Cooldude2606, line 368: This is defiantly the simplest solution. In the future we might need to review the type inference of `SubscriptionRequest` and `SubscriptionResponse` to make this work without casting to `any`. (No action needed)

Cooldude2606 (approved):
Everything looks good, only one comment for an improvement we can do later.

PR #520, opened 2023-11-25
Add VSCode workspace config

Hornwitser:
Make developing with VSCode slightly nicer by configuring the editor to use the right line endings and indents by default.

This requires changing all the JSON files to use a consistent indenting, as a per path setting is not possible in VSCode (nor does it make much sense). I changed the indents on the JSON files to all use tabs to be consistent with the JavaScript code.

### DIFF
--- a/.vscode/.gitignore
+++ b/.vscode/.gitignore
@@ -1,0 +1,3 @@
+*
+!.gitignore
+!settings.json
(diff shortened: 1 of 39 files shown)


Cooldude2606 (approved):
LGTM.

PR #524, opened 2023-11-28
Port plugin web code to TypeScript

Hornwitser:
Port the web interface code in plugins to TypeScript. This solves a lot of issues with using TypeScript in the web interface by putting it into the root build script and directly referencing the TypeScript sources in configs and imports.


packages/lib/scripts/compile_validator.js
```diff
@@ -22,7 +22,11 @@ const ajv = new Ajv({
...
-	`// generated by scripts/compile_validator.js\n// @ts-nocheck\n/* eslint-disable */\n${moduleCode}`
-);
+function write(output) {
+	fs.outputFileSync(
+		path.join(__dirname, output),
+		`// generated by scripts/compile_validator.js\n// @ts-nocheck\n/* eslint-disable */\n${moduleCode}`
+	);
+}
+write("../dist/src/data/message_validate.js");
+write("../src/data/message_validate.js");
```
Cooldude2606, line 32: Where is `src/data/message_validate.js` used? From what I see tsc doesn't touch it because only `ts`, `tsx` and `d.ts` files are included. I also don't see any files directly requiring this file from the src dir.

packages/controller/web/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable arrow-body-style, no-console */
```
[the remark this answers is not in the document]
Hornwitser: The idea here was to do minimal changes so that git would properly detect the rename. I agree that it should be rewritten.

packages/lib/scripts/compile_validator.js
Hornwitser: It's required from `src/data/messages_core.ts`. It can't currently be included by `tsc` as that would cause a circular dependency loop when trying to create the file.

Cooldude2606 (approved):
LGTM. No changes which would block merge.

PR #527, opened 2023-11-29
Implement autosaving of data held in memory

### LINKED ISSUE #523: Autosave all data
The configuration and user lists are only saved when the controller is shut down, meaning they'll suffer data losses when the controller crashes. There should be a proper data saving system that:
- Provides a config option for how frequently data should be autosaved to disk.
- Provides a simple way for individual systems to determine if data has changed so that writes can skipped.
- Provides a global synchronous event which data to be written can be gathered from so that saving does not give inconsistent data.

Hornwitser:
Add controller.autosave_interval config that specifies the frequency which controller config, hosts, instances, users, mod packs and plugin specific data which is held in memory should be automatically flushed to disk if they have changed since the last time they were saved.  This prevents data loss in the event of a server crash.

Plugins can use the onSaveData hook to save its own data at the same time as the controller does and this obsoletes making an autosave interval in the plugin.

Resolves #523.

### DIFF
--- a/packages/controller/src/Controller.ts
+++ b/packages/controller/src/Controller.ts
@@ -720,8 +720,10 @@ export default class Controller {
 
 		let instance = new InstanceInfo({ config: instanceConfig, status: "unassigned" });
 		this.instances!.set(instanceId, instance);
+		this.instancesDirty = true;
 		await lib.invokeHook(this.plugins, "onInstanceStatusChanged", instance);
 		this.addInstanceHooks(instance);
+		this.instanceDetailsUpdated(instance);
 		return instance;
 	}
 
@@ -766,13 +768,14 @@ export default class Controller {
 
 		// Assign to target
 		instance.config.set("instance.assigned_host", hostId);
+		this.instancesDirty = true;
 		if (hostId !== null && newHostConnection) {
 			await newHostConnection.send(
 				new lib.InstanceAssignInternalRequest(instanceId, instance.config.serialize("host"))
 			);
 		} else {
 			instance.status = "unassigned";
-			this.instanceUpdated(instance);
+			this.instanceDetailsUpdated(instance);
 		}
 	}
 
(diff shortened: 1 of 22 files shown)


packages/controller/src/Controller.ts
```diff
@@ -361,9 +360,85 @@ export default class Controller {
...
 		logger.info("Goodbye");
 	}
 
+	onAutosaveIntervalChanged() {
+		if (this.autosaveInterval) {
+			clearInterval(this.autosaveInterval);
+			this.autosaveInterval = undefined;
+		}
+		const autosaveIntervalSeconds = this.config.get("controller.autosave_interval");
+		if (autosaveIntervalSeconds > 0) {
```
Cooldude2606, line 374: Is it documented that a negative or zero value results in no autosave?

packages/controller/src/Controller.ts
```diff
@@ -766,13 +841,15 @@ export default class Controller {
...
 
+			this.instancesDirty = true;
 			lib.invokeHook(this.plugins, "onInstanceConfigFieldChanged", instance, group, field, prev);
 		});
 
 		this.instanceUpdated(instance);
 	}
 
 	instanceUpdated(instance: InstanceInfo) {
+		this.instancesDirty = true;
```
Cooldude2606, line 852: `instanceUpdated` is called by `addInstanceHooks` which is called by `_startInternal` right after the config is loaded. Therefore, after every startup the instances will be considered dirty, is this the expected behaviour? Neither hosts nor modpacks have this behaviour.

Cooldude2606 (commented):
Additional points:
- A test case (or an issue) should be created to cover the case of three consecitive calls to `saveData` (this will cover all branches of the function at once)
- With the dirty bits implemented there is no longer a concern about data being saved after an error in `_startInternal` and so empty maps rather than null can be used for hosts, instances, and modpacks. (also for mods, although im not sure if this even gets saved)

packages/controller/src/Controller.ts
Hornwitser: No this is not expected behaviour. I looked into this and found that instanceUpdated should really be instanceDetailsUpdated, and did some refactoring to make instances not be marked as dirty on startup.

packages/controller/src/Controller.ts
psihius: Why even allow those values? If the value is out of bounds - use default or produce an error.

You know the Merphy law, right? :)

Hornwitser: Lots of code needed to treat it specially. Need to get this project done...

Cooldude2606 (approved):
LGTM. There are some cases of `let` being used over `const` but thats nit picks.

PR #528, opened 2023-11-29
Add the ctl counterpart to the revoke host action.

### LINKED ISSUE #512: Ability to revoke host token
In case of a rogue host we need the ability to disable access for it quickly without having to reissue all credentials. In short implement a tokenValidAfter property for hosts that the controller keeps track of and that can be set to the current time with a command/Web UI button and that will kill existing connections to that host.

Hornwitser:
Resolves #512.

### DIFF
--- a/test/integration/clusterio.js
+++ b/test/integration/clusterio.js
@@ -27,6 +27,34 @@ async function checkInstanceStatus(id, status) {
 	assert.equal(instances.get(44).status, status, "incorrect instance status");
 }
 
+async function startAltHost() {
+	let config = "alt-host-config.json";
+	let configPath = path.join("temp", "test", config);
+	await fs.remove(configPath);
+	await fs.remove(path.join("temp", "test", "alt-instances"));
+	await execCtl(`host create-config --id 5 --name alt-host --generate-token --output ${config}`);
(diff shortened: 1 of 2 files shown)


test/integration/clusterio.js
```diff
@@ -27,6 +27,34 @@ async function checkInstanceStatus(id, status) {
 	assert.equal(instances.get(44).status, status, "incorrect instance status");
 }
 
+async function startAltHost() {
+	let config = "alt-host-config.json";
+	let configPath = path.join("temp", "test", config);
+	await fs.remove(configPath);
+	await fs.remove(path.join("temp", "test", "alt-instances"));
+	await execCtl(`host create-config --id 5 --name alt-host --generate-token --output ${config}`);
```
Cooldude2606, line 35: Any reason why `5` was selected as the id? Would it be worth having it as an argument to this function?

Cooldude2606 (approved):
LGTM. Nothing blocking merge, just one question below.

test/integration/clusterio.js
Hornwitser: It used to be 5 before the code was moved to this function. 5 was selected because it needed a number and 5 sounded nice.

PR #529, opened 2023-11-29
Use .tmp before the file extension

### LINKED ISSUE #486: Fix .tmp files being selected for startup by default
When starting up an instance there may be a .zip.tmp file left over from a previous start attempt that failed. In the case this is the file with the latest timestamp it should not be the default choice for the save to load.

Hornwitser:
When creating temporary files use the pattern `.tmp.ext` instead of `.ext.tmp` to be consistent with Factorio which creates `.tmp.zip` files during saving.

If patching fails this will leave a `.tmp.zip` file in the instance saves directory which is ignored by the default save selector.

Resolves #486.

### DIFF
--- a/packages/host/src/patch.ts
+++ b/packages/host/src/patch.ts
@@ -417,7 +417,7 @@ export async function patch(savePath: string, modules: SaveModule[]) {
 	root.file("clusterio.json", JSON.stringify(patchInfo, null, "\t"));
 
 	// Write back the save
-	let tempSavePath = `${savePath}.tmp`;
+	let tempSavePath = savePath.replace(/(\.zip)?$/, ".tmp.zip");
 	let stream = zip.generateNodeStream({ compression: "DEFLATE" });
 	let fd = await fs.open(tempSavePath, "w");
 	try {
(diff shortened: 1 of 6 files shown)


packages/host/src/patch.ts
```diff
@@ -417,7 +417,7 @@ export async function patch(savePath: string, modules: SaveModule[]) {
 	root.file("clusterio.json", JSON.stringify(patchInfo, null, "\t"));
 
 	// Write back the save
-	let tempSavePath = `${savePath}.tmp`;
+	let tempSavePath = savePath.replace(/(\.zip)?$/, ".tmp.zip");
```
Cooldude2606, line 420: Is there a reason you have used `.replace` here when in the other two cases you have used `path.parse`?

Cooldude2606 (approved):
LGTM. Nothing blocking merge, just one question below.

packages/host/src/patch.ts
Hornwitser: It matches the way it's done for Factorio save paths in other places like Hosts.ts https://github.com/clusterio/clusterio/blob/faadf873fb98e070c0732634de4266d5ff90c055/packages/host/src/Host.ts#L754

Issue #513, opened 2023-09-28
server_select plugin deletes your entire character when used with Space Exploration's satellite view


Judisch:
If you use the server_select plugin while in Space Exploration's satellite view, your entire character gets deleted. It even deletes your personal logistics settings!

The satellite view separates your camera from the player entity, similar to the godmode cheat in vanilla. So it makes sense why this would happen. Maybe the plugin could double check that the player entity is attached?

I've already voided my character 3 times... it's very easy to forget you're in satellite view. 

Steps to reproduce:
1) Install space exploration
2) Launch some satellites to unlock the satellite view
3) Open the satellite view (hotkey N by default)
4) Use server select plugin to switch to another server
5) Switch back to original server
6) Your character is now deleted

psihius:
Thanks for the report.

This might require more profound integration work to make the whole thing work with Clusterio, probably as a dedicated plugin or something like that. Ideally, one or a few people from the community would step up and take charge for it - none of us who work on Clusterio on the regular play SE and the amount of work required is something we do not have time to do.

Anyone willing to track the details for this one and provide a possible fix?

Hornwitser:
The only thing server_select does is tell the client to disconnect from the current server and connect to a different server. This is more or less equivalent to using escape to quit the current server and then connect to another one in the server browser. For this reason it sounds exceedingly unlikely that this issue has anything to do with server_select.

Are you sure this is not caused by a different plugin like inventory sync?

Reavershark:
This is caused by inventory sync as it will upload your satellite view inventory on disconnect, which is just an artillery remote, some wire and maybe some blueprints.
SE has `remote_view_is_active(player)` and `remote_view_stop(player)` (remote means satellite here) in its remote interface.

Reavershark:
This patch fixes it for us:
```diff
diff --git a/plugins/inventory_sync/module/inventory_sync.lua b/plugins/inventory_sync/module/inventory_sync.lua
index 46fc354f..8997a99c 100644
--- a/plugins/inventory_sync/module/inventory_sync.lua
+++ b/plugins/inventory_sync/module/inventory_sync.lua
@@ -379,6 +379,13 @@ inventory_sync.events[defines.events.on_pre_player_left_game] = function(event)

        local player = game.get_player(event.player_index)
        local player_record = global.inventory_sync.players[player.name]
+
+       -- SE Only: Drop the player out of remote view so we can find its character
+       if remote.interfaces["space-exploration"] then
[...]

PR #530, opened 2023-11-29
Fix inventory wipe on SE remote view

Hornwitser:
Exit the Space Exploration remote view if its active to prevent the empty remote view "character" being serialized as the inventory instead of the real character for the player.

Original-patch-by: @Reavershark
Fixes: #513

### DIFF
--- a/plugins/inventory_sync/module/inventory_sync.lua
+++ b/plugins/inventory_sync/module/inventory_sync.lua
@@ -56,6 +56,19 @@ function inventory_sync.serialize_player(player, player_record)
 		return
 	end
 
+	-- mod-compat: Exit Space Exploration satelite view to restore the real character
+	if remote.interfaces["space-exploration"] then
+		local status, result = pcall(function()
+			if remote.call("space-exploration", "remote_view_is_active", { player = player }) then
+				remote.call("space-exploration", "remote_view_stop", { player = player })
+			end
+		end)
+		if not status then
+			log("ERROR: Exiting remote view failed for " .. player.name .. ": " .. result)
+			player.print("ERROR: Exiting remote view failed: " .. result)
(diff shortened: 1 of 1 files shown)


plugins/inventory_sync/module/inventory_sync.lua
```diff
@@ -56,6 +56,19 @@ function inventory_sync.serialize_player(player, player_record)
...
+	-- mod-compat: Exit Space Exploration satelite view to restore the real character
+	if remote.interfaces["space-exploration"] then
+		local status, result = pcall(function()
+			if remote.call("space-exploration", "remote_view_is_active", { player = player }) then
+				remote.call("space-exploration", "remote_view_stop", { player = player })
+			end
+		end)
+		if not status then
+			log("ERROR: Exiting remote view failed for " .. player.name .. ": " .. result)
+			player.print("ERROR: Exiting remote view failed: " .. result)
```
Cooldude2606, line 68: If this mode is similar to editor mode (as in it is not support due to the empty inventory) then should this not also call `error` as above?

Cooldude2606 (approved):
LGTM. Nothing blocking merge, just one question about a rare edge case.

plugins/inventory_sync/module/inventory_sync.lua
Hornwitser: Consider what would happen if the the remote view mode was removed entirely from Space Exploration, remote calls included. If this called `error` then instead of printing an error to the console you would no longer be able to use inventory sync at all. If it fails to switch the mode it should still try to sync it as the alternative is worse.

This is different from the editor mode where if it fails to toggle it off there's no way to access the character inventory, no way to sync the data.

PR #531, opened 2023-12-01
Refactor User and Role into data classes

### LINKED ISSUE #525: Move lib/PlayerStats to lib/data/PlayerStats

Hornwitser:
Use the JSON data class pattern for User, Role, and PlayerStats and move them into lib/src/data.  This means that User is purely data now and can't reference role objects directly.  To make it possible easily check permissions attached to roles and store data only visible on the controller on user objects a ControllerUser type is introduced which is linked with the UserManager and can therefore reference role objects directly.

Resolves #525.


packages/lib/src/data/User.ts
```diff
@@ -0,0 +1,153 @@
...
+
+		if (this.instanceStats.size) {
+			json.instance_stats = [...this.instanceStats].map(([id, stats]) => [id, stats.toJSON()]);
+		}
+
+		return json;
+	}
+
+	recalculatePlayerStats() {
+		this.playerStats = User._calculatePlayerStats(this.instanceStats);
```
Cooldude2606, line 114: This seams like an easy test which can be added to the `test/lib/users.js`

packages/controller/src/ControlConnection.ts
```diff
@@ -694,55 +678,29 @@ export default class ControlConnection extends BaseConnection {
 
 	async handleRoleDeleteRequest(request: lib.RoleDeleteRequest) {
 		let id = request.id;
-		let role = this._controller.userManager.roles.get(id);
-		if (!role) {
-			throw new lib.RequestError(`Role with ID ${id} does not exist`);
-		}
```
Cooldude2606, line 700: With this check being removed, `userPermissionsUpdated` will be called even if the user did not have the role. This could request in extra network traffic. `Set.delete` returns true if the value was in the set which can be used to reimplement this check.

packages/controller/src/ControlConnection.ts
```diff
@@ -694,55 +678,29 @@ export default class ControlConnection extends BaseConnection {
...
-				[...user.instances],
-				user.isAdmin,
-				user.isBanned,
-				user.isWhitelisted,
-				undefined,
-				undefined,
-				user.playerStats,
-				undefined,
-			));
+			list.push(user);
```
Cooldude2606, line 703: Could this instead be `return [...this._controller.userManager.users.values()]`

test/lib/users.js
```diff
@@ -68,56 +68,26 @@ describe("lib/users", function() {
...
-			user.notifyLeave(12);
-			assert(lib.User.onlineUsers.has(user));
-			assert.deepEqual(user.instances, new Set([8]));
-
-			user.notifyLeave(8);
-			assert(!lib.User.onlineUsers.has(user));
-			assert.deepEqual(user.instances, new Set());
+		it("should keep invalid roles", function() {
+			let user = lib.User.fromJSON({ name: "test", roles: [1, 4, 55] });
+			assert.equal(user.roleIds.size, 3, "Unexpected count of roles");
```
Cooldude2606, line 87: Why is this desired?

Cooldude2606 (changes requested):

test/lib/users.js
Hornwitser: Because user is a data class that is indpendent from role now so it shouldn't filter invalid roles and because it can't access the roles that exist it can't filter them either.

packages/controller/src/ControlConnection.ts
Hornwitser: Good idea. I added the check back for both for the role and for only updating modified users.

Cooldude2606 (approved):
All changes were accounted for, so LGTM.

PR #541, opened 2024-01-14
Real-time metrics

Hornwitser:
Implements basic real time collection of CPU, memory and disk usage on the systems that the controller and hosts runs on. This is not intended to replace a Prometheus, and Grafana setup, but rather provide the basic metrics to give an idea of what resources are available and being used at a glance while using the web interface. The stats collected are also exposed to Prometheus which means that for basic insights it's no longer necessary to set up a Node Exporter or similar tool to collect these.

In short this adds pretty indicator bars to the controller and hosts pages:

image

### DIFF
--- a/packages/controller/src/Controller.ts
+++ b/packages/controller/src/Controller.ts
@@ -53,6 +53,8 @@ export default class Controller {
 	config: lib.ControllerConfig;
 	app: Application;
 
+	/** True if (@link Controller.systemMetrics} has changed since last time it was saved */
+	systemMetricsDirty = false;
 	/** True if {@link Controller.hosts} has changed since last time it was saved */
 	hostsDirty = false;
 	/** True if {@link Controller.instances} has changed since last time it was saved */
@@ -89,6 +91,7 @@ export default class Controller {
 	devMiddleware: any | null = null;
 
 	autosaveInterval?: ReturnType<typeof setInterval>;
+	systemMetricsInterval?: ReturnType<typeof setInterval>;
 
 	logDirectory: string = "";
 	clusterLogIndex: lib.LogIndex | null = null;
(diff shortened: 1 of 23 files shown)


packages/controller/src/Controller.ts
```diff
@@ -452,6 +507,25 @@ export default class Controller {
...
+			json = JSON.parse(await fs.readFile(filePath, { encoding: "utf8" }));
+
+		} catch (err: any) {
+			if (err.code !== "ENOENT") {
+				throw err;
+			}
+			return new Map();
+		}
+
+		return new Map(json.map(m => lib.SystemMetrics.fromJSON(m)).map(m => [m.id, m]));
```
Cooldude2606, line 522: I find it odd that we are covering the `ENOENT` case but not the success case. Although I surpose this comes under integration tests which arent a priority.

packages/controller/src/routes.ts
```diff
@@ -39,7 +39,7 @@ function mergeSamples(destinationResult: lib.CollectorResultSerialized, sourceRe
 	}
 
 	for (let entry of receivedSamples) {
-		sourceResult.samples.push(entry);
+		destinationResult.samples.push(entry);
```
Cooldude2606, line 42: This appears to be a bug fix, a regression test would be beneficial here.

packages/lib/src/data/messages_controller.ts
```diff
@@ -204,6 +204,108 @@ export class LogMessageEvent {
 	}
 }
 
+export class SystemMetrics {
```
Cooldude2606, line 207: None of the getter are tested, this is a simple one to add as you only need to call `fromJson` and then check against known good values.

Cooldude2606 (approved):
Only comments are about test cases so nothing needs to be added in this PR.

packages/controller/src/routes.ts
Hornwitser: The commit changing the line describes it in detail. I don't think a regression test is worth the effort here because this line of code will likely never be hit.

PR #543, opened 2024-01-14
Explicit config creation

### LINKED ISSUE #548: Creating config files should be an explicit operation
Users get confused when running the Controller/Host from a different directory than where the config files are in and it creates a blank config that acts like a separate installation. Prevent this confusion by exiting with an error message if the config doesn't exist and provide a command/option to create the config.

Laar:
This changes the behaviour when starting a controller, host or ctl when the config file can not be found. The intention (according to the backlog) is to prevent accidentally starting in the wrong directory.

Specifically it adds the following
- Starting without a config gives an error (not starting)
- A generic `config create` command is added to create configs. This fails if the config file already exists.

### Thoughts
The create-config command creates a basic config for all three types (controller, host and ctl). There are already two commands to create configs,  `ctl create-host` for creating host configs and `controller bootstrap create-ctl-config` for ctl configs. This duplication seems rather suspect.

### Todo
- [x] Update ChangeLog
- [x] Update docs/contributing.md (checked, no references to creating configs, everything is done through the create script.)
- [x] Update README.md (checked, no references to creating configs. Everything is done through the create script.)
- [ ] Write explicit tests
- [x] Update create.js

### DIFF
--- a/packages/lib/src/shared_commands.ts
+++ b/packages/lib/src/shared_commands.ts
@@ -129,15 +129,17 @@ export function configCommand(yargs: any) {
  * Handle the actions that are made available by configCommand.
  *
  * @param args - yargs args object.
- * @param instance - Config instance.
+ * @param instanceLoader - Function that loads the config instance.
  * @param configPath - Path to configuration file.
  */
 export async function handleConfigCommand(
 	args: Record<string, unknown>,
-	instance: libConfig.Config<any>,
+	instanceLoader: () => Promise<libConfig.Config<any>>,
 	configPath: string
 ) {
 	let command = (args._ as string[])[1];
+
+	let instance = await instanceLoader();
 
 	if (command === "list") {
 		for (const name of Object.keys(instance.constructor.fieldDefinitions)) {
--- a/packages/lib/src/shared_commands.ts
+++ b/packages/lib/src/shared_commands.ts
@@ -3,6 +3,7 @@
  * @module lib/shared_commands
  */
 import path from "path";
+import fs from "fs-extra";
 
 import * as libConfig from "./config";
 import * as libFileOps from "./file_ops";
@@ -117,6 +118,7 @@ export function configCommand(yargs: any) {
 		})
 		.command("show <field>", "Show value of the given config field")
 		.command("list", "List all configuration fields and their values")
+		.command("create", "Create the default config")
 		.demandCommand(1, "You need to specify a command to run")
 		.help()
 		.strict()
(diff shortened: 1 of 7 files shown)


Hornwitser:
To provide some context about the other commands that make configs:
- `ctl create-host` creates a host config with `host.controller_url` set, and `host.name` and `host.id` provided on the command line, along with the option to generate a token for it. The idea behind it was that it would allow a cluster admin to generate a config and then send it to whoever was setting up the host.
- `bootstrap create-ctl-config` creates a ctl config with the url to the controller set and a token for the provided user. It's a shortcut to doing `clusterioctl control-config set control.controller_url https://example.com/` followed by `clusterioctl control-config set control.controller_token <long-string-of-characters>`

Both of these provide a shortcut to creating a config and filling in some values.

Code wise this looks fine, my only comment is that the three separate config loaders functions feels like they could be a static method on the `Config` class, though the TypeScript typings for that might not be that simple and straight forward to get working.

You will also need to update the `packages/create/create.js` script to handle creating the configs if they don't exist.

Cooldude2606:
I could not find any logic errors in the code you have written. There are a few bits of lint found by CI but other than that your changes make the repetition of the loader functions very clear. Perhaps it can be moved into a static as Hornwitser has stated, such as:

`lib.Config.load(filepath, location, raiseOnMissing)` 
followed by
`await lib.handleConfigCommand(args, lib.HostConfig.load, args.config)`

Laar:
I've checked the documentation. All the config creation goes through the create script (host, controller), or through the bootstrap command (ctl). The only exception is in the 'managing-a-cluster.md' where it uses the ctl to create a host.

packages/lib/src/shared_commands.ts
```diff
@@ -123,16 +126,29 @@ export function configCommand(yargs: any) {
...
+	emptyConfigSupplier: () => libConfig.Config<any>
 ) {
 	let command = (args._ as string[])[1];
 
+	if (command === "create") {
+		if (await fs.exists(configPath)) {
+			logger.error(`Config "${configPath}" already exists`);
+			return;
+		}
+		await libFileOps.safeOutputFile(configPath, JSON.stringify(emptyConfigSupplier(), null, "\t"));
```
Cooldude2606, line 147: Rather than using `fs.exists` you should use a try catch for `EEXIST`. Also as the happy path here is that the file does not exist, and because the config file is quite small, it is better to use `fs.outputFile` rather than `safeOutputFile` since the safe version is designed to overwrite existing files that can't risk being corrupted.

Cooldude2606 (commented):
Been a while since I last checked this PR so thought I give it a bump. All the logic looks correct and I didnt spot any flaws or refactors other than the one below. As soon as you have 3 or 4 tests written this is good to merge.

PR #545, opened 2024-01-15
Remote host config

Hornwitser:
Implement the ability to show and edit the host config remotely from the Web UI and `clusterioctl`.

### DIFF
--- a/packages/ctl/src/commands.ts
+++ b/packages/ctl/src/commands.ts
@@ -58,7 +58,7 @@ async function configToKeyVal(data: string) {
 
 function serializedConfigToString(
 	serializedConfig: lib.ConfigSchema,
-	configGroup: typeof lib.ControllerConfig | typeof lib.InstanceConfig,
+	configGroup: typeof lib.ControllerConfig | typeof lib.HostConfig | typeof lib.InstanceConfig,
 	disallowedList: Record<string, unknown>,
 ) {
 	let allConfigElements = "";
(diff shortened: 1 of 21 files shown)


packages/ctl/src/commands.ts
```diff
@@ -289,6 +289,149 @@ hostCommands.add(new lib.Command({
...
+		});
+		editorSpawn.on("data", (data) => {
+			process.stdout.pipe(data);
+		});
+		let doneEmitter = new events.EventEmitter();
+		editorSpawn.on("exit", async (exit) => {
+			const data = await fs.readFile(tmpFile, "utf8");
+			const final = await configToKeyVal(data);
+			for (let index in final) {
+				if (index in final) {
```
Cooldude2606, line 403: Bit of a nit pick, but these are keys not indexes. Comment also applies to use of `finalIndex` in `configToKeyVal`

packages/ctl/src/commands.ts
```diff
@@ -289,6 +289,149 @@ hostCommands.add(new lib.Command({
...
+		if (editor === undefined) {
+			throw new lib.CommandError(`No editor avalible. Checked CLI input, EDITOR and VISUAL env vars
+							  Try "ctl controller config edit <editor of choice>"`);
+		}
+		let disallowedList = {"host.id": 0};
+		let allConfigElements = serializedConfigToString(
+			config,
+			lib.HostConfig,
+			disallowedList
+		);
```
Cooldude2606, line 389: Currently keys in the disallowed list are silently ignored with a `continue` it would be better UX to have a warning emited to avoid cases where the user doesn't understand why it is not applying the changes.

Cooldude2606 (approved):
Might want to add the warning to `host config edit` otherwise nothing is blocking a merge.

packages/ctl/src/commands.ts
Hornwitser: This was copied from the controller and instance config commands, so if anything is to be done it should be done on those as well. Which is a cleanup I considered to be out of scope for this PR.

packages/ctl/src/commands.ts
Hornwitser: There are many issues I'd like to fix, but right now my priority is getting Alpha 14 released so this is something that can wait. Feel free to open an issue about it (or add it to the backlog if you have access there).

PR #546, opened 2024-01-16
Use X-Forwarded-For header on trusted proxies

Hornwitser:
If the controller is behind a proxy then the socket.remoteAddress will be that of the proxy instead of the end client that connected.  In this case the X-Forwarded-For header will contain the address of the end client, but this should only be relied upon if the party connecting is a trusted proxy, otherwise a malicious user could spoof the IP address that is logged by providing a forged X-Forwarded-For header.

Implement a trusted proxy list consisting of a configured list of IP addresses and/or CIDR blocks and use the last address<sup>[1]</sup> in the X-Forwarded-For header if the remoteAddress address of the connecting party is in the trusted proxy list.

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

### DIFF
--- a/packages/controller/src/Controller.ts
+++ b/packages/controller/src/Controller.ts
@@ -1,5 +1,5 @@
 import type winston from "winston";
-import type { AddressInfo } from "net";
+import { BlockList, type AddressInfo, isIPv4, isIPv6 } from "net";
 import type { ControllerArgs } from "../controller";
 import express, { type Request, type Response, type NextFunction, type Application } from "express";
 import type { Static } from "@sinclair/typebox";
@@ -74,6 +74,7 @@ export default class Controller {
 
 	/** WebSocket server */
 	wsServer: WsServer;
+	trustedProxies: BlockList;
 	debugEvents: events.EventEmitter = new events.EventEmitter();
 	private _events: events.EventEmitter = new events.EventEmitter();
 
@@ -153,6 +154,7 @@ export default class Controller {
 		this.app.locals.controller = this;
 		this.app.locals.streams = new Map();
 
+		this.trustedProxies = this.parseTrustedProxies();
 		this.wsServer = new WsServer(this);
 
 		this.modStore.on("change", mod => {
@@ -228,6 +230,8 @@ export default class Controller {
 				this.onAutosaveIntervalChanged();
 			} else if (field === "controller.system_metrics_interval") {
 				this.onSystemMetricsIntervalChanged();
+			} else if (field === "controller.trusted_proxies") {
+				this.trustedProxies = this.parseTrustedProxies();
 			}
 			lib.invokeHook(this.plugins, "onControllerConfigFieldChanged", field, curr, prev);
 		});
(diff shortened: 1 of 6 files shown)


packages/controller/src/Controller.ts
```diff
@@ -404,6 +408,34 @@ export default class Controller {
 		logger.info("Goodbye");
 	}
 
+	parseTrustedProxies() {
+		const trustedProxies = new BlockList();
+		const proxiesString = this.config.get("controller.trusted_proxies");
+		if (proxiesString) {
+			const proxies = proxiesString.split(",").map(s => s.trim());
```
Cooldude2606, line 415: Thoughts on allowing the list to be split by spaces and commas, same as how they are allowed for port ranges?
Could use `.split(/[, ]/)` then filtering out empty strings after trim.

packages/controller/src/WsServer.ts
```diff
@@ -59,6 +59,20 @@ export default class WsServer {
...
+	remoteAddr(req: IncomingMessage) {
+		const remote = req.socket.remoteAddress ?? "";
+		if (
+			remote && req.headers["x-forwarded-for"]
+			&& (
+				(req.socket.remoteFamily === "IPv4" && this.controller.trustedProxies.check(remote, "ipv4"))
+				|| (req.socket.remoteFamily === "IPv6" && this.controller.trustedProxies.check(remote, "ipv6"))
+			)
+		) {
+			return (req.headers["x-forwarded-for"] as string).split(",").slice(-1)[0].trim();
```
Cooldude2606, line 71: Thoughts on using a while loop to get the rightmost address that is not trusted?

Cooldude2606 (approved):
LGTM. No changes required, just two points of consideration / future changes.

packages/controller/src/WsServer.ts
Hornwitser: For what gain? Performance? Clarity?

packages/controller/src/Controller.ts
Hornwitser: The port ranges allowing space delimited values is more of a byproduct of making the parsing logic simpler to implement. IMHO it would be better to rework the port ranges parsing to be more strict.

packages/controller/src/WsServer.ts
Cooldude2606: The gain would be getting the first address that you don't consider a trusted proxy, which is hopefully the client. This would allow for multiple proxies to be used. Taken from [Selecting an IP address](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#selecting_an_ip_address).

Hornwitser: The code already fetches the rightmost address.
```js
> "forged, 10.0.0.1".split(",").slice(-1)[0].trim()
'10.0.0.1'
```
Here `slice(-1)` returns a slice of the array starting from the last element (negative numbers index from the end.

Hornwitser: To put it differently, I don't think anyone are operating their controller behind a chain of proxies so I will put off implementing the logic to support that until someone actually show up needing it. I believe this practise is called lazy coding.

PR #552, opened 2024-01-17
Additional system info

Hornwitser:
Adds additional system information about hosts and the controller, specifically the hostname, Node.js version, kernel type, machine instruction set type, and CPU model, along with the address a host connects from.

These system metrics are displayed on the host and controller pages, looking something like this:
image

To simplify the distribution of system information I decided to repurpose the existing SystemMetrics subscription into a more generic SystemInfo subscription that includes the metrics along with the additional information added in this PR.


test/lib/data/SystemInfo.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
+const assert = require("assert").strict;
+
+const lib = require("@clusterio/lib");
+const { SystemInfo } = lib;
+
+
+describe("lib/data/ModPack", function() {
```
Cooldude2606, line 8:
```suggestion
describe("lib/data/SystemInfo", function() {
```

Cooldude2606 (changes requested):

Cooldude2606 (approved):

PR #553, opened 2024-01-17
Use a relative path for commands in create

Laar:
The create.js script in development mode used to use hard coded paths to the other cluster commands (ctl, controller etc.). This means that in dev mode a cluster could only be created from the root directory of the project.

By using relative paths one can also create a cluster in a subdirectory. For example by using the following:

mkdir test-cluster
cd test-cluster
node ../packages/create --dev

This is for example useful for testing the create.js in a separate directory.

### DIFF
--- a/packages/create/create.js
+++ b/packages/create/create.js
@@ -113,21 +113,21 @@ async function execFile(cmd, args) {
 
 async function execController(args) {
 	if (dev) {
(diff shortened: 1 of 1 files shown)


Cooldude2606 (approved):
LGTM. I've checked a sample of other `path.join` calls and I believe these are the only three which needed to be changed.

Hornwitser (approved):
This is an excellent idea.

PR #554, opened 2024-01-17
Implement stop and restart of controller/hosts

### LINKED ISSUE #549: Command to restart Controller/Host
When running from systemd, or bash/cmd script it should be possible to pass a parameter saying restart if exit code is a special value, and then program the systemd unit/script to start the Controller/Host again if this exit value is received. Then a UI/ctl command to restart the Controller/Host(s) can be added.

Hornwitser:
Implement remotely initiated stopping and restarting of the controller and hosts.  Restarting is implemented by exiting with a non-zero exit code and relying on a process manager to start the procress again.  To prevent accidentally trying to restart on a system that doesn't have a process manager present the `--can-restart` command line option needs to be passed to enable the restart operation.

Resolves #549

### DIFF
--- a/packages/controller/controller.ts
+++ b/packages/controller/controller.ts
@@ -244,6 +244,10 @@ async function initialize(): Promise<InitializeParameters> {
 				.demandCommand(1, "You need to specify a command to run");
 		})
 		.command("run", "Run controller", yargs => {
+			yargs.option("can-restart", {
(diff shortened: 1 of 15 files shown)


Cooldude2606 (approved):
Great stuff, I had a comment ready about "stop" being dangerous and needing a confirmation but you beat me to it and included it already. So no changes needed.

PR #556, opened 2024-01-18
Add logging for express requests and errors

Cooldude2606:
Adds logging of all requests sent to express under a new "http" log level (lower than info, higher than server and verbose). Also includes logging of errors from routes and middleware such as `express.json` rather than using the default `console.error`.


packages/controller/src/Controller.ts
```diff
@@ -165,6 +166,31 @@ export default class Controller {
...
+						httpVersion: req.httpVersion,
+						headers: req.headers,
+						query: req.query,
+					},
+				});
+			});
+
+			next();
+		});
+
```
Hornwitser, line 193: There's already a middleware that does the same thing as this one in order to populate endpointDurationSummary further down in this file. These should be merged into one.

Hornwitser (changes requested):
The code logging requests does almost exactly the same thing as the code updating the `endpointDurationSummary`. Other than than that this looks good.

packages/controller/src/Controller.ts
Cooldude2606: Merged in commit [0b2fcdc](https://github.com/clusterio/clusterio/pull/556/commits/0b2fcdcec471a5ba029a0f2b017ba8e35d44174b)

Hornwitser (approved):

PR #558, opened 2024-01-19
Document migration of plugins to alpha 14

Hornwitser:
These are the changes I found I needed to do to port the fdgl_integration, server_select, and discord_bridge plugins to alpha 14. There may be things I'm missing, though this should be good for now.

### DIFF
--- a/docs/devs/plugin-migration-to-alpha-14.md
+++ b/docs/devs/plugin-migration-to-alpha-14.md
@@ -28,7 +28,7 @@ const libLuaTools = require("@clusterio/lib/lua_tools");
 const libErrors = require("@clusterio/lib/errors");
 
 // Using imports
-new libLink.RequestError(...)
+new libErrors.RequestError(...)
 libLuaTools.escapeString(...)
 ```
 
@@ -80,7 +80,7 @@ Previously the info of a plugin would be loaded by requiring `./plugin_dir/info`
 To solve this you can either add `"main": "info.js",` to package.json or rename `info.js` to `index.js` (the rest of this guide assumes the entrypoint was renamed to `index.js`).
 The motivation behind this change was to allow TypeScript to output its build to a sub-folder of the plugin, if you're using TypeScript you'll likely have the entrypoint be located in `dist/index.js` instead.
 
-The expected place the declaration of the plugin is exported also changed from `module.exports` to `module.exports.plugin`:
+The expected place where the declaration of a plugin is exported has also changed from `module.exports` to `module.exports.plugin`:
 
 **Alpha 13**
 ```js
@@ -181,7 +181,7 @@ const plugin = {
             title: "Option",
             description: "A config option.",
             type: "string",
-            initial_value: "string",
+            initial_value: "foo",
             optional: true,
         },
     },
@@ -211,7 +211,7 @@ export const plugin: lib.PluginDeclaration = {
             title: "Option",
             description: "A config option.",
             type: "string",
-            initialValue: "string",
+            initialValue: "foo",
             optional: true,
         },
     },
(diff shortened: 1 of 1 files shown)


docs/devs/plugin-migration-to-alpha-14.md
```diff
@@ -0,0 +1,523 @@
...
+Previously code could import sub packages of lib such as lib/lua_tools.
+These imports need to be changed to all import from lib directly.
+
+**Alpha 13**
+```js
+const libLuaTools = require("@clusterio/lib/lua_tools");
+const libErrors = require("@clusterio/lib/errors");
+
+// Using imports
+new libLink.RequestError(...)
```
Cooldude2606, line 31: Should be `libErrors` because you don't show `libLink` being required.

docs/devs/plugin-migration-to-alpha-14.md
```diff
@@ -0,0 +1,523 @@
...
+
+// Using library components
+new lib.RequestError(...)
+lib.escapeString(...)
+```
+
+
+## Change import of base plugins
+
+Previously `BaseMasterPlugin` and `BaseInstancePlugin` were imported from `@clusterio/lib`, these have been moved to `@clusterio/controller` and `@clusterio/host` respectively.
```
Cooldude2606, line 56: Has `BaseHostPlugin` moved as well?

docs/devs/plugin-migration-to-alpha-14.md
```diff
@@ -0,0 +1,523 @@
...
+```
+
+
+## Change main entrypoint of plugin
+
+Previously the info of a plugin would be loaded by requiring `./plugin_dir/info`, this has been changed to require `./plugin_dir` which in turn means that `info.js` is no longer the default entrypoint.
+To solve this you can either add `"main": "info.js",` to package.json or rename `info.js` to `index.js` (the rest of this guide assumes the entrypoint was renamed to `index.js`).
+The motivation behind this change was to allow TypeScript to output its build to a sub-folder of the plugin, if you're using TypeScript you'll likely have the entrypoint be located in `dist/index.js` instead.
+
+The expected place the declaration of the plugin is exported also changed from `module.exports` to `module.exports.plugin`:
```
Cooldude2606, line 83: I shouldn't be one to talk about grammar, because mine is arwful, but...
```suggestion
The expected place where the declaration of a plugin is exported has also changed from `module.exports` to `module.exports.plugin`:
```

docs/devs/plugin-migration-to-alpha-14.md
```diff
@@ -0,0 +1,523 @@
...
+**Alpha 14 (JS)**
+```js
+const plugin = {
+    name: "plugin_name",
+    controllerConfigFields: {
+        "plugin_name.option": {
+            title: "Option",
+            description: "A config option.",
+            type: "string",
+            initial_value: "string",
```
Cooldude2606, line 184: Using `"string"` here might have people assume it is the type of the inital value. I recomend something like "foo" or "initial option" (same applies to ts version)

docs/devs/plugin-migration-to-alpha-14.md
```diff
@@ -0,0 +1,523 @@
...
+```js
+async onControllerConfigFieldChanged(field, curr, prev) {
+    if (field === "plugin_name.option") {
+        // Do stuff with curr and or prev
+    }
+}
+```
+
+**Alpha 14 (TS)**
+```js
```
Cooldude2606, line 245:
```suggestion
```ts
```

docs/devs/plugin-migration-to-alpha-14.md
Hornwitser: No, it was added in alpha 14.

Cooldude2606 (approved):

PR #561, opened 2024-01-20
Add time unit suffixes to variables

Hornwitser:
Add suffixes like Ms and Ns to variables whose time unit is not in seconds.  This is aligning with a new code style of putting units on variables that are not in SI base units.


Cooldude2606 (approved):
I have not checked for missed renames, but I have checked that all that were made are correct.

Issue #565, opened 2024-01-21
Feature Request: Versions in instance list


n0t49a1n:
to help with version control and being able to see the cluster as a whole
Would like to see the Version of the instance in the instance tab

below is a mock up of what i need
![image](https://github.com/clusterio/clusterio/assets/87333/9f2c4a48-e9e4-4923-a37a-aab33b04ebe0)

n0t49a1n:
id love to try this but i have no idea how this web works at all

CCpersonguy:
I'll make an attempt at this. The UI change seems straightforward, I'm reading code to figure out which events(s) should be used to send this info to the Controller.

Cooldude2606:
From a quick look, my approach would be expanding `InstanceDetails` to include the factorio version, then within `InstanceInfo::toInstanceDetails` you have access to the instance config where the game version is stored.

PR #573, opened 2024-01-23
Display Instance's Factorio version in the web UI

CCpersonguy:
Display each Instance's Factorio version on the "Instances" page in the web UI for #565.

Instances decide which Factorio version to use during startup (the configured version might be "latest", or a specific version might not be installed), so all Instances are "unknown" when the Host first starts. Instances that are running or which were previously running on their current Host, display the current or last used Factorio version respectively.

Once an Instance starts, it includes the version in its InstanceStatusChangedEvents. Hosts also send their Instances' versions as part of HostInstanceUpdate, in case the Host is reconnecting to the Controller. (A comment notes that InstanceStatusChangedEvents are also sent when a Host reconnects, but that doing so is a hack. HostInstanceUpdate seems slightly more "correct").

Alternatives:

- The Host has access to the Instance config and the same filesystem, so in theory it could determine which version will be used, but I think it's better to keep Instance-config-parsing as the Instance's responsibility.
- Web UI could display the _configured_ version instead, which would remove several of the message changes, but that seems less useful. Two instances could both be configured to use "latest", but if they're on different hosts they could end up using different versions.

TODO:

- [ ] Add tests
- [ ] Verify that message parsing/passing works with mis-matched Controller and Host versions (forwards + backwards compatible). Should be fine since the new field is Optional, but should double-check.


Hornwitser:
This is looking pretty good.

I have not reviewed the code in detail so my only comment on it right now is that I would like to see the version running being displayed as part of the details on `web_ui/src/components/InstanceViewPage.tsx` as well.

> The Host has access to the Instance config and the same filesystem, so in theory it could determine which version will be used, but I think it's better to keep Instance-config-parsing as the Instance's responsibility.

It's possible to add new versions of Factorio to the factorio directory while the host is running so implementing this correctly would be involved. Showing the version that's running is the most useful and simple to implement right now.

> Web UI could display the configured version instead, which would remove several of the message changes, but that seems less useful. Two instances could both be configured to use "latest", but if they're on different hosts they could end up using different versions.

This is even less useful than that. If you change the configured version of a running instance the configured version changes but the running version does not, so that would make it miss leading.

Hornwitser:
@CCpersonguy do you have an update on this? Would be nice to get this feature in.

Danielv123:
image

Seems to work well enough for me. I added the factorio version to the instanceViewPage as well. Only thing I am not quite sure about is the status of the version after the instance is stopped. Currently it shows as unknown until the instance is started, at which point it shows the current version even after it is stopped.

After the instance has stopped the version might not be accurate anymore.

I still think this is more useful than showing "unknown", but maybe showing the configured version instead with a small icon to note that its the configured version would be more useful?

Ex currently a stopped instance will show "unknown". You won't know what version it is before starting it. A more useful behaviour would be to show the config value of "latest (configured)", which at least gives some indication as to what will happen when starting it.

packages/web_ui/src/components/InstanceList.tsx
```diff
@@ -86,6 +93,13 @@ export default function InstanceList(props: InstanceListProps) {
 			sorter: (a, b) => strcmp(instancePublicAddress(a), instancePublicAddress(b)),
 			responsive: ["lg"],
 		},
+		{
+			title: "Factorio Version",
+			key: "version",
+			render: (_, instance) => instance.factorioVersion ?? "unknown",
```
Danielv123, line 99: Here we show the current version read from the server or "unknown". Should we show the configured value (usually "latest") with some icon indicating that it is the configured rather than actual value instead?

packages/web_ui/src/components/InstanceViewPage.tsx
```diff
@@ -59,6 +59,9 @@ function InstanceDescription(props: InstanceDescriptionProps) {
 				buttonContent={assigned ? "Reassign" : "Assign"}
 			/>}
 		</Descriptions.Item>
+		<Descriptions.Item label="Version">
+			{instance.factorioVersion ?? "unknown"}
```
Danielv123, line 63: Same here

Hornwitser:
I think showing the configured version when it's not running is the simplest and most useful thing to do right now. That means if the configured version is "latest" and it's currently running on "1.1.100" then when you stop it it should go from "1.1.100" to "latest". It would be best if this logic is not implemented in the Web UI, as that means the ctl interface would also show it correctly.

In the future I want the version to run to be optionally controlled by the mod pack, and have the host obtain the Factorio install automatically if it does not have it.

Danielv123:
I don't particularly like that idea. Doing that would obfuscate the source of the information provided, as well as just duplicating a value already accessible and transmitted over the wire. Are you sure its not a better idea to do this on the display side? That would also enable us to better customize the display of the value and indicate its source.

Supplying factorio version from the modpack is definitely sensible, as is auto downloading factorio if the correct version is not available, although I'd rather see that implemented as a function for modpacks being able to override settings in general, not factorio versions specifically.

I can see modpacks being interested in overriding the following:
- Whitelist - Might be using a mod to handle permission management with ingame permissions
- Version - for compatibility reasons
- Various plugin settings

I guess looking at it, the utility of anything but version is dubious. Overall I think it would be very low priority no matter how its implemented, with automatic download of the latest factorio version being far more useful.

Hornwitser:
> I don't particularly like that idea. Doing that would obfuscate the source of the information provided, as well as just duplicating a value already accessible and transmitted over the wire. Are you sure its not a better idea to do this on the display side? That would also enable us to better customize the display of the value and indicate its source.

If that's the case then it should be called `runningFactorioVersion` and only have a value when the instance is running.

Hornwitser:
Ask yourself this; what do you expect to happen when you start the following instance?
image
(Or alternatively, sourced from ctl)
image
(Or sourced from the API with InstanceDetailsListRequest)

a. It will always start using Factorio version 1.1.101
b. It may start using Factorio version 1.1.101, but it could also start using any other version.

The current code in this PR does the latter, and it sounds like you want it to be this way.

In the future, if the version becomes part of the mod pack and the factorioVersion in InstanceDetails does not show the configured version when stopped then the users of the API have to do the following logic in order to answer the question "which version of Factorio does this stopped instance start with?"
```
version = instance.factorioVersion
if instance.status == "stopped" then
    config = fetch instance config
    version = config.factorio.version
    if version is null then
        mod pack = fetch mod pack config.factorio.mod_pack_id or the default mod pack if that is null
        version = mod pack.factorioVersion
    endif
endif
```


This is not nice for the user. And I fail to see how it's more clear or maintainable as we're effectively reimplementing the version selection logic of the server in the client in order to display what the server will do. This sort of duplicate work easily breaks as there's nothing linking the two pieces of logic.

There's also an access problem with consulting the config for the version an instance will start with. The instance config currently contains credentials which means it might not be granted to everyone who can see the instances and start or stop them. In that case do you just not get to see which version of Factorio an instance will start with?

Danielv123:
After implementing it and playing around with it I agree - your way definitely made more sense.

psihius (approved):

psihius (approved):

Hornwitser (approved):
Looks good to me. Thank you for your contribution @CCpersonguy and apologies for the long wait. I have been on a hiatus for a while now and have just about got back into working on Clusterio.

PR #576, opened 2024-01-24
Fix plugin broadcasts not being filtered

Hornwitser:
When a plugin broadcasts an event to instances there may be hosts and instances in the cluster which does not have the plugin loaded.  This would cause an error to be logged for every instance which does not have the plugin loaded.

Filter out broadcast events sent to hosts and instances such that instances and hosts that does not have the event do not received them.


Cooldude2606 (approved):
Fixes are implemented and the tests are good. Only issue is the `send` and `bcast` functions within the tests. But after discussion, I agree that removing those would be a large undertaking that should be left for another time.

PR #577, opened 2024-01-24
Small things 15

Hornwitser:
These are the last small changes I have in mind before alpha 15.


Cooldude2606 (approved):

Cooldude2606:
Maybe the string format of Address can be changed to match the input of shorthand in some way so that it can be tested. But this is out of scope for this PR.

PR #582, opened 2024-02-05
Fix unhandled rejection on disconnect

Hornwitser:
If a link is closed while the prepareDisconnect action is in progress the link will still try to signal the disconnect ready over the socket when the prepareDisconnect action completes, which throws due to the socket no longer being present.  This error is not caught due to the catch happening before the finally on the promise and triggers an unhandled rejection.

Fix by rearranging the finally to come before the catch on the promise so that errors are always handled, and checking that the link still has a session before attempting to send the disconnct ready signal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Changes
 
 - Fixed an unhandled error in Subspace Storage crashing the host.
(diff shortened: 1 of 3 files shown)


Cooldude2606 (approved):
Simple changes, no change in happy path logic.

PR #585, opened 2024-02-08
Update message section of writing-plugins document

Danielv123:
While migrating my plugins I noticed this section wasn't updated, so I gave it a go. Alternatively, it could be removed and replaced with a link to defining-link-messages, but it feels like it fits in where it is.


Hornwitser (approved):
Looks good. I have some suggestions for changes, like specifying which properties are static, and rewording of some sections. But these don't have to be changed.

PR #587, opened 2024-02-09
Make resource references clickable

### LINKED ISSUE #572: Make resource references clickable
tl;dr Change resources references labels into clickable elements that navigate you to the page for that resource. (Example resources: host, instance, user, role, etc)

### Motivation
There are many places where the name/id of a resource is displayed but in order to view that resource you must navigate to the list of all resources of that type and then click the one you want. This is slow and unintuitive and might only get worse as more plugins are developed.

### Solution
Anywhere that a resource reference is displayed on the web ui should be made clickable with a href or otherwise. It should be obvious to the user that the elements are clickable but are not buttons. This could mean the text is underlined and is highlighted on hover, or something like a chip style that google uses, or a translucent box with a chain icon as used by discord.
```[tasklist]
### Tasks
- [x] Host assign box on instance page
- [x] Host in instance list
- [x] User role tags in users table -> Role settings page
```

Danielv123:
Resolves #572 unless there are any more references you'd expect to be clickable I haven't found.

This makes the following clickable:
* Host in instanceList now redirects to host page instead of instance page
* Host on instance page now redirects to the host page
* Clicking the role tag in the users list redirects to the role page

I am not entirely sure about the color of the link. Sure, its more visible that its a link and expected to do something different when its blue, but its also a bit less visible.

### DIFF
--- a/packages/web_ui/src/components/InstanceList.tsx
+++ b/packages/web_ui/src/components/InstanceList.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Link, useNavigate } from "react-router-dom";
+import { useNavigate } from "react-router-dom";
 import { message, Button, Space, Table, Typography } from "antd";
 import CopyOutlined from "@ant-design/icons/CopyOutlined";
 import type { SizeType } from "antd/es/config-provider/SizeContext";
@@ -10,6 +10,7 @@ import { useHosts } from "../model/host";
 import InstanceStatusTag from "./InstanceStatusTag";
 import StartStopInstanceButton from "./StartStopInstanceButton";
 import { InstanceDetails } from "@clusterio/lib";
+import Link from "./Link";
 
 const strcmp = new Intl.Collator(undefined, { numeric: true, sensitivity: "base" }).compare;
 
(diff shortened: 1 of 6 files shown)


Hornwitser:
This is creating nested `<a>` tags in the HTML.
image

Danielv123:
*sigh*Should I use the react-router-dom link component, react-router-dom link component with antd styling or antd link component?React-router-dom link does not take a component prop like shown in your example.On 9 Feb 2024 07:43, Hornwitser ***@***.***> wrote:
This is creating nested <a> tags in the HTML.
image.png (view on web)

Hornwitser:
The component prop was removed in v6 of react-router-dom, so it looks like the solution is to make our own Link component like describe at https://reactrouter.com/en/main/upgrading/v5#remove-link-component-prop

Danielv123:
Hi, I updated the PR according to the migration document. Got a bit confused with typescript and that onClick handler, but the linter doesn't complain and the duplicated <a> tags are gone.

Hornwitser:
This looks good. Only missing an export in `packages/web_ui/src/index.js` so that plugins can also make use of the `Link` component.

PR #592, opened 2024-02-12
Add pagination to saves list

Hornwitser:
Prevent the saves list from pushing the console and config off the screen by turning on the pagination for it.


Cooldude2606 (approved):
Table renders correctly so LGTM.

PR #600, opened 2024-02-25
Increase libLink event emitter leak warning limit

Danielv123:
The browser console has a warning about 11 event emitters possibly being an event listener leak. It is not.

I suggest increasing the limit to 20. We could also set it to 0 to disable the warning entirely, but I suppose it could be nice to be warned if we actually have a leak.


Hornwitser (approved):

Hornwitser:
20 should be enough for now.

PR #602, opened 2024-02-27
React to drag and drop

### LINKED ISSUE #591: React to drag and drop
Currently you can upload a save or mod by drag and drop onto the "Upload" button, the saves list also support(ed) drag and drop onto it. But there's no way for a user to know this and the interface doesn't react when you try dragging over it.

It would be preferable to have a component that reacts to the drag event by rendering a a large drop zone over the table that shows the saves/mods you can upload to so that this feature is discoverable.

Danielv123:
Added drag and drop overlay to upload button on mods page and save table on instance page.

Also added drag and drop uploading to the mods table for good measure.

Resolves #591

### DIFF
--- a/packages/web_ui/src/components/Dropzone.tsx
+++ b/packages/web_ui/src/components/Dropzone.tsx
@@ -3,11 +3,19 @@ import { DraggingContext } from "../model/is_dragging";
 import { useContext } from "react";
 
 // This component is used to display a dropzone when a file is being dragged over the parent component
-export function Dropzone() {
+export function Dropzone({ disabled = false }: { disabled?: boolean }) {
 	const isDroppingFile = useContext(DraggingContext);
 
+	const textStyle: { color?: string } = {};
+	let text = "Drop to upload";
+	if (disabled) {
+		textStyle.color = "red";
+		text = "Target is offline";
+	}
+	const borderColor = disabled ? "gray" : "rgb(22, 119, 255)";
+
 	return <div
-		className="dropzone" // Don't remove this class, linked to SiteLayout.tsx
+		className={`dropzone ${disabled ? "disabled" : "enabled"}`} // Don't remove this class, linked to SiteLayout
 		style={{
 			position: "absolute",
 			top: "0",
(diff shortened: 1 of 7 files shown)


Cooldude2606 (commented):
When a host is offline, draging a save file for upload will instead prompt the user to download the zip file. (tested on firefox)

Cooldude2606:
The box and button outline are displayed correctly. However, the default behaviour of prompting a download is not prevented when the target is missed or the host is offline. Also, out of scope for this PR, but we need to add error messages when upload fails because of falid validation or duplicate item.

Danielv123:
This is a browser default - I am not sure if we want to override that?

As for host offline, I propose making the icon red or something to show the field not currently being an available target but still showing that it is a drop target

I was fairly sure I had seen error messages for file uploads in the past? Might have been in the console or something.

Cooldude2606:
While it is the default, it is also a bad user experince. Here is what it looks like on Google Drive when you drag a file into the wrong area. One Drive also has the same behaviour.

google drive

I agree that there should be something to show that upload is possible but not at the current time.

I have checked, there is an error message in the console. A 500 response from the server with a message explaining the reason. We shouldn't consider this to be user facing so is still something for us to work on later.

Danielv123:
If microsoft doesn't care about the defaults than I don't either. I am using the default "copy"/"none" icons from the browser - on macos for some reason the "move" and "none" looks the same, not sure how it looks to you?

Cooldude2606:
I'm on windows 10 firefox and the copy and none works fine for me. Another issue has arose tho: when the instance state is unknown you can drag onto the area, this causes default behaviour to occur (since we arent stopping the interaction) but the strange part is that it becomes impossible to interact with it a second time until you refresh the page.
img
It could be argued that the correct behaviour is that the first interaction should also be blocked, so this might be an issue that the state is not being set on refresh.

Cooldude2606 (approved):
LGTM. Everything works as expected now. (Firefox, Windows 10)
The arguments for the dropzone component are styled differently to others, but I'm not an expert on react so can't comment on it. Either way it works so I have no issue there.

PR #605, opened 2024-03-05
Fix possible undefined title error

Cooldude2606:
When building a clean dev install, including during CI, there is the following error:

`plugins/global_chat prepare: ../../packages/web_ui/src/components/SavesList.tsx(153,9): error TS18048: 'option.title' is possibly 'undefined'.`

There is a race condition for which plugin will error first but all of them can throw this error.
I do not know what has changed in the last week to cause this error to appear but it is preventing all CI pipelines.

The fix is an assertion that the value is defined, it is not clear why an `any` type requires this assertion but it works.

### DIFF
--- a/packages/web_ui/src/components/SavesList.tsx
+++ b/packages/web_ui/src/components/SavesList.tsx
@@ -150,7 +150,7 @@ function TransferModal(props: ModalProps) {
 						autoFocus
 						showSearch
 						filterOption={(input, option) => (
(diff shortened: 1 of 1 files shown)


Hornwitser:
Looks like this happens because https://github.com/react-component/select just released v14.12.0 which changed the type of `title` from any to an optional string.

Hornwitser (approved):

PR #606, opened 2024-03-06
Cache ModInfo to reduce startup time

Hornwitser:
Store and use a cache of the loaded ModInfo data to mod-info-cache.json in the mods directory during startup.  For simplicity this does not cache mods added to the controller or host after startup, these get added to the cache on the next startup.

_I've been trying to do reviews, but my local dev environment has become really slow to start up due to all the mods I have._

### DIFF
--- a/packages/lib/src/ModStore.ts
+++ b/packages/lib/src/ModStore.ts
@@ -1,8 +1,10 @@
 import fs from "fs-extra";
 import path from "path";
+import { Static } from "@sinclair/typebox";
 import { logger } from "./logging";
 import { ModInfo } from "./data";
 import TypedEventEmitter from "./TypedEventEmitter";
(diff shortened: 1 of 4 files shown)


Cooldude2606 (commented):
Don't see any errors. Would be nice to have an info log saying how many mods were loaded from the cache because currently it may give the impression that no mods where loaded. I see two possible outputs that could be used:
- `Loaded the info for {X} mods from the cache` (output after cache read, less accurate because cached mods may have been removed)
- `Loaded the info for {X} mods` (output on cache write, more accurate and includes mods not in the cache, might confuse people unaware of the cache behaviour)

Also considered `Loaded {X} mods, {Y} new {Z} cached` but that is more complex and doesn't provide much more value, so isn't worth the complexity imo.

Cooldude2606 (approved):

PR #607, opened 2024-03-09
Fix plugin hooks called before host removal

Hornwitser:
Due a bad order of operations when a host connection was closed the event handler listening on "close" to remove the host from the hostConnections mapping was added after the HostConnection's event handler listening on "close" and invoking plugin hooks in response to the host going offline.  This in turn causes plugins that send broadcasts in repsonse to thos hooks to throw an error due to the broadcast sending code assuming connections that exist are valid target to send messages to.

Fix by ordering the listener for the removal of the host connection before the listener invoking plugin events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Many thanks to the following for contributing to this release:
 ### Changes
 
 - Fixed Inventory Sync reporting a negative database size [¤598](https://github.com/clusterio/clusterio/pull/598).
+- Fixed plugin events being invoked while a host connection was in an invalid state.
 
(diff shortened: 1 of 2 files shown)


Cooldude2606 (approved):
Changes fix issue addressed. Although, this made me question if the order of `HostConnectionEvent` and and the on `close` of `HostConnection` is correct. Currently, the plugin hook for close will have `this.info.connected === true` because the close handler hasn't ran yet. Should this also be reordered?

Hornwitser:
I don't know. If something encounters that as a problem I suppose it can be fixed then.

PR #613, opened 2024-04-23
Correctly handle module files missing from a previously patched save

### LINKED ISSUE #611: Unable to restart instance after a patched in file is deleted
I updated factorio from 1.1.104 to 1.1.107 using the ingame updater. After updating, I get the following error both on the host and in the web interface developer tools when attempting to restart the running instance to apply the update:

[error] TypeError: Cannot read properties of null (reading 'async')
    at /Users/danielv/project_files/clusterio/packages/host/dist/node/src/patch.js:98:100
    at Array.map (<anonymous>)
    at SaveModule.fromSave (/Users/danielv/project_files/clusterio/packages/host/dist/node/src/patch.js:98:61)
    at /Users/danielv/project_files/clusterio/packages/host/dist/node/src/patch.js:182:135
    at Array.map (<anonymous>)
    at PatchInfo.fromSave (/Users/danielv/project_files/clusterio/packages/host/dist/node/src/patch.js:182:115)
    at patch (/Users/danielv/project_files/clusterio/packages/host/dist/node/src/patch.js:330:37)
    at async Instance.prepareSave (/Users/danielv/project_files/clusterio/packages/host/dist/node/src/Instance.js:661:9)
    at async Instance.handleInstanceStartRequest (/Users/danielv/project_files/clusterio/packages/host/dist/node/src/Instance.js:823:24)

The issue remains after restarting the host and controller, however does not apply to newly created instances.

Danielv123:
Resolves #611

image

Logs missing files instead of throwing an error.

### DIFF
--- a/packages/host/src/patch.ts
+++ b/packages/host/src/patch.ts
@@ -77,16 +77,16 @@ export class SaveModule {
 
 	static async fromSave(json: Static<typeof SaveModule.jsonSchema>, root: JSZip) {
 		const module = new this(lib.ModuleInfo.fromJSON(json));
-		module.files = new Map(
-			await Promise.all(json.files.map(async filename => {
-				const file = root.file(filename);
+		module.files = new Map(await Promise.all(json.files
+			.map(filename => ({filename, file: root.file(filename)}))
+			.filter(({filename, file}) => {
 				if (file === null) {
 					lib.logger.warn(`Missing file ${filename} in save`);
-					return [filename, Buffer.alloc(0)] as const;
 				}
-				return [filename, await file.async("nodebuffer")] as const;
-			}))
-		);
+				return file !== null;
+			})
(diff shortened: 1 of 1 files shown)


Cooldude2606 (commented):
> Log a warning that the file went missing. The current patch logic already replaces all patched in files with files from the modules to patch in.

The above is implemented as was suggested by Hornwitser.
However, is leaving the empty buffers in the map a good idea?

As seen by your edit to line 398/405 you needed to add a check to account for non-existent files in the map.
In the future if we provided a method to extract and save modules from a save file, then we would end up writing empty files if we didnt also account for the empty buffers.
In your issue this was a git ignore file that went missing, and I'd argue an incorrectly empty one is more dangerous than no file at all.

This can be fixed with a `.filter` after the `.map`, should we?

Cooldude2606 (approved):

PR #614, opened 2024-04-23
Use shell:true when running non-user input .cmd files on windows

### LINKED ISSUE #612: installation broken because node vulnerability fix
latest LTS of node broken the installation of clusterio

info Please wait, installing standalone
error Failed to install: spawn EINVAL

blog post
https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2/
release notes
https://nodejs.org/en/blog/release/v20.12.2

how to replicate: use latest LTS node version

i tested on version v20.12.1 its ok, but version v20.12.2 (LTS with the fix) broken the installation

Danielv123:
Required after CVE fix: https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2/
Resolves #612

### DIFF
--- a/packages/create/create.js
+++ b/packages/create/create.js
@@ -108,7 +108,7 @@ async function safeOutputFile(file, data, options={}) {
 async function execFile(cmd, args) {
 	logger.verbose(`executing ${cmd} ${args.join(" ")}`);
 	return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


Cooldude2606 (approved):
The other occurrence of `exec` is only called when the platform is not windows. Therefore this fix works as intended.
Could attempt performance optimisation by only setting the option to true when on windows, but this script only runs once and would increase platform heterogeneity.

PR #616, opened 2024-04-28
Handle bad mod setting types

Hornwitser:
The mod settings stored in Mod Packs can have any type and does not need to exist or match the type defined in the settings prototypes extracted from the game. Gracefully handle mismatched types in the Web UI well enough that it's still possible to operate all the other controls when viewing a Mod Pack.

Since the correct handling the mismatch of types in the Web UI would be overly complicated a new mechanism is introduced during data export which correct any mod settings with the wrong types. This means that as long as a data export is done after changing the mod list the mod settings UI will function properly.

### DIFF
--- a/packages/web_ui/src/components/ModPackViewPage.tsx
+++ b/packages/web_ui/src/components/ModPackViewPage.tsx
@@ -424,7 +424,15 @@ function InputColor(props: {
 	value: lib.ModSettingColor,
 	onChange: (value: lib.ModSettingColor) => void,
 }) {
-	const value = props.value;
+	// In certain edge cases it's possible a value for this mod setting does not exist
+	// or is of the wrong type. Fallback to black if this is the case.
+	function isColor(input: unknown) {
+		if (typeof input !== "object" || input === null) {
+			return false;
+		}
+		return "r" in input && "g" in input && "b" in input && "a" in input;
+	}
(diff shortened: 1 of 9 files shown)


packages/web_ui/src/components/ModPackViewPage.tsx
```diff
@@ -424,7 +424,15 @@ function InputColor(props: {
...
 	onChange: (value: lib.ModSettingColor) => void,
 }) {
-	const value = props.value;
+	// In certain edge cases it's possible a value for this mod setting does not exist
+	// or is of the wrong type. Fallback to black if this is the case.
+	function isColor(input: unknown) {
+		if (typeof input !== "object" || input === null) {
+			return false;
+		}
+		return "r" in input && "g" in input && "b" in input && "a" in input;
```
Danielv123, line 433: Not quite sure where this comes from, but isn't alpha always optional, defaulting to 1?

Danielv123 (approved):
Looks good to me.

packages/web_ui/src/components/ModPackViewPage.tsx
Hornwitser: In the data stored to mod-settings.dat by the game alpha is always present. The values comes from `ModPack.fillDefaultSettings` which sources them from settings prototypes and normalize them with `normalizeColor` which makes sure all components are always present.

Issue #617, opened 2024-04-30
How to use a modded mod-setting.dat file in a server


YR1044:
This repo's readme  says` For managing the mod settings, the mod-settings.dat and mod-list.json files will also be linked/copied over if present in sharedMods, but since Factorio creates these files automatically it might be necessary to delete them from each instance's mods folder for changes to them in sharedMods to take effect.`

but there is no  `sharedMods` folder in my case, and I tried to put my mod-setting.dat in my instance's `mods` folder, but it was overwrited as soon as I start my server up.

How to deal with it? Thanks a lot

YR1044:
I am running in a windows 10 pc with standalone mode

Hornwitser:
Sorry about that, the README is currently out of date.

The current way manage mods and change mod settings is as follows:
1. Under the "Mods" on the controller's web interface upload the mods you wish to use. Alternatively you can copy them to the "mods" folder on the controller and restart the controller.
2. Create a mod pack and add the mods you wish to use to the mod pack.
3. Assign the mod pack to an instance and run an export data action. This option is available under the "More" button when the instance is stopped.
4. The mod settings should now be available to edit in the Mod Pack.

If you already set everything up in single player and don't want to set it up again you could use the command in the [Migrating from alpha 13](https://github.com/clusterio/clusterio/blob/master/docs/migrating-from-2.0.0-alpha.13.md#before-updating) document to create a mod pack string for your configuration and then import that as step 2. Note that there's currently an issue with mods that have leading zeros in the version number when doing that, so if you if you have a mod with a leading zero in the version you will need to remove it from the mod pack and the add it back again for it to work.

PR #618, opened 2024-05-04
Update mod management section of readme

Danielv123:
As mentioned in issue #617, the mod management section of the readme is outdated. I also added a few pictures of the relevant buttons and configuration interface as I know a few people have had issues finding them in the past. Its also useful to give people an idea that this functionality exists and what it does - I can tell you that I did not know it looked like that before I configured it to do this PR.

It looks pretty sweet though - I don't think any other server management tools have the ability to edit mod settings directly like that.


Danielv123:
Oh crap it autoformatted the TOC. Give me a moment

Cooldude2606 (commented):
The addition of pictures is nice. However, I don't think the process is described well.
1) You use "To create a modpack" followed by "you can then create a modpack". It would be better to word this as "There is currently no mod portal integration, so you first need to navigate to the mods page and upload your mods. You can then create a moodpack..."
2) You say "you can then create a modpack" but don't describe how. It should explain that you press create, enter the name and factorio version, then press add and select your mods. An image of the "Add Mod" search element would help here.

Cooldude2606 (approved):
Reads better now +1

PR #619, opened 2024-05-04
Installer promt for http port

### LINKED ISSUE #564: Add prompt for `controller.http_port` during install.
Tl;dr When running `create` for standalone or controller you should be asked to enter a port number to host the web ui on.

### Motivation
When installing a cluster for the first time it should be as seamless as possible. Currently you run into the situation where the default port is already in use and you can not start the controller. The workaround is to manually edit the json for the controller config, this is simple to do but is not seamless.

### Solution
You should be prompted on installation to enter a port number for the web ui to be hosted on. This way a known available port can be selected allowing the controller to start without any issues. After which the web interface can be used to edit the remaining config values which is both safer and less technical.

Danielv123:
Resolves #564

### DIFF
--- a/packages/create/create.js
+++ b/packages/create/create.js
@@ -551,6 +551,25 @@ async function inquirerMissingArgs(args) {
 				},
 			},
 		], answers);
+
+		if (args.httpPort) { answers.httpPort = args.httpPort; }
+		answers = await inquirer.prompt([
+			{
+				type: "input",
+				name: "httpPort",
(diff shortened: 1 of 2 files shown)


Cooldude2606 (approved):
Works as intended. Although are the tests meant to be skipped?

Danielv123:
Tests work fine locally but are slow. They fail entirely in CI for what seems like random reasons. I just disabled them, leaving them there in case anyone wants to mess with it later.

PR #621, opened 2024-05-08
Fix breaking change in pnpn v9 with workspace package linking.

Cooldude2606:
As per the [pnpn v9 change log](https://github.com/pnpm/pnpm/releases/tag/v9.0.0):

> [link-workspace-packages](https://pnpm.io/npmrc#link-workspace-packages): disabled by default. This means that by default, dependencies will be linked from workspace packages only when they are specified using the [workspace protocol](https://pnpm.io/workspaces#workspace-protocol-workspace).

Previously the default was `true` which allowed packages to be linked within the workspace as long as the version was compatible. With the new value of `false` this no longer happens and you must be explicit in the package linking. This resulted in fresh dev installs failling at build and prepare stages. Thankfully, this is an easy change to make and pnpn will [update the version on publish](https://pnpm.io/workspaces#publishing-workspace-packages) to match the workspace versions.

The changes made:
- Add web ui as a dev dep at the root of the project.
- Change all workspace versions to `workspace:*` to allow linking.

Possible change:
- We could update the CI to use pnpn v9 in / after this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
 		"@clusterio/controller": "workspace:*",
 		"@clusterio/ctl": "workspace:*",
 		"@clusterio/host": "workspace:*",
+		"@clusterio/web_ui": "workspace:*",
 		"@clusterio/lib": "workspace:*",
 		"@swc/core": "^1.4.0",
 		"@typescript-eslint/eslint-plugin": "^6.4.1",
--- a/packages/controller/package.json
+++ b/packages/controller/package.json
@@ -19,7 +19,7 @@
 		"node": ">=18"
 	},
 	"dependencies": {
-		"@clusterio/lib": "^2.0.0-alpha.18",
+		"@clusterio/lib": "workspace:*",
 		"@sinclair/typebox": "^0.30.4",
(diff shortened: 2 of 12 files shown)


Hornwitser (approved):
Looks good. I did not find any missed entries, and it builds fine in my dev environment with pnpm 8.

Cooldude2606:
There was some [discussion](https://discord.com/channels/450361298220351489/450361298220351491/1237846719002316860) around if `workspace:*` or `workspace:^` should be used. The difference between the two is their behaviour during `pnpm publish`; the first will insert the exact package version used, while the second will prepend a `^` to allow any compatible version. e.g. `workspace:*` -> `2.0.0-alpha.18` vs `workspace:^` -> `^2.0.0-alpha.18`

It was decided to stay with `workspace:*` because each alpha version of v2.0.0 is expected to contain breaking changes so although `2.0.0-alpha.19` would be accepted by `^2.0.0-alpha.18` we do not maintain the guarantee that it will be compatible. Once we are at full release it would be more appropriate to use `workspace:^`

PR #622, opened 2024-05-15
Fix hard to reproduce crash on join

Danielv123:
Apparently it is possible to run into a crash on join here

image

Apparently its fixable through deleting this from the save:

image

I assume the reason this works is because on being repatched it also reruns the code to update the globals with all players in the save.

Eternity cluster discord link: https://discord.com/channels/1193199619740540978/1193199620336144476/1239838359564582973

I don't have any way to reproduce, but I think this should fix it in the future - at the very least a nil check shouldn't do any harm.

### DIFF
--- a/plugins/inventory_sync/module/inventory_sync.lua
+++ b/plugins/inventory_sync/module/inventory_sync.lua
@@ -357,8 +357,14 @@ end
 
 -- Download inventory from controller
 inventory_sync.events[defines.events.on_player_joined_game] = function(event)
-	-- Send acquire request even if an active download is currently in plogress
 	local player = game.get_player(event.player_index)
+
+	-- It's possible Factorio doesn't invoke the on_player_created event when loading a save in single player
+	if not global.inventory_sync.players[player.name] then
+		create_player(player, false)
+	end
+
+	-- Send acquire request even if an active download is currently in plogress
 	inventory_sync.acquire(player, false)
 
 	-- Clear active upload if it existsi


Hornwitser:
I was able to get the same error by loading up the latest save on https://www.hornwitser.no/eternity-saves/instances/copper_smelting_13/saves/ in Factorio 1.1.107 with mods synced to the save.

Cooldude2606:
I have tested the events in a minimal scenario. (Printing: init, load, created, joined)

My method: create a new single player world, save the world, change username in settings, reload the save.
```
[Name is Cooldude2606]
[Start new game]
On Init
Player Created: Cooldude2606
Player Joined: Cooldude2606
[Save game A]
[Rename to Cooldude260607]
[Load save game A in SP]
On Load
```
I noticed that when I loaded the save, my name does not get updated and I am still playing as `Cooldude2606` despite the rename.

The same behaviour is observed when making a multiplayer world and loading it in single player.
However, if you instead host the save game, then a new player is created and on join is triggered.

If you then save and reload the world after hosting it (resulting in `#game.players > 1`) then name changes are reflected in game.
The `on_player_joined_game` handler will always be raised after load, but `on_player_created` is not.
Although, this seams to be partily expected because `#game.players == 2` indicating no new player was created, and instead the name of player with index 1 was changed.

```
[Name is Cooldude2606]
[Host MP new game]
On Init
Player Created: Cooldude2606
Player Joined: Cooldude2606
[Save game A]
[Rename to Cooldude260607]
[Load save game A in SP]
On Load
[Host save game A in MP]
On Load
Player Created: Cooldude260607
Player Joined: Cooldude260607
[Save game B]
[Rename to Cooldude260608]
[Load save game B in SP]
On Load
Player Joined: Cooldude260608
```

I believe there is an inconsistency between when `on_player_joined_game` is raised (`#game.players > 1`) and when player with index 1 gets renamed (`#game.players > 1 and game.players[name] == nil and not game.is_multiplayer()`).
We should probably raise this on the forums.

Danielv123:
Good job investigating, your fix seems a lot cleaner than mine.

Cooldude2606 (approved):
LGTM. Matches existing code within `on_player_created` and `on_server_startup`

PR #623, opened 2024-05-15
Utility functions

Danielv123:
These are functions I need for plugins, but they feel more like they belong to lib. Factorio versions could be useful in the future as I know there are plans to manage factorio versions with modpacks, memoization is just generally useful especially when calling external APIs.

Thoughts?

### DIFF
--- a/packages/lib/src/memoize_timeout.ts
+++ b/packages/lib/src/memoize_timeout.ts
@@ -1,0 +1,48 @@
+/**
+ * Memoize a function with a timeout. Calling the function again after the timeout will erase the cache.
+ * @param fn Function to memoize
+ * @param time Time in ms to cache results
+ * @returns Memoized function
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function memoizeTimeout(fn: Function, time: number): Function {
+	let cache = new Map();
+	let expiryTimers = new Map() as Map<string, NodeJS.Timeout>;
+
+	return (...args: any) => {
+		// Create hash.
+		const hash = JSON.stringify(args);
+
+		// Erase cache.
+		if (!expiryTimers.has(hash)) {
+			expiryTimers.set(hash, setTimeout(() => {
+				cache.delete(hash);
+				const interval = expiryTimers.get(hash);
+				clearInterval(interval);
+				expiryTimers.delete(hash);
+			}, time));
+		}
+
+		// Find in cache or store new values.
+		if (cache.has(hash)) {
+			return cache.get(hash);
+		}
+		let result = fn(...args);
+		// Handle async functions
+		if (result instanceof Promise) {
+			// Add a Promise for the result to the cache to prevent duplicate calls
+			cache.set(hash, result);
+			return new Promise((resolve, reject) => {
+				result.then((res: any) => {
+					cache.set(hash, res);
+					resolve(res);
+				}).catch((err: any) => {
+					reject(err);
+				});
+			});
+		}
+		// Handle synchronous functions
+		cache.set(hash, result);
+		return result;
+	};
+}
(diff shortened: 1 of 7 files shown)


Cooldude2606:
I like the idea of having lib provide methods for interacting with common external APIs. For the tests, it would be good to have a flag to skip them similar to fast test, something along the lines of "NO_EXTERNAL" this can help reduce the risk of getting rate limited.

It is also good to have common functions in the lib, but the file structure might need some consideration if we plan to add more in the future.

Danielv123:
Yeah, something like no external could be useful, especially for local development. With scraping being something that could randomly break I think it should always run in CI. Might be enough to just run mark it as slowtest?

Not quite sure what you mean with more structure. I also though for a bit it might make sense to namespace the folders as single objects, but that messes with the es modules import pattern. See for example antd having hundreds of exports from a single module without issues.

Cooldude2606:
By structure I meant that the lib would have a lot of file at its root if we start having one file per util function. At the moment its just memozie so its fine, but if more were added then it may be better to add a sub directory for the functions.

packages/lib/src/memoize_timeout.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Memoize a function with a timeout. Calling the function again after the timeout will erase the cache.
+ * @param fn Function to memoize
+ * @param time Time in ms to cache results
+ * @returns Memoized function
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function memoizeTimeout(fn: Function, time: number): Function {
```
Hornwitser, line 8: This erases the the type information of the function passed into it. So for example if you do
```ts
function foo(input: string) {
    return input;
}
const memoized = memoizeTimeout(foo, 1000);
memoized(10, true);
```

Ther'es no type error as the function signature was lost. Instead we can declare a generic param extending `(...args: any) => any` to indicate it's a function (instead of the bad type Function), and use `Parameters<T>` and ReturnType<T> to create a new callable type with the same parameterns and return value. Note that it's not possible to use `T` as the returned value as that's making a promise that the returned type is exactly the same as the input type, which is not the case as a new function is returned that might not have the same properties as the input. See [this StackOverflow answer](https://stackoverflow.com/questions/69783310/type-is-assignable-to-the-constraint-of-type-t-but-t-could-be-instantiated) for an in depth explanation.

```suggestion
export function memoizeTimeout<T extends (...args: any) => any>(
	fn: T,
	time: number
): (...args: Parameters<T>) => ReturnType<T>
```

packages/lib/src/memoize_timeout.ts
```diff
@@ -0,0 +1,48 @@
...
+		const hash = JSON.stringify(args);
+
+		// Erase cache.
+		if (!expiryTimers.has(hash)) {
+			expiryTimers.set(hash, setTimeout(() => {
+				cache.delete(hash);
+				const interval = expiryTimers.get(hash);
+				clearInterval(interval);
+				expiryTimers.delete(hash);
+			}, time));
```
Hornwitser, line 23: When the timer here is created it'll keep the event loop alive when the program exits. So if you do `memoizeTimeout(foo, 300e3)` for example and the function is called before shutdown it'll hang the application for 5 minutes before exiting. Calling [`.unref()`](https://nodejs.org/api/timers.html#timeoutunref) on the timeout will prevent this from happening.

```suggestion
			expiryTimers.set(hash, setTimeout(() => {
				cache.delete(hash);
				const interval = expiryTimers.get(hash);
				clearInterval(interval);
				expiryTimers.delete(hash);
			}, time).unref());
```

Hornwitser (changes requested):
The memoize function needs some work to be usable. Other than that it looks good to me.

PR #624, opened 2024-05-17
Fix undefined stack on some IPC errors

Cooldude2606:
Quick fix for IPC errors outputing `undefined` when there is no stack trace.
Reported by [Laar on discord.](https://discord.com/channels/450361298220351489/450361298220351491/1241120354873643112)


Danielv123 (approved):
I have encountered this issue before, nice to have an actual solution

PR #625, opened 2024-05-18
Fix loopback routing message and add tests

Cooldude2606:
Fixes the incorrect error message when an instance attempts to make a request to itself.
Also adds test cases to prevent regression in this area.

~~Draft: It is noted that events log a warning rather than error, intended behaviour needs to be discussed.~~
Events have been changed to throw an error before being sent over a connector if it would loopback. If the event has already been sent and a loopback is discovered then it will be dropped and a warning emited.

Sending events on the controller to invalid addresses now throws an error rather than silently returning. This is to match the new behaviour of events throwing an error before being sent and the existing behaviour of requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,20 @@ Many thanks to the following for contributing to this release:
 [@username](https://github.com/username)
 -->
 
+## Version 2.0.0-alpha.19
+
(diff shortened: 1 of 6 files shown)


Danielv123 (approved):
Looks good to me

PR #626, opened 2024-05-18
Allow declaring custom config input components for objects

Danielv123:
Input components allow us to customize the config field editing experience in plugins, but due to an oversight this did not include object definitions.

An alternative way of implementing this fix would be to define the default editor for objects as an inputComponent, but that doesn't quite fit with how inputComponents are explicitly selected and not based on the type in the definition properties.

Use of both inputcomponents and object config definitions are rare, so I expect this will be sufficient.

### DIFF
--- a/packages/web_ui/src/components/BaseConfigTree.tsx
+++ b/packages/web_ui/src/components/BaseConfigTree.tsx
@@ -270,7 +270,7 @@ function computeTreeData(
 				children: [],
 			};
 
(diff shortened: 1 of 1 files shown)


Cooldude2606 (approved):
I'm not getting any errors from the assertion, and its not possible for an object key to be undefined so it ever return the wrong result. So I guess this is fine despite the types disagreeing.

PR #627, opened 2024-05-18
Add useInstanceConfig hook

Danielv123:
Currently the web interface implements configs by pulling the state directly in the components where they are used. I need to access the instance config for a plugin, which would require me to duplicate parts of the instance config retrival code. Instead, I implement it here under lib. This allows 3rd party consumers to get immediate access if we decide to implement a subscription model for configs in the future.


Danielv123:
What is up with npm modules having native bindings that always inevitably fails >_>

Cooldude2606 (approved):
I don't see any issues here and refactoring the config component seams too large of task and should be considered later.

PR #628, opened 2024-05-18
Fix test starting alt host under windows

Cooldude2606:
One of the tests which uses the alt host was mistakening not skipped on windows causing another test to fail.
However, it was found that all tests which skip on windows pass without leaving any child processes behind, this needs further investigation.


Danielv123:
Is there any reason not to merge this if it was just mistakenly left out in one of the tests?

Cooldude2606:
Only if we want to consider looking into running all tests on windows. But that can be done as its own thing. So no not really any reason not to merge, just needing a second pair of eyes on it.

PR #630, opened 2024-05-19
Add devcontainer configuration file

Danielv123:
When I am on the go with shitty wifi it is very difficult to install npm modules locally for PR reviews etc. Codespaces is great in that it fixes most of the issues with a high latency connection like that. This devcontainer file provides enough setup that `pnpm test` passes when ran without having to look up factorio download urls etc.


Cooldude2606 (approved):
Spoke on discord. Addition of this file has been justified. It can be merged once we fix CI.

PR #631, opened 2024-05-20
Fix broken types

Danielv123:
Some time between antd 5.15.0 and 5.17.3 they moved around some types, which caused our builds to break. This fixes the issue.


Cooldude2606 (approved):
Small and justified change, and CI is passing. LGTM.

PR #632, opened 2024-05-20
Codecov webpack bundle analyzer

Danielv123:
We track code coverage with codecov. I see they have a new feature to track bundle sizes, could be useful or at least interesting to look at.


Cooldude2606 (commented):
Seams like something interesting. I don't see any issues in the added code, but this is something to discuss.

PR #633, opened 2024-05-22
Update documentation links for typescript change

Laar:
Checked all links in the documentation after finding one that still pointed to the old javascript structure.


Cooldude2606 (approved):
All updated links are valid. I don't see any missed.

PR #635, opened 2024-05-24
Add bun compatibility

Danielv123:
When doing development on terrible networks pnpm installs usually fail. Bun is an alternative nodejs runtime and package manager which seems to be a whole lot more resilient.

Bun requires using the npm workspace location specification rather than the pnpm yaml file. This does not conflict, pnpm will continue to operate undisturbed.

It also complained about my earlier type fix. This time I changed it to use the same type declaration as in the antd documentation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,11 @@
 	"engines": {
 		"node": ">=18"
 	},
+	"workspaces": [
+		"packages/*",
+		"plugins/*",
+		"external_plugins/*"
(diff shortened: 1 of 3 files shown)


Cooldude2606 (approved):
Changes were discussed on [discord](https://discord.com/channels/450361298220351489/475279193777635337/1243551652670341221) before PR was created.

PR #639, opened 2024-05-29
Fix 634 - Fatel error when shutting down host from webui

### LINKED ISSUE #634: Stopping host from WebUI triggers error
### Error
When shutting down a host from the web UI the following error is seen in the terminal that runs the host
> [fatal] Uncaught exception:
TypeError: Cannot read properties of undefined (reading 'equals')
    at Host.sendEvent ([project root]/clusterio/packages/lib/dist/node/src/link/link.js:422:32)
    at Host.sendTo ([project root]/clusterio/packages/lib/dist/node/src/link/link.js:381:25)
    at Host.send ([project root]/clusterio/packages/lib/dist/node/src/link/link.js:372:21)
    at LinkTransport.log ([project root]/clusterio/packages/lib/dist/node/src/logging_utils.js:150:23)
    at LinkTransport._write ([project root]/clusterio/node_modules/.pnpm/winston-transport@4.7.0/node_modules/winston-transport/modern.js:82:19)
    at doWrite ([project root]/clusterio/node_modules/.pnpm/readable-stream@3.6.2/node_modules/readable-stream/lib/_stream_writable.js:390:139)
    at writeOrBuffer ([project root]/clusterio/node_modules/.pnpm/readable-stream@3.6.2/node_modules/readable-stream/lib/_stream_writable.js:381:5)
    at Writable.write ([project root]/clusterio/node_modules/.pnpm/readable-stream@3.6.2/node_modules/readable-stream/lib/_stream_writable.js:302:11)
    at DerivedLogger.ondata ([project root]/clusterio/node_modules/.pnpm/readable-stream@3.6.2/node_modules/readable-stream/lib/_stream_readable.js:629:20)
    at DerivedLogger.emit (node:events:531:35)

### Suspected cause
In  #625 loopback routing was completely disallowed. This is done by the following lines in [packages/lib/link/link.ts]
```
if (this.connector.src.equals(dst)) {
	throw new Error(`Message would return back to sender ${dst}.`);
}
```
This looks good considering the constructor of BaseConnector
```
	constructor(
		public src: libData.Address,
		public dst: libData.Address,
	) {
		super();
	}
```
However, in WebSocketClientConnector this type guarantee is broken by the `_reset` and `_invalidate` methods:
```
	_reset() {
		...
		this.src = undefined as any;
	}

	_invalidate() {
		...
		this.src = undefined as any;
	}
```
Indeed commenting these lines out removes the error (but with unknown consequences).

Cooldude2606:
Depends on #638 for webpack build fix. Fixes #634 

Fixes the fatel error on the host when shutdown from the webui. Achieved by adding `abstract get valid` to BaseConnector which indicates when it is safe to send messages on a link. This property only needs to be checked on fringe cases during setup and tear down contexts such as attempting to send logs after disconnection.

### DIFF
--- a/packages/lib/src/link/connectors.ts
+++ b/packages/lib/src/link/connectors.ts
@@ -34,6 +34,7 @@ export abstract class BaseConnector extends events.EventEmitter {
 	}
 
 	protected abstract send(message: libData.MessageRoutable): void;
+	abstract get valid(): boolean;
 
 	forward(message: libData.MessageRoutable) {
 		const seq = this._seq;
@@ -345,6 +346,14 @@ export abstract class WebSocketBaseConnector extends BaseConnector {
 	 */
 	get hasSession() {
 		return ["connected", "resuming"].includes(this._state);
+	}
+
+	/**
+	 * True if the connector is valid and can accept new messages,
+	 * only needs to be checked on fringe cases during setup and tear down.
+	 */
+	get valid() {
+		return this.hasSession;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Danielv123 (approved):
Finally, this is a nice QOL for development :)

Issue #589, opened 2024-02-11
Add clean script


Hornwitser:
It would be useful to have a `clean` script which deletes the `{packages,plugins}/*/dist` folders and can be used before publishing to make sure no leftover files from previous builds are accidentally included into the published packages.

This script should ideally work on both Windows and Linux, and be documented as part of the publishing procedure in docs/contributing.md.

Danielv123:
How about making a CI task for publishing instead? That reduces the chance of human error significantly.

PR #640, opened 2024-05-29
Add clean script for remove dist and other artifacts

Cooldude2606:
Adds a script to remove dist directories from all packages and optionaly remove other artifacts such as test files or configs.
This is based on the feature request #589 but it does not attempt to implement a CI publish action.


clean.js
```diff
@@ -0,0 +1,114 @@
...
+		case "fast": {
+			removeBuildArtifacts();
+			break;
+		}
+
+		default: {
+			// eslint-disable-next-line no-console
+			console.error("First argument must be one of: all, tests, fast, all-dry, tests-dry, fast-dry");
+		}
+	}
```
Danielv123, line 111: While this is pretty simple with so few arguments, is there a reason why you aren't using yargs here? We use it for the other packages, so it is already included, and it gives nice expressive cli usage hints, errors etc.

package.json
```diff
@@ -1,6 +1,7 @@
 {
 	"name": "root",
 	"private": true,
+	"files": [],
```
Danielv123, line 4: According to the docs this is a whitelist of files to publish - why have you set it, and why is it blank?

clean.js
Cooldude2606: Didn't even think about it. Seams like a good idea so i'll add that.

package.json
Cooldude2606: This was the fix I found for [node/no-unpublished-require](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-require.md) on line 11 of clean.js
What is needed is for clean.js to be excluded from publishing, which can be done with either the files property or an npmignore file.
Since the root package.json doesnt describe an actual package and is marked as private (which prevent publishing) it is safe to have it list no files. This method means we do not need to maintain an npmignore file if more helper scripts are added at a later date.

Danielv123 (approved):

PR #641, opened 2024-06-02
Improve config, message, and module templates

Cooldude2606:
Makes the following changes to the plugin templates:
- Add webpack if config definations are included
- Add option to not include config definitions
- Only include message handling when multiple contexts are present
- Better comment explaining the edge cases of on_server_startup
- Lua module will include an empty instance plugin if no instance plugin is present
- Removed redudent checks for web presence in web template files
- Removed web entry point when web plugin is not defined

### DIFF
--- a/packages/create/templates/common/module/control.lua
+++ b/packages/create/templates/common/module/control.lua
@@ -7,7 +7,9 @@ local MyModule = {
 }
 
 --- global is 'synced' between players, you should use your plugin name to avoid conflicts
--- setupGlobalData should either be removed or called during clusterio_api.events.on_server_startup
+-- The following helper function is desync safe and avoids conflicts if called during on_server_startup
+-- There are edge cases in the following events: pre player left, player left, console command, config changed
+-- If you use any of these events and access globalData then make sure to call this function first
 local globalData = {}
 local function setupGlobalData()
 	if global["__plugin_name__"] == nil then
@@ -28,7 +30,7 @@ local function bar()
 	game.print("bar")
 end
 
---- Clusterio provides a few custom events, on_server_startup is the most useful and should be used in place of on_load
+--- Clusterio provides a few custom events, on_server_startup is the most useful and should be used in place of on_load and on_init
 MyModule.events[clusterio_api.events.on_server_startup] = function(event)
 	setupGlobalData()
 	game.print(game.table_to_json(event))
--- a/packages/create/templates/common/module/control.lua
+++ b/packages/create/templates/common/module/control.lua
@@ -6,9 +6,9 @@ local MyModule = {
 	on_nth_tick = {},
 }
 
---- global is 'synced' between players, you should use your plugin name to avoid conflicts
--- The following helper function is desync safe and avoids conflicts if called during on_server_startup
--- There are edge cases in the following events: pre player left, player left, console command, config changed
+--- global is 'synced' between players, you should use your plugin name to avoid name conflicts
+-- The following helper function is desync safe and avoids name conflicts after being called during on_server_startup
+-- There are edge cases in the following events: pre player left, player left, console command, and config changed
 -- If you use any of these events and access globalData then make sure to call this function first
 local globalData = {}
 local function setupGlobalData()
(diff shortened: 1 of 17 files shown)


packages/create/templates/common/module/control.lua
```diff
@@ -28,7 +30,7 @@ local function bar()
 	game.print("bar")
 end
 
---- Clusterio provides a few custom events, on_server_startup is the most useful and should be used in place of on_load
+--- Clusterio provides a few custom events, on_server_startup is the most useful and should be used in place of on_load and on_init
```
Laar, line 33: This is not a replacement for on_load. Moreover, I think a separate on_load is needed to restore the global alias when joining a multiplayer.. That on_load has to also work on first load (i.e. when the `global["__plugin_name__"] == nil` on the server, or the case where global["__plugin_name__"] still contains premigrated data.

Cooldude2606: > I think a separate on_load is needed to restore the global alias when joining a multiplayer.

As tested [here](https://discord.com/channels/450361298220351489/450361298220351491/1246843044167422042), there are only a few events in which the global alias would be invalid, these events are mentioned in the comment on line 11. It is also safe to setup metatables on global data within this method, as long as the same edge case events are covered. (Although you should use `register_metatable` for this)

> This is not a replacement for on_load.

The only part which is not possible is the registering of conditional event handlers which must still happen within `on_load`. I find this to be an anti pattern within scenario code and should only be used within mods, but I can add this as an additional comment. Any use not mentioned here or the previous comment are considered [illegitimate uses of the event](https://lua-api.factorio.com/latest/auxiliary/data-lifecycle.html#[4]-).

> That on_load has to also work on first load (i.e. when the `global["__plugin_name__"] == nil`)

It is impossible to restore a reference when the value is nil. It is also impossible to modify global to not be nil during on_load. Your only option here is to use on_server_startup and to make sure the edge case events are accounted for. The same is true for "premigrated data" although we currently don't have any method of telling a plugin that it has been updated to a new version, so that is currently undefined behaviour.

Laar: > As tested here, there are only a few events in which the global alias would be invalid, these events are mentioned in the comment on line 11. It is also safe to setup metatables on global data within this method, as long as the same edge case events are covered. (Although you should use register_metatable for this)

As far as I see it is not called from the on_load() of the module. When joining the game the game is saved, transmitted and loaded by the new player. During the load of the save no method is called. So for the joining player the `globalData` variable will *always* be `{}`. This is a problem as soon as there is actually data stored in the global, as that will not be in `globalData` for the client, but will be for the server. Hence the `on_load()` function [here](https://discord.com/channels/450361298220351489/450361298220351491/1246805038861652080)

> The only part which is not possible is the registering of conditional event handlers which must still happen within on_load. I find this to be an anti pattern within scenario code and should only be used within mods, but I can add this as an additional comment. Any use not mentioned here or the previous comment are considered [illegitimate uses of the event](https://lua-api.factorio.com/latest/auxiliary/data-lifecycle.html#%5B4%5D-).

That link specifies "The [global](https://lua-api.factorio.com/latest/auxiliary/global.html) table can be accessed and is safe to read from, but not write to, as doing so will lead to an error.", yet this method modifies the global table.

> It is impossible to restore a reference when the value is nil. It is also impossible to modify global to not be nil during on_load. Your only option here is to use on_server_startup and to make sure the edge case events are accounted for. The same is true for "premigrated data" although we currently don't have any method of telling a plugin that it has been updated to a new version, so that is currently undefined behaviour.

And that is precisely why I would expect that the on_load() implementation has a giant precautionary warning attached to it.

Cooldude2606: Attempted to make a [module](https://gist.github.com/Cooldude2606/16f48622d3eef1265fc91b39a98faa87) for this and you were correct that on_load still needs to exist, although the handler only needs to do is assign the alias and nothing else, init and migration of global can happen within server startup. I will modify the template to match the patterns used on the working module.

Danielv123 (approved):

PR #647, opened 2024-07-04
Host shutdown on duplicate connection

### LINKED ISSUE #350: Shutdown slave when another one connects with the same id
Currently if two instances of the same slave is started up then an infinite reconnect loop happens where each take turn in disconnecting the other from the master. Something should be done to break this loop. My thoughts are either disallow connecting when the connection is active, or shutdown the slave if its disconnected due to another connection with the same id being received on the master.

Cooldude2606:
When the controller detects a connection from a host with an id it already has a connection from, then it will close the existing connection in favour of the new one. If the orignal process is still responsive then it will attempt to reconnect and create an infinite loop between the two processes fighting over the controller connection.

This PR causes the orginal process to shutdown if its connection is closed because of a duplicate connection. This stops the infinite reconnect loop and closes #350. I believe this is the correct approach to take because it cleans up old / forgot host processes.

Although there is a common error of duplicate host ids resulting from coping host configs between machines. In this case it would be better to prevent new connections when the remote addresses of the connections are different.

### DIFF
--- a/packages/controller/src/WsServerConnector.ts
+++ b/packages/controller/src/WsServerConnector.ts
@@ -114,7 +114,7 @@ export default class WsServerConnector extends lib.WebSocketBaseConnector {
 	_timedOut() {
 		logger.verbose("Connector | Connection timed out");
 		this._reset();
-		this.emit("close");
+		this.emit("close", lib.ConnectionClosed.Timeout, "Connection timed out");
 	}
 
 	_attachSocketHandlers() {
@@ -131,7 +131,7 @@ export default class WsServerConnector extends lib.WebSocketBaseConnector {
 
 			if (this._closing) {
 				this._reset();
-				this.emit("close");
+				this.emit("close", code, reason);
 
 			} else {
 				this._state = "resuming";
@@ -188,7 +188,7 @@ export default class WsServerConnector extends lib.WebSocketBaseConnector {
 
 		if (this._state === "resuming") {
 			this._reset();
-			this.emit("close");
+			this.emit("close", lib.ConnectionClosed.Reset, "Forced Closed");
 			return;
 		}
 
@@ -224,7 +224,7 @@ export default class WsServerConnector extends lib.WebSocketBaseConnector {
 	setClosing() {
 		if (this._state === "resuming") {
 			this._reset();
-			this.emit("close");
+			this.emit("close", lib.ConnectionClosed.Reset, "Forced Closed");
 			return;
 		}
 
--- a/packages/controller/src/WsServerConnector.ts
+++ b/packages/controller/src/WsServerConnector.ts
@@ -188,7 +188,7 @@ export default class WsServerConnector extends lib.WebSocketBaseConnector {
 
 		if (this._state === "resuming") {
 			this._reset();
-			this.emit("close", lib.ConnectionClosed.Reset, "Forced Closed");
+			this.emit("close");
 			return;
 		}
 
@@ -224,7 +224,7 @@ export default class WsServerConnector extends lib.WebSocketBaseConnector {
 	setClosing() {
 		if (this._state === "resuming") {
 			this._reset();
-			this.emit("close", lib.ConnectionClosed.Reset, "Forced Closed");
+			this.emit("close");
 			return;
 		}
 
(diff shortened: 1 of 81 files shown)


packages/lib/src/link/connectors.ts
```diff
@@ -9,6 +9,25 @@ import { logger } from "../logging";
...
+ * Numbered codes describing why a connection was closed
+ * See: https://github.com/Luka967/websocket-close-codes
+ */
+export enum ConnectionClosed {
+	// Codes between 1000 and 1015 are pre-allocated for web sockets
+	Normal = 1000, // CLOSE_NORMAL
+	ServerQuit = 1001, // CLOSE_GOING_AWAY
+	ProtocolError = 1002, // CLOSE_PROTOCOL_ERROR
+	PolicyError = 1008, // Policy Violation
+	ServerError = 1011, // Server Error
```
Hornwitser, line 22: Shouldn't this use the standard namings that are available in the registry? see https://www.iana.org/assignments/websocket/websocket.xhtml#close-code-number

Normal -> NormalClosure
ServerQuit -> GoingAway
PolicyError -> PolicyViolation
ServerError -> InternalError

packages/lib/src/link/connectors.ts
```diff
@@ -9,6 +9,25 @@ import { logger } from "../logging";
...
+
+	// Codes after 4000 are available for applications
+	// Ours are somewhat based on http client error codes
+	BadRequest = 4000,
```
Hornwitser, line 26: While we're inspired by HTTP, I don't think copying out-right makes sense. The "Request" here would be referring to receiving a malformed WebSocket message, which is not the same as a request in http, and not the same as what "Request" refers to in link layer.

Perhaps call it BadMessage? Or MalformedMessage.

packages/controller/src/WsServerConnector.ts
```diff
@@ -114,7 +114,7 @@ export default class WsServerConnector extends lib.WebSocketBaseConnector {
 	_timedOut() {
 		logger.verbose("Connector | Connection timed out");
 		this._reset();
-		this.emit("close");
+		this.emit("close", lib.ConnectionClosed.Timeout, "Connection timed out");
```
Hornwitser, line 117: I don't think exposing the WebSocket codes in the close event of the connector is a good idea. The user of the connector which is the Link class has no concept of WebSockets, and in some cases uses virtual connectors which doesn't involve WebSockets at all.

The Connector is supposed to be abstraction that handles the WebSockets part for Link so that what Link can use the connector as a way to send messages to the other side.

Hornwitser (changes requested):
I don't like the naming of the close codes and them being passed to downstream event handlers. See review comments.

packages/lib/src/link/connectors.ts
Cooldude2606: Orginaly I picked names based on the error messages we already used. But I'm not attached to them so I'll update to iana standard.

packages/lib/src/link/connectors.ts
Cooldude2606: I like the sound of Malformed Message, I'll update to that.

packages/controller/src/WsServerConnector.ts
Cooldude2606: Orginaly the code for killing the host was handled by the host; which meant I needed to pass the close reason on to detect it. However, during development I removed this code and found a cleaner way by emiting errors. Passing the code onwards is a reminent from then which I didn't remove as I suspect it could have a use later. Although you raise a good point for why it should not be included so I'll remove it.

Hornwitser: Any particular reason why there's still 5 `this.emit` adding the WebSocket codes to the events left in the PR?

Cooldude2606: None, they should have been removed and I missed them.

Hornwitser (approved):
All issues have been addressed. Code looks excellent.

Cooldude2606:
Failing test is unrelated to changes made in this PR. It has been corrected in #662

PR #649, opened 2024-07-06
Fix missing codecov upload token

Cooldude2606:
As title says. Repo admin will be required to add the token to secrets.


Danielv123 (approved):
Codecov token was apparenty added 3 months ago

PR #655, opened 2024-09-02
Fix Subscription Issues

Cooldude2606:
Two major changes here:
1) Removed some assumptions that were made that prevented subscription listeners from working on an instance.
This meant allowing the host to forward the event from the instance to the controller; correctly storing subscription listeners by address rather than by connection; and a bunch of network and event stuff to correctly clean up a listener.
2) Removed the assumption that all subscriptions events contained the `updates` property.
This meant fixing inconsistent types; always checking if the property existed; fixing lots of the tests that made incorrect assumptions.

Minior Features:
- There is output made during the start of tests to indicate something is happening when start up is slow.
- Allow for silent testing to reduce the amount of text output to the console.
- Fixed double firing of listener callbacks on first connection.
- Fixed synced being set to true too early on first connection.
- Fixed edge case interaction between last response and last snapshot for values with updated at as 0.

### DIFF
--- a/packages/controller/src/BaseConnection.ts
+++ b/packages/controller/src/BaseConnection.ts
@@ -35,7 +35,7 @@ export default class BaseConnection extends lib.Link {
 
 		this.handle(lib.SubscriptionRequest, this.handleSubscriptionRequest.bind(this));
 		this.connector.on("close", () => {
-			this._controller.subscriptions.unsubscribe(this);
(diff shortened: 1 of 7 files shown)


psihius (approved):

Hornwitser (approved):
Looks good to me. Though I would have preferred if unrelated changes implementing silent tests was a separate commit instead of baked into the one fixing subscritions.

PR #663, opened 2024-09-16
Implement shared Factorio credentials

Hornwitser:
Adds a new credential option to config entries and uses it to add a Factorio username and token config option to the controller and hosts. The purpose of this is to allow setting the Factorio username and token once in the controller config and have the hosts fetch it from the controller when they need it to start instances published on the server list.

Because we sometimes have large clusters with dozens of people donating hardware there's also an option for the controller to not share the token with hosts. In this case the token can be configured on each host, and instances assigned to a given host will use that token.

Since tokens are sensitive and more open sharing of config access has been something that hasn't been done solely because "it contains tokens" credential options function as write only configs that can't be read back remotely (unless configured so, which the factorio_token config is not configured to do). This doesn't really pose a problem because you rarely if ever need to read back the token once you've set it. I couldn't find any simple way to implement knowing if an optional credential is set or not, so there's no way to tell in the UI or API if a credential has been set remotely. Another consideration done is that the credentials are removed from server-settings.json after the Factorio server has started up, this is because copy and paste of whole instance folders is an expected use case for moving instances between hosts and even clusters, and the credentials shouldn't accidentally come along in this case.

Note that there's one caveat when using the Factorio username and token configs and that is if "username" and "token" is set in factorio.settings even if it's to empty strings they will take precedence and override the controller/host credentials.

### DIFF
--- a/packages/host/src/Instance.ts
+++ b/packages/host/src/Instance.ts
@@ -583,11 +583,23 @@ rcon.print(game.table_to_json(players))`.replace(/\r?\n/g, " ");
 	 * entries from the given settings object.
 	 *
 	 * @param overrides - Server settings to override.
+	 * @param includeCredentials - Include Factorio username and token from host/controller config.
 	 * @returns
 	 *     server example settings with the given settings applied over it.
 	 */
-	async resolveServerSettings(overrides: Record<string, unknown>) {
+	async resolveServerSettings(overrides: Record<string, unknown>, includeCredentials: boolean) {
 		let serverSettings = await this.server.exampleSettings();
+		if (includeCredentials && !overrides.username && !overrides.token) {
+			let credentials = {
+				username: this._host.config.get("host.factorio_username") ?? undefined,
+				token: this._host.config.get("host.factorio_token") ?? undefined,
+			};
+			if (!credentials.username && !credentials.token) {
+				Object.assign(credentials, await this.sendTo("controller", new lib.GetFactorioCredentialsRequest()));
+			}
+			if (credentials.username) { serverSettings.username = credentials.username; }
+			if (credentials.token) { serverSettings.token = credentials.token; }
+		}
 
 		for (let [key, value] of Object.entries(overrides)) {
 			if (!Object.hasOwnProperty.call(serverSettings, key)) {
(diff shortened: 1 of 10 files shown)


packages/host/src/Instance.ts
```diff
@@ -583,11 +583,23 @@ rcon.print(game.table_to_json(players))`.replace(/\r?\n/g, " ");
...
+		if (includeCredentials && !overrides.username && !overrides.token) {
+			let credentials = {
+				username: this._host.config.get("host.factorio_username") ?? undefined,
+				token: this._host.config.get("host.factorio_token") ?? undefined,
+			};
+			if (!credentials.username && !credentials.token) {
+				Object.assign(credentials, await this.sendTo("controller", new lib.GetFactorioCredentialsRequest()));
+			}
+			if (credentials.username) { serverSettings.username = credentials.username; }
+			if (credentials.token) { serverSettings.token = credentials.token; }
```
Cooldude2606, line 601: When starting an instance it is not clear which credentials are being used. This makes it easy to misconfigure and hard to diagnose. This can be fixed by adding a log message when the credentials are injected into the settings file: `Using Factorio credetials from controller/host/instance config.`

packages/web_ui/src/components/BaseConfigTree.tsx
```diff
@@ -43,6 +43,9 @@ function renderInput(inputComponents: Record<string, InputComponent>, def: lib.F
 		return <Checkbox/>;
 	}
 	if (def.type === "string") {
+		if (def.credential) {
+			return <Input.Password/>;
```
Cooldude2606, line 47: With the addition of a password input, browsers will recognise this as a login form and attempt auto complete and ask the user if their want to save the password. This can be fixed using `<Input.Password autocomplete="new-password"/>` ([src](https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion#managing_autofill_for_login_fields))

Cooldude2606 (commented):
I was unable to test the credentials fully because of the wifi I am currently forced to use; however, I did not find anything during my manual code analysis that would suggest this feature does not work.

packages/web_ui/src/components/BaseConfigTree.tsx
Hornwitser: I set the token to `section-factorio new-password` and the username to `section-factorio username`. Hopefully that should be well behaved if more credentials are added. To make this possible I added an autoComplete entry in the field definition which sets the autocomplete property. This also makes it possible for plugins to fiddle with it should they face conflicts with it.

packages/host/src/Instance.ts
Hornwitser: Good idea. I implemented this. And also spent way to long trying to figure out why it didn't work. Turns out it would load external credentials if factorio.settings.token is an empty string, and then overwrite the loaded credential with the empty string 🤦.

Cooldude2606: Nice that we already caught a bug by adding logging :D

packages/web_ui/src/components/BaseConfigTree.tsx
Cooldude2606: My browser no longer auto fills on page load, when previously it did. As discussed on discord there doesn't seam to be a perfect solution and this is good enough.

Cooldude2606 (approved):
I was able to test it better now, and both of my previous comments have been addressed so I'm happy with this.

PR #664, opened 2024-09-21
Add InputRole as a drop down selector

Cooldude2606:
Adds a role drop down input method and changes controller default role to use it.
Also contains change log updates I forgot to include.


Hornwitser (approved):
lgtm. Tested changing the default role in the web ui and it works as intended.

PR #665, opened 2024-09-21
Add hidden and readonly config properties

Cooldude2606:
Adds hidden and readonly config properties which are used by the web interface to inform the user that a field can not be modified or is compeltely hidden due to there been alternative methods of accessing the data. These values do not implement any security logic and are commented as such. Command line clients such as clusterioctl can, but are not expected to, follow these hints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Many thanks to the following for contributing to this release:
 - Monorepos are now supported for external plugins [#654](https://github.com/clusterio/clusterio/pull/654).
 - Changed default role selection to be a dropdown menu rather than id [#664](https://github.com/clusterio/clusterio/pull/664).
 - Added credential option to config entries to allow writing sensitive data that can't be read back remotely.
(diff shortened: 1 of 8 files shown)


Hornwitser (commented):
Should the built-in input types like mod-pack and role also support the read-only flag? What about object configs?

In theory we could implement read-only server side by specifying which parties can write to a value. Most of the logic already exists with the credentials change, as there's now a distinction between the access checks for reading a value and writing to a value.

Hornwitser (approved):
Looks good and is good to merge.

Two things I will note is that the error message when attempting to write to read only field from ctl might be a bit confusing as it says "Error: Field 'controller.test_readonly' is not accessible from control" rather than saying the field is not writable. And that having the property setting read access being called "credential" and the property setting write access being called "readonly" with the fall back to both being called "access" is kind confusing. But fixing these are out of scope for this PR.

Cooldude2606:
Strong agree on the naming being bad; "readonly" was selected to match the pattern of "if present it becomes a X field except for at Y locations". This is defiantly on the list of improvements to make after this PR.

Issue #668, opened 2024-10-27
Install problem ?


b4xster:
`factorio@b4xnuc:~/clusterio$ npm init "@clusterio"

> npx
> clusteriocreate

? Operating mode to install Standalone (install both controller and host on this computer)
? Admin account name XXXX
? Public DNS/IP address of this server XXXX
? (Linux only) Automatically download latest factorio release? Yes
? Plugins to install
fatal
+------------------------------------------------------------+
| Unexpected error occured installing clusterio, please      |
| report it to https://github.com/clusterio/clusterio/issues |
+------------------------------------------------------------+
TypeError: Cannot read properties of null (reading '0')
    at downloadLinuxServer (/home/factorio/.npm/_npx/860a49d89133e22f/node_modules/@clusterio/create/create.js:718:78)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async main (/home/factorio/.npm/_npx/860a49d89133e22f/node_modules/@clusterio/create/create.js:846:3)

factorio@b4xnuc:~/clusterio$
`
thx for help :/

Cooldude2606:
This is caused by the following line, we will look into how the name has been changed for factorio 2.0 and update accordingly.
`const version = filename.match(/(?<=factorio_headless_x64_).*(?=\.tar\.xz)/)[0];`

For now you can answer "no" to downloading the latest stable and install it your self under the factorio dictionary. Ie `clusterio/factorio/2.0.10`

Cooldude2606:
Re-openning so we can track that a solution has not been merged into main.

PR #671, opened 2024-10-27
Added 2.0 zip file name to version matcher

Cooldude2606:
Adds support for the new zip file name used in 2.0. e.g. `factorio-headless_linux_2.0.10.tar.xz` vs the previous `factorio_headless_x64_1.1.110.tar.xz`

Also raises a more clear error message should there be another change in the future.

Fixes: Install problem ? #668


Hornwitser (approved):
Looks good to me.

Hornwitser:
Might want to add a changelog entry for it.

PR #672, opened 2024-10-27
Update CI to support testing of multiple factorio versions

Cooldude2606:
We have plans to support the latest version and one minor previous. 2.0 will be added as `latest` to CI once enough changes have been made to support it.


Hornwitser (approved):
I suppose this will work.

PR #673, opened 2024-10-28
Mark RCON password as a credential

Cooldude2606:
Makes RCON password a credential field which can only be read by the instance.

### DIFF
--- a/packages/lib/src/config/definitions.ts
+++ b/packages/lib/src/config/definitions.ts
@@ -398,8 +398,8 @@ export class InstanceConfig extends classes.Config<InstanceConfigFields> {
 			optional: true,
 		},
 		"factorio.rcon_password": {
-			credential: ["instance"],
-			description: "Password for RCON, randomly generated if null",
+			credential: ["host"],
+			description: "Password for RCON, randomly generated if null. Does not persist between host transfers.",
 			restartRequired: true,
 			type: "string",
 			optional: true,
--- a/packages/lib/src/config/definitions.ts
+++ b/packages/lib/src/config/definitions.ts
@@ -398,6 +398,7 @@ export class InstanceConfig extends classes.Config<InstanceConfigFields> {
 			optional: true,
 		},
 		"factorio.rcon_password": {
+			credential: ["instance"],
 			description: "Password for RCON, randomly generated if null",
 			restartRequired: true,
 			type: "string",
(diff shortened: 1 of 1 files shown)


Hornwitser:
Would this even work? What happens if the instance is moved to another host? The controller keeps the authoritative copy of the instance config, but it can't read it? Also `instance` is not supposed to be a valid location any more, as it's equivalent in access to the host and I merged it into `host` a while back.

Cooldude2606:
Was not aware of the merge of `instance` into `host` I will update accordingly.
As for not including controller in it, is this not also how `host.factorio_token` works? If it is moved to a new host then a new password should be set because it is assumed to be a different physical machine where it may have a different owner.

Hornwitser:
Host configs don't move, a different host has a different config. But instances do move. Though I'm not sure what utility RCON password has outside of a host, so maybe this makes sense. It'll cause the password to be wiped when the instance is moved though.

Hornwitser:
I tested this an the rcon password is still persisted to database/instances.json on the controller, and passed to other hosts when the instance is reassigned. This is because when you set a field from ctl it calls `config.set(field, value, "control")` which only checks if control has write access, which it does. The field is then persisted using JSON.stringify which calls Config.toJSON, which calls Config.toRemote with "controller" as location and because this is local it only checks for write access. The same happens when the file is loaded, fromJSON calls the constructor which calls update which only checks for write access.

Cooldude2606:
What do we consider correct behaviour in this scenario? Is this a bug to follow up in a new PR, or is it accepted behaviour and I remove the comment I added to the description. While it is convenient to have, the same persisted behaviour can be achieved by adding the controller to the credential array, which makes two ways of defining the same interaction. Therefore, I believe that this would be a new bug to consider and the controller should not store these values, only forward them to their correct location.

Hornwitser:
It's not a trivial thing to fix because the way config updates currently work is that a request to set a field calls `config.set(field, value, "control")` so the field has to be stored in memory if it's write only on the controller otherwise the next step which is to call `Controller.instanceConfigUpdated` wouldn't have the value to send to the host. That method calls `config.toRemote("host")` to get all values and sends that to the host.

This is different from how host configs are handled in that the requests to update the config is routed directly to the host. Would probably have to rework how read and write access on configs in handled as well as how instance configs are updated on hosts to make it work like you're suggesting.

Also this kind of unpredictable behaviour is already documented in https://github.com/clusterio/clusterio/blob/master/docs/config-system.md#defing-fields

> Note that instance config fields should have both "controller" and "host" or unexpected behaviour may occur

Cooldude2606:
Noted, this is a known area of UB with a none trivial solution. I will update the access to be controller+host and remove the comment from the description.

Hornwitser (approved):

PR #674, opened 2024-10-28
Rename controller.external_address

Cooldude2606:
Renamed `controller.external_address` to `controller.public_url` to avoid confusion of the value it should contain.
- Changed from `external` to `public` based on the name used for `host.public_address`
- Changed from `address` to `url` based on it expecting a full URL with protocol not just an address.


Hornwitser:
This is a breaking change and should be documented as such. Ideally old configs using `controller.external_address` should keep working, either by some kind of fallback when controller.public_url doesn't exist or a migration being applied to it.

Hornwitser (approved):
The change is no longer breaking so my previous comment no longer applies and this looks good to me.

Issue #676, opened 2024-10-28
Race condition between between server creation and host shutdown


Cooldude2606:
TL;DR There is a breif period where server state is "created" but the processes has not been spawned, this creates a race condition during shutdown of a host.

When a host is started with auto starting instances, they will immediately start their servers; but if the host is also shutdown at the same time, such as during CI tests, then there is a brief period where server state is "created" but the server process is not yet spawned. This occours because of the time needed to write the server config and map gen settings to the disk, a large latency or slow write increases the chance of a crash and an orphan server process.

The code below is where the issues occurs: (within `@host/server.ts::create`)
```ts
this._state = "create";
await this._writeConfigIni();
await this._writeMapSettings(mapGenSettings, mapSettings);
this._server = child_process.spawn( ... )
```

The suggested fix is to make these into synchronous writes to remove the option for shutdown / kill to be called during the transient state. Error handling will need to be performed to return to a valid state if the writes / spawn throw an error.

PR #677, opened 2024-10-29
Use sync file writes and catch errors to reset state.

Cooldude2606:
By using sync file writes it prevents interrupts between the state change and the server process starting. Also adding error catching to reset the state following an error writing to the files. Fixes: #676


Hornwitser (approved):
I haven't tested it, but from reading the code this looks good and should work. While I normally try to avoid using sync, in this case the async solution would be far more complicated and outweighs any benefit from avoiding a small bit of sync fs access on the relatively rare instance starts.

Cooldude2606:
CI fail is from codecov patch, the 6 missed lines are from the error handlers which are more effort to test than the value gained. This can be revisited once there are less high priority items.

PR #678, opened 2024-10-29
Fix host being unable to sync user state to instances after controller connection

Cooldude2606:
Updated the tests for host.handleSyncUserListsEvent to include message validation.
Fixed the sync events missing host as a possible source which would raise an error if a user was banned while the host was disconnected from the controller and the host reconnected while instances were running.


Hornwitser (approved):
Looks good to me.

PR #679, opened 2024-10-31
Fix whitelist remove command with tests

Cooldude2606:
Adds tests for banlist and whitelist commands being send to running instances. Also fixes issue with the whitelist remove command.


Hornwitser (approved):
Tests and lint pass on my system, more tests are always good :).

PR #680, opened 2024-11-01
A number of small changes with minimal impact.

Cooldude2606:
Changes:
- Make silent testing accessible via `pnpm silent-test` similar to fast testing.
- Enable source maps during testing, this includes for spawns processes like host and controller.
- Add the new node LTS version to our test matrix, 18.x will continue to be supported until EoL.
- Add a default autoComplete for credential fields to fix issues with password autofill.
- Added missing items to the change log.

A single PR is justified by the minimal impact of all these changes. All changes have previously been discussed without objection and no new functionally has been added. The change to credential behaviour is not breaking because it was introduced in alpha 19 which has not been published.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        node-version: [18.x, 20.x, 22.x]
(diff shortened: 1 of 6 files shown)


Danielv123 (approved):

Danielv123:
Having silent-test is nice - I always get confused about what error lines are expected test output and which are failing tests.

PR #682, opened 2024-11-11
Refactor users to use lowercase id

Cooldude2606:
Because factorio is case insensitive but clusterio is case sensitive it has caused some issues with ban syncing where a player can be banned using a different case to their factorio username. This refactor makes clusterio case insensitive for usernames. Duplicate users are merged with a backup file created. Existing tokens are still valid because a new `getByName` method is used to minimise the impact on end users. Some areas, such as the webui and control account, have been left using `user.name` because these are purely visual.

### DIFF
--- a/packages/lib/src/data/PlayerStats.ts
+++ b/packages/lib/src/data/PlayerStats.ts
@@ -92,4 +92,19 @@ export default class PlayerStats {
 		}
 		return json;
 	}
+
+	merge(other: PlayerStats) {
+		this.joinCount += other.joinCount;
+		this.onlineTimeMs += other.onlineTimeMs;
+		if (!this.firstJoinAt || (other.firstJoinAt && other.firstJoinAt > this.firstJoinAt)) {
+			this.firstJoinAt = other.firstJoinAt;
+		}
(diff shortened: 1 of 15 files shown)


packages/lib/src/data/PlayerStats.ts
```diff
@@ -92,4 +92,19 @@ export default class PlayerStats {
 		}
 		return json;
 	}
+
+	merge(other: PlayerStats) {
+		this.joinCount += other.joinCount;
+		this.onlineTimeMs += other.onlineTimeMs;
+		if (!this.firstJoinAt || (other.firstJoinAt && other.firstJoinAt > this.firstJoinAt)) {
```
Danielv123, line 99: Why would you take the higehr number for firstJoinAt? Aren't we interested in the lowest number, ie when the player first joined?

Cooldude2606: I had a reason about it being "the most recently updated" but your reasoning is better; I'll update to take the the earlier time.

Danielv123 (approved):

PR #686, opened 2024-11-18
Add 2.0 support for clusterio api and 1st party plugins

Cooldude2606:
Adds a lua compatibility lib to allow plugins to run on v1.1 and v2.0. Also within this PR are a number of changes from Danielv123 and Hornwitser which were made within a private repo. Together with #684 this allows clusterio to work with basic functionality for factorio 2.0. The tests are not yet passing for factorio 2.0 because of the subspace storage mod not having compatibility.

### DIFF
--- a/plugins/inventory_sync/module/gui/dialog_failed_download.lua
+++ b/plugins/inventory_sync/module/gui/dialog_failed_download.lua
@@ -1,7 +1,10 @@
 local ensure_character = require("modules/inventory_sync/ensure_character")
+local get_script_data = require("modules/inventory_sync/get_script_data")
 
 local dialog_failed_download = {}
 
+--- @param player LuaPlayer
+--- @param acquire_response table
 function dialog_failed_download.create(player, acquire_response)
 	if player.gui.screen.dialog_failed_download then
 		player.gui.screen.dialog_failed_download.destroy()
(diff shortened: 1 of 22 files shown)


plugins/inventory_sync/module/gui/dialog_failed_download.lua
```diff
@@ -121,23 +124,24 @@ dialog_failed_download.events = {
 		end
 
 		if event.element.name == "inventory_sync_failed_download_abort" then
-			-- Give the player a charatcer and let them play with that
-			local player = game.get_player(event.player_index)
+			-- Give the player a character and let them play with that
+			local player = assert(game.get_player(event.player_index))
```
Danielv123, line 128: Why the assert here?

Danielv123 (approved):
I see there are a few TODOs in there for new features that aren't supported by serialization and whatnot - I am fine leaving those for later, but if so we should create issues for them so they don't get lost.

plugins/inventory_sync/module/gui/dialog_failed_download.lua
Cooldude2606: LuaLS complaining about nil check; I think we should take a decision on if we want to use `assert` or `--- @cast player -nil` or `--[[ @as LuaPlayer ]]`

PR #694, opened 2024-11-26
Support synchronizing crafting queue in 2.0 with new storage api

### LINKED ISSUE #690: Implement serialization of factorio 2.0 crafting queue
Introduced by #686 in [inventroy_sync/serialize.lua](https://github.com/clusterio/clusterio/blob/05ca4eb25edfc12ebca6479620bcb3cad1572dee/plugins/inventory_sync/module/serialize.lua#L309)

To get all first party plugins into a bare minimum working state for factorio 2.0 the serialization of the crafting queue was not implemented. This should be implemented using the new api while maintaining backwards compatibilty with the `v2_storage_api` / `v2_crafting_queue` flag.

Danielv123:
Resolves #690 

I don't see any reason to mention it working at all as a feature - this is already the state in 1.1 with alpha 18. I did however fix a thing that was added to the 1.1 api after the original implementation

## Changelog

```
### Features
- Remember crafting progress of latest craft in the crafting queue [#694](https://github.com/clusterio/clusterio/pull/694)
```

### DIFF
--- a/plugins/inventory_sync/module/serialize.lua
+++ b/plugins/inventory_sync/module/serialize.lua
@@ -131,11 +131,11 @@ function serialize.serialize_crafting_queue(player)
 	-- Save current items
 	local inventory = player.get_main_inventory()
 	local old_items = inventory.get_contents()
+	local crafting_queue_progress = player.crafting_queue_progress
 
 	-- Cancel old crafts to get the items back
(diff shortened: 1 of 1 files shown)


Cooldude2606 (approved):
LGTM. Couldn't find any errors and the sync worked correctly during local testing.

PR #695, opened 2024-11-26
Create pull_request_template.md

Danielv123:
Describe why and what has been changed by this pullrequest, feel free to also ask for input on decisions you are unsure about etc.
The changelog lines below will be copied directly from merged pull requests to the changelog on new releases. Changes are expected to either link to the related issue or pull request.

## Changelog
```
### Features
- Added doodad to web interface [#999](https://github.com/clusterio/clusterio/pull/999)

### Changes
- ...

### Bugfixes
- ...

### Breaking changes
- ...
```


Cooldude2606 (commented):
Based on the discussion we had about linking issues vs prs, I think we should include a line in the template about using an issue link if available, and a pr link if not. The example url should also use `issue/999` rather than `pull/999` to encourage this behaviour (although I'm pretty sure GitHub redirects anyway)

PR #696, opened 2024-11-26
Feature/synchronize new logistics slots

### LINKED ISSUE #689: Implement serialization of factorio 2.0 personal logistic slots
Introduced by #686 in [inventroy_sync/serialize.lua](https://github.com/clusterio/clusterio/blob/05ca4eb25edfc12ebca6479620bcb3cad1572dee/plugins/inventory_sync/module/serialize.lua#L80)

To get all first party plugins into a bare minimum working state for factorio 2.0 the serialization of personal logistics was not implemented. This should be implemented using the new api while maintaining backwards compatibilty with the `v2_logistic_api` flag.

Danielv123:
Synchronize the new logsistics slots. Supports all settings, filters etc. Will fully sync unnamed groups, named groups will be added but filters won't be modified to avoid conflicting with general logistics on the server you are joining.

Generalized sync of logistics groups across servers will have to fall under a different plugin.

resolves #689 

## Changelog
```
### Features
- Synchronize personal request slot logistics groups [689](https://github.com/clusterio/clusterio/issues/689)
```

### DIFF
--- a/plugins/inventory_sync/module/serialize.lua
+++ b/plugins/inventory_sync/module/serialize.lua
@@ -85,12 +85,12 @@ function serialize.serialize_personal_logistic_slots(player)
 		}
 		for i = 1, logistic_point.sections_count do
 			local section = logistic_point.get_section(i)
-			table.insert(serialized.sections, {
+			serialized.sections[i] = {
 				group = section.group,
 				active = section.active,
 				multiplier = section.multiplier,
 				filters = section.filters,
-			})
+			}
 		end
 		return serialized
 	end
@@ -140,7 +140,10 @@ function serialize.deserialize_personal_logistic_slots(player, serialized)
 			local section = logistic_point.add_section()
 			for i, slot in pairs(serialized) do
 				section.set_slot(i, {
-					value = { name = slot.name },
+					value = {
+						name = slot.name,
+						quality = "normal",
+					},
 					min = slot.min,
 					max = slot.max,
 				})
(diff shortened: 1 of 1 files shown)


plugins/inventory_sync/module/serialize.lua
```diff
@@ -77,7 +77,22 @@ end
...
-		return nil -- TODO implement v2_logistic_api
+		local logistic_point = player.get_requester_point()
+		local serialized = {
+			enabled = logistic_point.enabled,
+			trash_not_requested = logistic_point.trash_not_requested,
+			sections = {},
+		}
+		for i = 1, logistic_point.sections_count do
+			local section = logistic_point.get_section(i)
+			table.insert(serialized.sections, {
```
Cooldude2606, line 88: Can be `serialized.sections[i] = { ... }` rather than `table.insert`

plugins/inventory_sync/module/serialize.lua
```diff
@@ -112,9 +127,46 @@ function serialize.deserialize_personal_logistic_slots(player, serialized)
...
+		for i = logistic_point.sections_count, 1, -1 do
+			logistic_point.remove_section(i)
+		end
+
+		-- If this is an array instead of a table, migrate to v2 format
+		if serialized[1] ~= nil then
+			local section = logistic_point.add_section()
+			for i, slot in pairs(serialized) do
+				section.set_slot(i, {
+					value = { name = slot.name },
```
Cooldude2606, line 143: `quality` must be specified because the default is "any qauality" which will then complain about "Can't specify non zero request with non trivial item filter condition" when `slot.min > 0`

Cooldude2606 (changes requested):
LGTM. There is a small optimisation which can be made however this is still good to merge and worked during my testing. Required change is related to migration from 1.1 format.

PR #697, opened 2024-11-26
Add backwards compatible json serialization capability

### LINKED ISSUE #687: Allow use of `table_to_json` and `json_to_table` during alls stages
Introduced by #686 in [host/modules/clusterio/compat.lua](https://github.com/clusterio/clusterio/blob/05ca4eb25edfc12ebca6479620bcb3cad1572dee/packages/host/modules/clusterio/compat.lua#L149)

Under factorio 2.0 the `helpers.table_to_json` method is available during all lifecycle stages.
Under factorio 1.1 the `game.table_to_json` method is only available during runtime stage.
This limitation is maintained by `compat.table_to_json` because it was not a breaking change.
However, going forward plugins may use these methods during any lifecycle stage and expect them to work in 1.1.
Therefore both `table_to_json` and `json_to_table` should have a backwards compatible implementation independent of `game`.

Danielv123:
resolves #687 using https://github.com/rxi/json.lua

unless there are situations where game is available but not game.table_to_json?

I also find the require paths confusing here, ended up using a relative require but I know they are known to have issues.

### DIFF
--- a/packages/host/modules/clusterio/compat.lua
+++ b/packages/host/modules/clusterio/compat.lua
@@ -2,6 +2,8 @@
 Adds various functions used to ensure compatibly between version including some polyfills
 ]]
 
+local json = require("json")
+
 --- @diagnostic disable: deprecated
 
 --- @class LibCompat
@@ -148,9 +150,14 @@ if major_v1_1 then
 	-- Can not use ... here because of luals type suggestion breaking when it is used
 	-- TODO maybe include a version that does not require game?
 
-	compat.table_to_json = function(tbl) return game.table_to_json(tbl) end
-	compat.json_to_table = function(json) return game.json_to_table(json) end
-	compat.write_file = function(filename, data, append, for_player) return game.write_file(filename, data, append, for_player) end
+	if game ~= nil then
+		compat.table_to_json = function(tbl) return game.table_to_json(tbl) end
+		compat.json_to_table = function(json) return game.json_to_table(json) end
+		compat.write_file = function(filename, data, append, for_player) return game.write_file(filename, data, append, for_player) end
+	else
+		compat.table_to_json = json.encode
+		compat.json_to_table = json.decode
+	end
 elseif major_v2 then
 	compat.table_to_json = helpers.table_to_json
 	compat.json_to_table = helpers.json_to_table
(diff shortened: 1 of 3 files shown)


packages/host/modules/clusterio/compat.lua
```diff
@@ -148,9 +150,14 @@ if major_v1_1 then
...
 	-- TODO maybe include a version that does not require game?
 
-	compat.table_to_json = function(tbl) return game.table_to_json(tbl) end
-	compat.json_to_table = function(json) return game.json_to_table(json) end
-	compat.write_file = function(filename, data, append, for_player) return game.write_file(filename, data, append, for_player) end
+	if game ~= nil then
+		compat.table_to_json = function(tbl) return game.table_to_json(tbl) end
+		compat.json_to_table = function(json) return game.json_to_table(json) end
+		compat.write_file = function(filename, data, append, for_player) return game.write_file(filename, data, append, for_player) end
+	else
```
Cooldude2606, line 157: This block will never be ran because `game` is only present during runtime and the file cannot be required at runtime. Therefore we should either always use `json.encode` and `json.decode` for 1.1 or we move the check inside of the function calls such as:
```lua
compat.table_to_json = function(tbl)
    if game then
        return game.table_to_json(tbl)
    end
    return json.encode(tbl)
end
```

Cooldude2606 (changes requested):
I gave a quick look at the json lib used, mostly at how it encoded and decoded tables, and was happy with it. The relative require path is fine for 2.0; it could cause issues for 1.1 however the file contains no state or side effects so the worst outcome is a slightly worse memory usage from it being loaded multiple times. The required change is to fix the dead code explained below.

Danielv123:
I assume the builtin version performs a lot better, so it would be preferable to use that when available. I am also pretty sure there is no case where game becomes unavailable after it has become unavailable.

Only argument i can see against it would be consistency if there are small differences between the output of the implementations - but I think its sufficiently rare for people to have hard dependencies on the output of the game and post game json serialization that its not important. 99% of json stuff happens in game, as we can already see from json only having been supported via luagamescript.

Cooldude2606 (approved):
You found a nicer way to do it than I thought of. So this looks good to me.

PR #701, opened 2024-11-28
Searchable selection for enums

Cooldude2606:
TL;DR If a config value is of type `"string"` and contains the `enum` property, then it will be displayed as a searchable selection dropdown. Other selection dropdowns have been made searchable.

- It was decided that only string enums would be supported until the need arises for further support.
- The following selections are now searchable: Host assignment, Default mod pack, Default user role, Mod pack version.
- Target instance during save transfer was previously searchable but used a function, this was changed to use `optionFilterProp`.
- User role assignment also used a filter function and was updated to use `optionFilterProp`.

## Changelog
```
### Changes
- String config values with whitelisted values (enum) now display as a selection dropdown. [#701](https://github.com/clusterio/clusterio/pull/701)
- Existing selection dropdowns now support search filtering. [#701](https://github.com/clusterio/clusterio/pull/701)
```

### DIFF
--- a/packages/web_ui/src/components/AssignInstanceModal.tsx
+++ b/packages/web_ui/src/components/AssignInstanceModal.tsx
@@ -77,13 +77,14 @@ export default function AssignInstanceModal(props: AssignInstanceModalProps) {
 			</Paragraph>
 			<Form form={form} initialValues={{ host: props.hostId ?? "null" }}>
 				<Form.Item name="host" label="Host">
-					<Select>
+					<Select showSearch optionFilterProp="name">
 						<Select.Option value={"null"}>
 							<Typography.Text italic>Unassigned</Typography.Text>
 						</Select.Option>
 						{[...hosts.values()].sort((a, b) => strcmp(a.name, b.name)).map((host) => <Select.Option
 							key={host["id"]}
 							value={host["id"]}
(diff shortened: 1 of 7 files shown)


packages/web_ui/src/components/AssignInstanceModal.tsx
```diff
@@ -77,13 +77,14 @@ export default function AssignInstanceModal(props: AssignInstanceModalProps) {
 			</Paragraph>
 			<Form form={form} initialValues={{ host: props.hostId ?? "null" }}>
 				<Form.Item name="host" label="Host">
-					<Select>
+					<Select showSearch optionFilterProp="name">
 						<Select.Option value={"null"}>
```
Danielv123, line 81: Should be mentioned that the <Select.Option> syntax is deprecated in antd - the new API is just an array to <Select /> instead. Not a big deal to update whenever they release v6 though.

Danielv123 (approved):

packages/web_ui/src/components/AssignInstanceModal.tsx
Cooldude2606: Noted, I also expect other things will break when updating to v6 so we can do all changes in a single PR.

PR #708, opened 2024-12-03
Use Get-Process instead of wmic on windows

Danielv123:
Use Get-Process instead of pidusage to get factorio CPU usage on windows. This is required due to wmic being deprecated and removed in new releases of windows 11. There is a pending pull request to pidusage to fix it, but there hasn't been a release in 2 years.

## Changelog
```
### Bugfixes
- Fix statistics exporter not working on newer versions of windows 11 due to wmic deprecation [#707](https://github.com/clusterio/clusterio/issues/707)
```

### DIFF
--- a/packages/host/src/Instance.ts
+++ b/packages/host/src/Instance.ts
@@ -4,6 +4,8 @@ import pidusage from "pidusage";
 import phin from "phin";
 import util from "util";
 import type { Static } from "@sinclair/typebox";
+import { exec } from "child_process";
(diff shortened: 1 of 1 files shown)


Cooldude2606 (approved):
I am on Windows 10 and this continues to work. However, it is under reporting the CPU usage in comparison to task manager, although, I retested master branch and it also under reports by the same margin and therefore this is not an issue that was introduced by this PR and so this is good for merge.

PR #711, opened 2024-12-11
Expand compat to 0.17

Cooldude2606:
Exapands compat lib to include as low as 0.17 because this is the supported version of subspace storage. The only limitation this brings in comparison to support for 1.1 is restricting `active_mods` to be runtime only, bumping minimum to 0.18 would remove this limitation.

This PR also changes how compat lib is used; however this is not a breaking change because the lib did not exist in alpha 18. The main change was the use of a metatable rather than manual function calls. This makes the lib more ergonomic and standardies the method of accessing different properties.

## Changelog
```
### Changes
- Compat lib now supports versions from 0.17 to 2.0 [#711](https://github.com/clusterio/clusterio/pull/711)
```

### DIFF
--- a/packages/create/templates/common/module/control.lua
+++ b/packages/create/templates/common/module/control.lua
@@ -12,9 +12,8 @@ local MyModule = {
 -- If you use any of these events and access globalData then make sure to call this function first
 local script_data = {}
 local function setup_script_data()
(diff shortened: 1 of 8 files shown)


Danielv123 (approved):
I like it, you were right that this way is better

PR #712, opened 2024-12-12
Auto mod update utils

O2theC:
functions and interfaces to help in the updating of mods

idk, i'm bad at this


packages/web_ui/src/components/ModsPage.tsx
```diff
@@ -129,6 +136,403 @@ function CreateModPackButton() {
 	</>;
 }
 
+// Define a basic type for the mod data we expect back from the new backend request
+// This should ideally match the structure returned by the backend.
+// We reuse the ModPortalMod structure name for convenience, but it's fetched via backend.
+interface ModPortalMod {
```
Cooldude2606, line 142: Why is this used rather than `Static<typeof lib.ModPortalDetailsSchema>` ?

packages/web_ui/src/components/ModsPage.tsx
```diff
@@ -129,6 +136,403 @@ function CreateModPackButton() {
...
+		if (!open || !factorioVersion) {
+			setAllMods([]); // Clear mods if modal closed or no version
+			return;
+		}
+
+		setLoading(true);
+		setError(null);
+		let canceled = false;
+
+		// Use the hypothetical new backend request
```
Cooldude2606, line 217: Comment needs to be updated because it is no longer hypothetical.

packages/web_ui/src/components/ModsPage.tsx
```diff
@@ -129,6 +136,403 @@ function CreateModPackButton() {
...
+						{/* Map allowed versions to Select Options */}
+						<Select.Option value="2.0">2.0</Select.Option>
+						<Select.Option value="1.1">1.1</Select.Option>
+						<Select.Option value="1.0">1.0</Select.Option>
+						<Select.Option value="0.18">0.18</Select.Option>
+						<Select.Option value="0.17">0.17</Select.Option>
+						<Select.Option value="0.16">0.16</Select.Option>
+						<Select.Option value="0.15">0.15</Select.Option>
+						<Select.Option value="0.14">0.14</Select.Option>
+						<Select.Option value="0.13">0.13</Select.Option>
```
Cooldude2606, line 385: Could this instead be a map rather than manually typed?

packages/web_ui/src/components/ModsPage.tsx
```diff
@@ -129,6 +136,403 @@ function CreateModPackButton() {
...
+								<Button
+									onClick={() => {
+										handleControllerDownload(
+											record.name,
+											record.title,
+											record.latest_release?.version,
+											factorioVersion
+										);
+									}}
+									disabled={!record.latest_release?.version}
```
Cooldude2606, line 445: `record.latest_release?` can be `record.latest_release!` here because of the check on line 435 `&& record.latest_release`

Cooldude2606 (commented):
No major concerns, only minor comments and suggestions. The tests are very well written only missing cases for ControlConnection however I would consider these out of scope due to the lack of a proper test harness for the class.

packages/lib/src/data/messages_mod.ts
```diff
@@ -197,6 +197,77 @@ export class ModSearchRequest {
...
+	static src = "control" as const;
+	static dst = "controller" as const;
+	static permission = "core.mod.search_portal" as const; // Reuse search permission for now
+
+	// Define allowed Factorio versions
+	static allowedVersions = [
+		Type.Literal("0.13"), Type.Literal("0.14"), Type.Literal("0.15"),
+		Type.Literal("0.16"), Type.Literal("0.17"), Type.Literal("0.18"),
+		Type.Literal("1.0"), Type.Literal("1.1"), Type.Literal("2.0"),
+	] as const;
```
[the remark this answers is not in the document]
Danielv123: Our patching stuff doesn't work with older versions, but I think as a server management tool it still works? Mostly a moot point though, I think it makes sense to just implement what the API allows then you can do what you want. The default selection in the list is 2.0 which will be the only thing used 99% of the time.

packages/controller/src/Controller.ts
```diff
@@ -96,6 +96,13 @@ export default class Controller {
 	clusterLogIndex: lib.LogIndex | null = null;
 	clusterLogBuildInterval: ReturnType<typeof setInterval> | null = null;
 
+	// Cache for mod portal requests
+	static MOD_PORTAL_CACHE_DURATION = 30 * 60 * 1000; // 30 minutes
```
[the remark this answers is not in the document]
Danielv123: We use shouting case for constants in quite a few places, especially the lua code, ex https://github.com/clusterio/subspace_storage/blob/master/src/config.lua - but I think for this I should probably just make it a controller config option instead.

docs/migrating-from-2.0.0-alpha.13.md
```diff
@@ -8,7 +8,7 @@ Please study the entire migration guide carefully before attempting to update an
 If you're using mods you need to create mod pack string for the mod configurations you're currently running.
 You can do this by running the follow RCON command on an instance that is currently running the mod configuration you want to use (you can also run this command in single player and find the string printed in Factorio's [Log file](https://wiki.factorio.com/Log_file)):
 
-    /sc local m={name="pack",description="",factorio_version=game.active_mods.base,mods={},settings={startup={},["runtime-global"]={},["runtime-per-user"]={}}} for n,v in pairs(game.active_mods) do table.insert(m.mods,{name=n,enabled=true,version=v}) end local function set(s,c) for n,v in pairs(s) do m.settings[c][n]=v end end set(settings.startup,"startup") set(settings.global,"runtime-global") if game.player then set(game.player.mod_settings,"runtime-per-user") else set(settings.player,"runtime-per-user") end local e=game.encode_string(game.table_to_json(m)) log(e) rcon.print(e)
```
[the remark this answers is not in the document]
Danielv123: Decided to keep both around, I expect them to be equally rarely used.

Cooldude2606 (approved):

PR #713, opened 2024-12-20
Fix version variables in compat.lua

Cooldude2606:
Fixs incorrect logic used in compat.lua in factorio 2.0 which was not caught by the tests as we currently do not run factorio 2.0 tests. Also removed some dead code.

## Changelog
None needed due to it being a fix for an unreported bug to a feature not present in alpha 18.


PHIDIAS0303 (approved):

Cooldude2606:
Review was provided by the bug reporter.

PR #731, opened 2025-01-12
Handle missing logistic requester point

Danielv123:
## Changelog
```
### Fixes
- Fix crash when leaving a server without a logistic requester point [#731](https://github.com/clusterio/clusterio/pull/731)
```


Cooldude2606 (approved):
I've checked `deserialize_personal_logistic_slots` and already has a check for nil input, so this change looks good to me.

Issue #723, opened 2024-12-29
Playtime rounding


PHIDIAS0303:
The rounding is a bit weird when it is 3 min 60 sec. (Data is 3.99 min)
Maybe there are some better way to display the numbers
![Screenshot_193](https://github.com/user-attachments/assets/4bda5a6b-2c63-4520-87ec-54dfdb4d2250)

PR #732, opened 2025-01-17
Time Format

PHIDIAS0303:
[#723](https://github.com/clusterio/clusterio/issues/723)

### Changes
- use floor instead of round to prevent something like 1 minute and 60 seconds to happen


Cooldude2606 (approved):
I don't see any case where we would want to round up to the nearest second, so I am happy to accept this change. And because it only impacts display text, I do not see this causing unintended side effects.

Issue #719, opened 2024-12-26
Create a default mod pack for base game and space age


Cooldude2606:
A set of default mod packs should be created for each supported version of factorio, currently this means:
- 1.1 Base Game
- 2.0 Base Game
- 2.0 Space Age

Creation of these mod packs should occur during use of the create tool or when a cluster is start with no mod packs present.

PHIDIAS0303:
1.1 and 2.0 can be the same mod pack i think. just differentiate by factorio version.

PHIDIAS0303:
which i think it should be possible to be blank for this stat.

PR #735, opened 2025-02-04
Add default mod packs

Cooldude2606:
Closes: #719 
When a cluster is first created (has no mod packs) a set of default mod packs will be added which includes space age and base game for all valid versions. These values are hardcoded because they are not expected to change often and it has been done in the same location as `getBuiltinMods` to keep all version specific code to a single location. Also added a missing test for datastore as I needed to add `get size` to the class and was in the test file.

## Changelog
```
### Features
- A fresh install of Clusterio will include default mod packs. [#719](https://github.com/clusterio/clusterio/issues/719)
```


Danielv123 (approved):
1 less step to do during setup ❤️

PR #738, opened 2025-02-27
Document release workflow

Danielv123:
Wrote this when doing the last release for subspace


Cooldude2606 (approved):
Should have other docs link to this one, but I can add that when I do the version bump.

Issue #717, opened 2024-12-26
Allow importing / exporting of ban/admin/white lists from / to a json file


Cooldude2606:
Provide an option for an existing json file to be imported into the cluster, as well as the abilty to export the current state of the cluster. This would need to work for ban lists, admin lists, and white lists.

PR #741, opened 2025-03-18
Add bulk user actions (Import/Export/Restore)

Cooldude2606:
Closes: #717

Adds web ui buttons and ctl commands which allow importing, exporting, and restoring user lists from json files. These lists include bans, admins, whitelist, and a combined format specific to clusteiro.  Please note the following behaviours:

* Importing is a strictly additive action, while restoring will match exactlly what is given by the files.
* Both the web ui and the ctl commands will attempt to guess the import type based on file name if it is not specified.
* The web ui supports multiple file uploads which are merged client side before being imported.
* When restoring, the client will save a backup file representing the state which was replaced.
* New bulk action permissions are added, but specific actions require their corresponding permission. e.g. importing bans requires "bulk import" and "set banned" and may require "create user".

The web ui components have been manually tested on the latest version of Firefox.

## Changelog
```
### Features
- Added import, export, and restore of user lists (bans / admins / whitelist / combined). [#717](https://github.com/clusterio/clusterio/issues/717)
```

### DIFF
--- a/packages/lib/src/permissions.ts
+++ b/packages/lib/src/permissions.ts
@@ -495,6 +495,12 @@ definePermission({
 	description: "Bulk export users including admin, whitelist, and bans.",
 });
 definePermission({
+	name: "core.user.bulk_restore",
+	title: "Bulk user restore",
+	description: "Bulk restore users including admin, whitelist, and bans.",
+});
+
+definePermission({
 	name: "core.log.follow",
 	title: "Follow cluster log",
 	description: "Receive new entries in the cluster log.  Required to see instance console.",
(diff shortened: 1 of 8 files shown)


packages/lib/src/permissions.ts
```diff
@@ -484,6 +484,22 @@ definePermission({
 	title: "Delete user",
 	description: "Delete users and all data stored for them.",
 });
+definePermission({
```
Danielv123, line 487: All of these effectively give cluster admin rights, right?

Danielv123 (approved):

packages/lib/src/permissions.ts
Cooldude2606: These permissions allow for the "bulk" request. To actully set/create users you still require the other permissions. The UI is setup to refect this. I will add this to the description of the permissions.

Issue #740, opened 2025-03-17
Download ctl config


Cooldude2606:
Add a button to download a generated ctl config for the currently logged in user, to make it easy to switch over to the command line tool.

PR #743, opened 2025-03-23
Add button to download ctl config

Cooldude2606:
As title says. The new button is under the profile dropdown menu in the top right.
Has dependency on #741 because of reusing the `saveJson` method. Will rebase once that PR is merged.

Closes: #740

## Changelog
```
### Features
- Added button to download ctrl config. [#740](https://github.com/clusterio/clusterio/issues/740)
```


Danielv123 (approved):
Added an icon and tooltip that gives a hint as to what the button does 
image

Cooldude2606:
I would argue that all of them should have icons if one of them does; but that can be done in a seperate PR because it is a very simple change.

PR #747, opened 2025-03-25
Add config cloning on instance creation

Cooldude2606:
As title says. When creating an instance you can select a config to clone from. Webui and CTL both supported.

## Changelog
```
### Features
- Added config cloning on instance creation. [#720](https://github.com/clusterio/clusterio/issues/720)
```

### DIFF
--- a/packages/web_ui/src/components/InstancesPage.tsx
+++ b/packages/web_ui/src/components/InstancesPage.tsx
@@ -1,6 +1,6 @@
 import React, { useContext, useState } from "react";
 import { useNavigate } from "react-router-dom";
-import { Button, Form, Input, Modal, Select } from "antd";
+import { Button, Form, Input, Modal, Select, Tooltip } from "antd";
 
 import * as lib from "@clusterio/lib";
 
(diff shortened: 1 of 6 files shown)


packages/web_ui/src/components/InstancesPage.tsx
```diff
@@ -53,6 +59,14 @@ function CreateInstanceButton() {
...
 					<Input />
 				</Form.Item>
+				<Form.Item name="instanceClone" label="Clone From">
+					<Select
+						defaultValue={-1}
+						options={[{ id: -1, name: "Default Config" }, ...props.instances.values()]
+							.map(i => ({ value: i.id, label: i.name }))
+						}
+					/>
+				</Form.Item>
```
Danielv123, line 69: Some hover text explaining what is cloned would be nice - its not obvious that its only cloning the instance config and not saves for example.

Danielv123 (approved):
Instance clones from the web interface could also benefit from being able to see the instance config that is being cloned, and possibly make modifications so you don't have to do the extra step of opening the new instance. An option to carry over saves also sounds useful. It would probably clutter the UI a bit though, so I am happy with this going in as is.

Cooldude2606:
Cloning saves is a nice idea, I'll add that to the backlog and revist it another time. As for viewing the config which is being cloned, i feel that will create a lot of cluter in the menu, and being able to edit it seams redudant when it automatically navigates to the instance page when you create a new instance, where you can then edit it.

PR #748, opened 2025-03-29
Add config option for console logging

Cooldude2606:
Adds a per instance config option which enables the deadicated console log for a server. This is useful for third party integrations which expect to be able to tail a file rather than communicate over a websocket. The file is always named `console.log` as I see no value in supporting custom file paths and the complications that brings.

## Changelog
```
### Features
- Added `factorio.console_logging` to enable deadicated console logs for a server [#748](https://github.com/clusterio/clusterio/issues/748)
```

### DIFF
--- a/packages/host/src/Instance.ts
+++ b/packages/host/src/Instance.ts
@@ -218,6 +218,7 @@ export default class Instance extends lib.Link {
 			enableWhitelist: this.config.get("factorio.enable_whitelist"),
 			enableAuthserverBans: this.config.get("factorio.enable_authserver_bans"),
 			verboseLogging: this.config.get("factorio.verbose_logging"),
+			consoleLogging: this.config.get("factorio.console_logging"),
 			stripPaths: this.config.get("factorio.strip_paths"),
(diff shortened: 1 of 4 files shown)


Danielv123 (commented):
Why do we default to naming the file console.log (can be confused with the js function) rather than factorio-current.log like the standard log? Doesn't factorio already write to the standard log by default anyways?

Cooldude2606:
Name was selected because the cmd line arg is `--console-log` and so `console.log`. Can be something else if you have a better suggestion, I'm not attached to it.
And while yes clusterio captures the console log (game chat) as part of it's logs, it is not written to a file in the same way that current log is, and this can be expected by third party integrations that are not aware of clusterio. This setting is designed to allow those integrations to work by providing them the game console output.
For example, I have a lot of python and js scripts that tail the file and do processing on it. These are extremely simple scripts which work with any version of factorio. Updating them to work with websockets, or migrating them to clusterio plugins, is much higher effort then passing an extra argument to the factorio process, especially when there is no value gained by it and it locks them to be clusterio only.

Danielv123 (approved):

Issue #754, opened 2025-04-14
Replace phin with fetch


Cooldude2606:
Currently `phin` is used as the http client in code used by the controller and hosts. This dependency has problems, notably with being deprecated, and not supporting streaming of request data. The built-in `fetch` API does not have these problems and could be migrated to, dropping phin as a dependency.

Danielv123:
Looking forward to it

Danielv123:
Apparently still used in a few places https://github.com/search?q=repo%3Aclusterio%2Fclusterio%20phin&type=code

PR #755, opened 2025-04-14
Remove `phin` without introducing new dependencies

Cooldude2606:
Progress towards #754

New Issue: fetch does not support custom agents (required for keep alive and custom CA) which means we will need to add a new dependency as a replacement for phin. Native fetch uses `undici` which would give us access to custom agents if added, or we could consider alternatives such as `axios`. This should be decided before merging this PR.

## Changelog
N/A - Internal Refactor

### DIFF
--- a/packages/create/create.js
+++ b/packages/create/create.js
@@ -6,7 +6,6 @@ const fs = require("fs-extra");
 const inquirer = require("inquirer");
 const os = require("os");
 const path = require("path");
-const phin = require("phin");
 const stream = require("stream");
 const util = require("util");
 const yargs = require("yargs");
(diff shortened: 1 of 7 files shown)


Cooldude2606:
After some discussion, there is only one custom agent property used by production code which is `ca` for self signed certs. Custom certs can be loaded using [NODE_EXTRA_CA_CERTS](https://nodejs.org/api/cli.html#cli_node_extra_ca_certs_file) so we do not require custom agents and can use native fetch with no dependencies. We expect that we will still require a dev dependency to make requests in certain tests which require keep alive or max sockets.

packages/create/create.js
```diff
@@ -624,7 +623,7 @@ async function inquirerMissingArgs(args) {
 			answers.publicAddress = args.publicAddress;
 		} else {
 			try {
-				let result = await phin("https://api.ipify.org/");
+				let result = await fetch("https://api.ipify.org/");
 				myIp = result.body.toString();
```
Hornwitser, line 627: This will cause `myIp` to become `"[object ReadableStream]"`. Need to use `await result.text()` here.

Cooldude2606: Fixed in 47c5e0fabd3c0156d11dbe2400564df4fa2f890a

Issue #718, opened 2024-12-26
Support large user databases upwards of 10k without slow load times


Cooldude2606:
Currently, a large enough user database (~5k) will start to lag the web ui on the users table. This needs to be improved because it is near unusable.

PHIDIAS0303:
should split pages for every 50 100 200 or whatever size

PR #756, opened 2025-04-16
Add user search and pagination

Cooldude2606:
Adds the abilty to search users by name and enables pagination to reduce lag.
Closes: #751 #718 

## Changelog
```
### Features
- Added username search to the users list. [#751](https://github.com/clusterio/clusterio/issues/751)

### Fixes
- Reduce page load time on users list by enabling pagination. [#718](https://github.com/clusterio/clusterio/issues/718)
```


Danielv123 (approved):
I like it. An obvious next step would also be allowing to filter by roles but that doesn't need to be included here

Issue #752, opened 2025-04-10
Add a safe (re)start option to start a controller / host without any plugins enabled


Cooldude2606:
As the title says, add a start and/or restart option which starts a controller / host with all plugins disabled. This can be automatic following an unexpected crash.

PR #757, opened 2025-04-27
Add recovery startup mode

Cooldude2606:
Adds a recovery mode to controller and host startup which limits their behaviour. The intention of this feature is to be proactive and provide a base from which we can expand should difficulties arise in the future; therefore, the scope of disabled behaviours is very limited at the moment.

**Considerations:**
- Although possible, I have not provided a way for ctl or the web ui to initiate a recovery state as I foresee this being used from the physical machine if the process fails to start normally. We can revist this if we find a need for it.
- I am not confident in my abilty to modify the existing "restart on failure" code, and therefore I have not linked it to recovery mode despite there being value in doing so. We may also want this as a config option to control this behaviour.
- Our current integration tests assume there is a singleton controller and host instance running, this makes it non-trival to test a feature which requires the controller to be restarted. Therefore, tests have not been written.

**Recovery Mode:**
- Controller / Host / Instance - Disable plugin loading
- Controller - Disable autosaving of data
- Controller - Reject incoming host connections
- Host - Disable auto start of instances

Closes: #752 

## Changelog
```
### Features
- Added a recovery mode to host and controller accessible as an option on the run command. [#752](https://github.com/clusterio/clusterio/issues/752)
```

### DIFF
--- a/packages/controller/src/WsServer.ts
+++ b/packages/controller/src/WsServer.ts
@@ -239,6 +239,13 @@ ${err.stack}`
 	}
 
 	async registerHost(data: lib.RegisterHostData, socket: WebSocket, req: IncomingMessage) {
+		if (this.controller.recoveryMode) {
+			logger.warn(`Recovery | rejected host connection attempt for ${this.remoteAddr(req)}`);
+			wsRejectedConnectionsCounter.inc();
+			socket.close(lib.ConnectionClosed.TryAgainLater, "Connection Refused: Controller in recovery mode");
+			return;
+		}
+
 		try {
 			let tokenPayload = jwt.verify(
 				data.token,
--- a/packages/controller/src/WsServer.ts
+++ b/packages/controller/src/WsServer.ts
@@ -242,7 +242,7 @@ ${err.stack}`
 		if (this.controller.recoveryMode) {
 			logger.warn(`Recovery | rejected host connection attempt for ${this.remoteAddr(req)}`);
 			wsRejectedConnectionsCounter.inc();
-			socket.close(lib.ConnectionClosed.TryAgainLater, "Connection Refused: Controller in recovery mode");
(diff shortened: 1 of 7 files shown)


packages/controller/src/WsServer.ts
```diff
@@ -239,6 +239,13 @@ ${err.stack}`
 	}
 
 	async registerHost(data: lib.RegisterHostData, socket: WebSocket, req: IncomingMessage) {
+		if (this.controller.recoveryMode) {
+			logger.warn(`Recovery | rejected host connection attempt for ${this.remoteAddr(req)}`);
+			wsRejectedConnectionsCounter.inc();
+			socket.close(lib.ConnectionClosed.TryAgainLater, "Connection Refused: Controller in recovery mode");
```
Danielv123, line 245: Does TryAgainLater make sense if the host is in recovery mode? Trying again later won't produce a different result, unless the controller config is changed to not be in recovery mode and relaunched. It doesn't seem like a temporary thing to me.

I'd expect to see this response for rate limiting, intermittent network errors or the controller rejecting requests due to high load.

Cooldude2606: I selected this code because recovery mode should be temporary (it is pretty useless putting the controller/host in recovery mode for an extended period). I like to think of it as a service becoming temporary unavailable for maintence.

This status was the closest fit without adding our own code; but there is no extra logic added on the client socket, so there is nothing preventing using our own code. So unless you have a suggestion for another then I will add `RecoveryMode = 4002` and use that.

Danielv123: Recovery code makes a lot more sense to me

Danielv123 (approved):
Not sure if 1013 should still be there if it's not being used anymore but otherwise I think it's ready to merge

Cooldude2606:
Although unused, it is part of the [websocket spec](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code) so there is no harm in it being in the enum.

PR #758, opened 2025-04-27
Add indicators for host version mismatch and required restart

Cooldude2606:
Adds an indicator to the restart button on the controller and host pages when a "requires restart" config value has been modified. This indicator also appears when a new version of clusterio has been installed but is not the active version. The host list shows this indicator next to the status and also shows an indicator when there is a version mismatch with the controller.

## Changelog
```
### Features
- Added pending update detection for host and controller. [#758](https://github.com/clusterio/clusterio/issues/758)
- Added restart required indicator for host and controller. [#758](https://github.com/clusterio/clusterio/issues/758)
- Added version mismatch indicator in host list. [#758](https://github.com/clusterio/clusterio/issues/758)
```

### DIFF
--- a/packages/lib/src/config/classes.ts
+++ b/packages/lib/src/config/classes.ts
@@ -203,6 +203,8 @@ export class Config<
 
 	/** Set to true when a field in the config is changed. */
 	dirty = false;
+	/** Set to true when a 'restart required' field in the config is changed. */
+	restartRequired = false;
 
 	/**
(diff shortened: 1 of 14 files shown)


packages/lib/src/config/classes.ts
Danielv123, line 1: This cleanup was nice

Danielv123 (approved):

PR #766, opened 2025-05-19
Add missing migrations for system info and tests

Cooldude2606:
Adds the missing migrations which should have been added for `canRestart` and `restartRequired`. Also adds the tests which were missing for these and existing migrations, and fixes 3 broken tests related to remote updates.


Danielv123:
I got one more possibly migration related error when checking out this pr 
image

Danielv123 (approved):

PR #765, opened 2025-05-16
Impliment decode for map exchange strings v2

Cooldude2606:
As title, enables clusterio to decode v1 and v2 map exchange strings. This also improves the testing for exchange string decode which resulted in a previously unknown bug being fixed.

Closes: #700 
Closes: #722

## Changelog
```
### Features
- Added support for v2 map exchange strings. [#700](https://github.com/clusterio/clusterio/issues/700)

### Fixes
- Map gen settings using the incorrect name for cliff settings following exchange string decode.
```

PR #767, opened 2025-05-20
Bump tests: Remove Node 18 & Add Factorio 2.0

Cooldude2606:
Node v18 is now end of life so we no longer offically support it; so we can stop testing it and bump our recomended ts config. At the same time the merge of #765 means that all tests are passing in factorio 2.0 and can be safely added to the test suite. Also fixed some flakey tests while I was checking results.

Draft until #766 is merged which implements fixes for broken tests.

## Changelog
```
### Meta
- Node 18 has reached end of life and is no longer maintained, we no longer guarantee support for it.
```

### DIFF
--- a/tsconfig.node.json
+++ b/tsconfig.node.json
@@ -2,7 +2,7 @@
 	"extends": "./tsconfig.base.json",
 	"compilerOptions": {
 		"lib": ["es2023"],
-		"module": "node16",
+		"module": "nodenext",
 		"moduleResolution": "node16",
 		"target": "es2022",
 		"forceConsistentCasingInFileNames": true,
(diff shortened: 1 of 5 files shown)


tsconfig.node.json
```diff
@@ -2,7 +2,7 @@
 	"extends": "./tsconfig.base.json",
 	"compilerOptions": {
 		"lib": ["es2023"],
-		"module": "node16",
+		"module": "nodenext",
 		"moduleResolution": "node16",
```
Danielv123, line 6: Could maybe also be nodenext, from the docs it seems to resolve to the same thing

Danielv123 (approved):

Issue #770, opened 2025-05-28
Add event maps to event emitters


Cooldude2606:
Previously we had used TypedEventEmitter to provide typing information on events. This was implemented for one event emitter in the code base. Since then, there has been an update to the types for `node:events` which allows an EventMap to been given to provide typing information for events without the need to refactor. We should add these maps to all event emitters.

PR #772, opened 2025-06-01
Add types to event emitters and clean up connections

Cooldude2606:
Adds event maps for all event emiters using base node types. This means the custom typed event emitter has been removed and replaced with base maps. This also led me to investigate the events being emitted by connections and clean up some areas of it. There are no changes in the logic only reducing repetition and validating that asserts were still applicable.

Closes: #770 

## Changelog
None, this is a refactor with no external changes.


Cooldude2606:
Low patch coverage is because of the changes in `ctl/commands.ts` being part of an interactive command which can not be tested in our current suite.

PR #781, opened 2025-07-21
Minimap plugin

Danielv123:
Adds a mapping plugin, drawing a representation of the mapview of instances in the web UI with support for timelapses.
It has been a priority to support history data while keeping the runtime performance of the instance and controller and storage consumption on the controller as low as possible, making some tradeoffs for web client rendering performance.

It heavily relies on custom binary formats to achieve the storage footprint desired.

Ideally it will be good enough to be left enabled by default without having to think about it to later be able to use it to review history for anti griefing purposes or making a neat timelapse of a multiplayer map.

One thing that is unique about the timelapsing feature as implemented here is that we only store changes, so a replay would skip over periods when players are taking a break without additional editing effort.

My intention is for this to be able to be used as a base for the web interface and other plugins. Some examples:

- The instance page could show a small window of the main factory. this would make it easier to visually tell instances apart
- Gridworlds should render multiple instances spread out on one canvas so it looks like one
- Universal edges should render its edges on the map as another overlay layer - not sure how to integrate that cleanly

Todo:
- [x] Tracking players
- [ ] More testing with multiple surfaces
- [ ] Testing performance with a longer history (ideally a full game playthrough)
- [x] Clean up user interface
- [x] Button to delete history, for example if you are switching savefile on an instance or the history gets too large
- [ ] More granular subscriptions - only send updates for clients viewing the right instance/surface/force

## Changelog
```
### Features
- Added minimap to web interface. [#999](https://github.com/clusterio/clusterio/issues/999)
```


Danielv123:
For some reason previous_recipe.name.name is correct. Doesn't look like it makes sense but thats how it works

plugins/minimap/controller.ts
```diff
@@ -0,0 +1,1077 @@
...
+						buffer.writeUInt8((x_int24 >> 8) & 0xFF, offset + 1);
+						buffer.writeUInt8(x_int24 & 0xFF, offset + 2);
+						offset += 3;
+
+						// y_tiles (3 bytes): Y coordinate as 24-bit signed integer
+						const y_int24 = Math.round(position.y) & 0xFFFFFF;
+						buffer.writeUInt8((y_int24 >> 16) & 0xFF, offset);
+						buffer.writeUInt8((y_int24 >> 8) & 0xFF, offset + 1);
+						buffer.writeUInt8(y_int24 & 0xFF, offset + 2);
+						offset += 3;
```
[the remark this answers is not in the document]
Danielv123: No, its a 24 bit signed int as per the docs.

plugins/minimap/docs/tile_data_storage.md
```diff
@@ -0,0 +1,118 @@
...
+- **Data**: 32×32×2 bytes (2048 bytes uncompressed)
+- **Compression**: Deflate (zlib)
+
+#### Pixel Changes Record (Type 2)  
+Incremental updates for pixels sharing the same color transition.
+
+| Bytes | Field               | Description                 |
+| ----- | ------------------- | --------------------------- |
+| 4     | `tick` (uint32)     | ⌊tick / 60⌋ timestamp       |
+| 2     | `count` (uint16)    | Number of pixels changed    |
```
Cooldude2606: I can't help but feel that this doesn't need to be 2 bytes. When can we expect more than a quater of a chunk (255 tiles) to be updated with the same transition? If needs be we could have a "Large" pixel change record as an extra type for some easy savings.

plugins/minimap/module/minimap.lua
```diff
@@ -0,0 +1,576 @@
...
+			local should_update = false
+			if has_moved_significantly(last_position and last_position.position or nil, position) then
+				should_update = true
+			elseif last_position and (current_sec - last_position.last_update_sec) >= 5 then
+				-- Send update every 5 seconds even if not moving (for timeout)
+				should_update = true
+			elseif not last_position then
+				-- First position for this player
+				should_update = true
+			end
```
Cooldude2606: `last_position` is non-nil following the call to `has_moved_significantly`. No code error, just redudant checks and an unreachable block.

plugins/minimap/module/minimap.lua
```diff
@@ -0,0 +1,576 @@
...
+	if not player or not player.valid then
+		return
+	end
+
+	-- Clear any existing position data for this player to ensure fresh tracking
+	if storage.minimap.player_positions then
+		storage.minimap.player_positions[player.index] = nil
+	end
+
+	-- The next position update will be sent when the player moves or in process_player_positions
```
Cooldude2606: How come there is no session start? If you add it on the first player pos update then you could be waiting upto 5 seconds before start session is recorded.

plugins/minimap/module/minimap.lua
```diff
@@ -0,0 +1,576 @@
...
+							start_tick = game.tick,
+							end_tick = nil,
+							recipe = recipe.name,
+							icon = (function()
+								local rp = prototypes.recipe[recipe.name]
+								if rp and rp.main_product then
+									return { type = rp.main_product.type, name = rp.main_product.name }
+								end
+								return nil
+							end)()
```
Cooldude2606: A closure like this isnt great when it would otherwise be two lines of code above `local data`

plugins/minimap/package.json
```diff
@@ -0,0 +1,57 @@
+{
+	"name": "@clusterio/plugin-minimap",
+	"version": "2.0.0-alpha.1",
```
Cooldude2606: Version must match current version on main, currently `22b`

plugins/minimap/package.json
```diff
@@ -0,0 +1,57 @@
...
+	},
+	"bugs": {
+		"url": "https://github.com/clusterio/clusterio/issues"
+	},
+	"engines": {
+		"node": ">=18"
+	},
+	"peerDependencies": {
+		"@clusterio/lib": "workspace:*"
+	},
```
Cooldude2606: Following the merge of #823 the peer deps here must include controller, webui, and host to have correct types. Similarly need to be added as dev deps, and existing deps checked for if `catalog:` should be used.

plugins/minimap/index.ts
```diff
@@ -0,0 +1,52 @@
...
+	ClearMinimapDataRequest,
+} from "./messages";
+
+// Define permissions
+lib.definePermission({
+	name: "minimap.view",
+	title: "View Minimap",
+	description: "View the interactive minimap of Factorio instances.",
+	grantByDefault: true,
+});
```
Cooldude2606: Do you think we should split this into "view live" and "view historic" ?

plugins/minimap/web/dataSources/SingleInstanceDataSource.ts
```diff
@@ -0,0 +1,191 @@
...
+				this.plugin.setInstanceSurfaceFilter(null, null);
+			}
+			return;
+		}
+		const filters = [{ instanceId: this.instanceId, surface: this.surface }];
+		if (this.plugin.setInstanceSurfaceFilters) {
+			this.plugin.setInstanceSurfaceFilters(filters);
+		} else if (this.plugin.setInstanceSurfaceFilter) {
+			this.plugin.setInstanceSurfaceFilter(this.instanceId, this.surface);
+		}
```
Cooldude2606: Why do both of these exist?

plugins/minimap/instance.ts
```diff
@@ -0,0 +1,249 @@
...
+			this.logger.error(`Failed to initialize minimap on instance: ${err}`);
+		}
+	}
+
+	async onOutput(output: lib.ParsedFactorioOutput) {
+		// Handle any special output parsing if needed for tile data
+		if (output.type === "action" && output.action === "CUSTOM") {
+			// Could handle custom factorio output here if the Lua module sends special messages
+		}
+	}
```
Cooldude2606: Can be removed. Although what where your plans here?

plugins/minimap/controller.ts
```diff
@@ -0,0 +1,1587 @@
...
+		this.nextPlayerIds.clear();
+		this.activePlayerSessions.clear();
+
+		this.logger.info("Cleared all minimap data");
+	}
+
+	async onInstanceStatusChanged(instance: InstanceInfo) {
+		if (instance.status === "running") {
+			return;
+		}
```
Cooldude2606: What happens here if a host loses connection? The instance status becomes "unknown" and it injects end session records while the player is still active on the server. This means player position events can come after a session end once the host reconnects.

plugins/minimap/controller.ts
```diff
@@ -0,0 +1,1587 @@
...
+
+		try {
+			const savePromises = [];
+			for (const [tileName, chunks] of queue) {
+				const promise = (async () => {
+					const tilePath = path.join(this.tilesPath, tileName);
+					let existingTile = Buffer.alloc(0);
+					if (await fs.pathExists(tilePath)) {
+						existingTile = await fs.readFile(tilePath);
+					}
```
Cooldude2606: This should either use eixsts sync, or try catch.

Cooldude2606 (commented):
This was a *quick* skim taking a few hours; some little picks in here but also some wider discussions to be had on this. There is at least one thing which would block me from mergering, namely the risk of path traversal on the controller.

packages/lib/src/prometheus.ts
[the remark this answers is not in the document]
Danielv123: I did this when it caused my host to not start after I botched some things for 2.1 support. I think I will revert it and figure it out when we get there.

plugins/minimap/web/CanvasMinimapPage.tsx
```diff
@@ -0,0 +1,2584 @@
...
+
+		// Update display zoom if zoom changed (throttled)
+		if (updates.zoomLevel !== undefined) {
+			if (throttledZoomUpdate.current) {
+				clearTimeout(throttledZoomUpdate.current);
+			}
+			throttledZoomUpdate.current = setTimeout(() => {
+				setDisplayZoom(viewStateRef.current.zoomLevel);
+			}, 100); // Update display every 100ms max
+		}
```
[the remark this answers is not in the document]
Danielv123: Good catch — this was a debounce; updated naming/comment accordingly (57b50048).

Danielv123:
**Task 6 Verification - Package.json Alignment**

✅ **VERIFIED: Already Aligned**

Evidence:
- Version: ✅ `2.0.0-alpha.22b` matches current main
- PeerDependencies: ✅ Contains @clusterio/controller, @clusterio/web_ui, @clusterio/host, @clusterio/lib with `workspace:*`
- DevDependencies: ✅ Contains all four clusterio packages with `workspace:*`  
- Catalog usage: ✅ Consistent `catalog:` for @types/*, @sinclair/typebox, fs-extra, antd, react, webpack

No code changes needed. PR #823 merge integration is complete.

plugins/minimap/package.json
Danielv123: ✅ Confirmed: plugins/minimap/package.json version is `2.0.0-alpha.22b`, matching current main branch baseline.

plugins/minimap/docs/player_position_tracking.md
```diff
@@ -0,0 +1,111 @@
...
+- **Storage**: 3 bytes per coordinate (x_tiles, y_tiles) (24 bits)
+
+### Binary Record Format (append-only)
+
+| Bytes                | Field                | Description                                    |
+| -------------------- | -------------------- | ---------------------------------------------- |
+| 1                    | `type`               | `0=Position`, `1=SessionStart`, `2=SessionEnd` |
+| **Position (0)**     |                      |                                                |
+| 4                    | `t_sec` (uint32)     | Seconds since epoch                            |
+| 4                    | `sec` (uint32)       | Seconds since map started (ticks / 60⌋)        |
```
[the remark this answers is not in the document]
Danielv123: Acknowledged. The space savings potential is significant (160→88→20 bytes/5s). This is an interesting optimization approach for a potential follow-up PR. The current design prioritizes simplicity and flexibility for the initial implementation, but the TimeDrift record type is worth reconsidering as the feature matures.

plugins/minimap/docs/tile_data_storage.md
Danielv123: Good observation about the practical limits. A "Large" pixel change record variant could be a viable optimization if we encounter cases where >255 pixel transitions occur frequently. This would be worth revisiting in a follow-up if telemetry shows it is needed.

plugins/minimap/instance.ts
```diff
@@ -0,0 +1,249 @@
...
+export class InstancePlugin extends BaseInstancePlugin {
+	pendingTileUpdates = new Map<string, Set<[number, number, [number, number, number, number]]>>();
+
+	async init() {
+		if (!this.instance.config.get("factorio.enable_script_commands")) {
+			throw new Error("minimap plugin requires script commands to be enabled.");
+		}
+
+		// Listen for tile data from the Lua module
+		this.instance.server.on("ipc-minimap:tile_data", (data: TileDataIpc) => {
```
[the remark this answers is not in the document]
Danielv123: Thanks for the note. The `.handle` API is acknowledged.

plugins/minimap/docs/player_position_tracking.md
```diff
@@ -0,0 +1,111 @@
...
+| 3                    | `y_tiles` (int24)    | Y coordinate in tiles                          |
+| 2                    | `player_id` (uint16) | Player dictionary ID                           |
+| **SessionStart (1)** |                      |                                                |
+| 4                    | `t_ms` (uint32)      | Session start time                             |
+| 2                    | `player_id` (uint16) | Player dictionary ID                           |
+| 1                    | `name_len` (uint8)   | UTF-8 byte length of player name               |
+| n                    | `name` bytes         | Player name string                             |
+| **SessionEnd (2)**   |                      |                                                |
+| 4                    | `t_ms` (uint32)      | Session end time                               |
+| 2                    | `player_id` (uint16) | Player dictionary ID                           |
```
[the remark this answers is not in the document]
Cooldude2606: I do not see any code on the lua side to detect surface changes, neither do I see any code on the instance or controller side for writing `SessionEnd`, although there is code for writing `SessionsStart`. The root cause here is that lua never emits `SessionEnd` for `on_player_changed_surface` which means the player remains in the `active_sessions` map for a surface if they move to a different surface and then disconnect.

Issue #451, opened 2022-01-25
Disabling servers from startup


ILLISIS:
There is no way to disable a server from starting with the 'start all' button.

PR #783, opened 2025-07-21
Exclude instance from "start all" function

Danielv123:
This PR implements the feature requested in [#451](https://github.com/clusterio/clusterio/issues/451), allowing users to exclude specific instances from being started by the "Start all" button in the web interface. Previously, the "Start all" button would attempt to start all stopped instances. This change introduces a new configuration field `instance.exclude_from_start_all` which, when set to `true`, will cause the "Start all" operation to skip that instance. The field is exposed in the instance configuration UI and documented.

```
## Changelog
### Features
- Added ability to exclude instances from the 'Start all' web UI button. [#451](https://github.com/clusterio/clusterio/issues/451)
```

---

**Open Background Agent:** 
[Web](https://www.cursor.com/agents?id=bc-bc30acf1-4c3e-4bd2-928a-efc910587faa) · [Cursor](https://cursor.com/background-agent?bcId=bc-bc30acf1-4c3e-4bd2-928a-efc910587faa)

Learn more about [Background Agents](https://docs.cursor.com/background-agent/web-and-mobile)


Cooldude2606 (commented):
Code looks good, just needs a test added.

Cooldude2606 (approved):

Issue #797, opened 2025-09-14
Have an automatic workflow to publish to NPM when a new release is made


Cooldude2606:
The publish workflow is manual and prone to human error. We can automate this.

PR #798, opened 2025-09-14
Add publish workflow

Cooldude2606:
Adds a new publish workflow which uploads the npm packages and factorio mods on release. If the version number is not incremented then no package / mod is published.

Before merge the following must be confirmed:
- [x] `FACTORIO_TOKEN` added to github workflow secrets with permission `ModPortal: Upload Mods`
- [x] `NPM_TOKEN` added to github workflow secrets with publish permission for `@clusterio`

Closes: #797

## Changelog
No external changes.

### DIFF
--- a/publish.js
+++ b/publish.js
@@ -89,7 +89,7 @@ async function publishMods(dir) {
 		const [file, name] = mod;
 		const releases = await requestModReleases(name);
 		if (!releases.some(release => release.file_name === file)) {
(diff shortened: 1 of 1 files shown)


Danielv123:
image

image


Added the token. Apparently npm now forces a max expiry of 90 days on tokens - hower one of my old tokens still has the expiry set in 2042 so not sure whats up with that.

PR #815, opened 2026-01-07
Add extra startup warnings to CLI: User count and Lock file

Cooldude2606:
1) When the bootstrap create admin command is run on the controller, it tries to change the user database file. If the controller is already running, this change is not saved. This is now fixed by using a lock file for the controller configuration. If a valid lock file is present, the bootstrap create admin, config set, and run commands will not start.

2) A user might try to run the create admin bootstrap command if they forgot to create the first admin user for their cluster. To stop this, the controller now checks how many users are loaded from the database file. If no users are found, the controller will refuse to start and will suggest using the create admin command. This check can be bypassed using a hidden command option.

## Changelog
```
### Features
- Added config lock files to prevent process duplication. [#815](https://github.com/clusterio/clusterio/issues/815)

### Changes
- The controller will now refuse to start if no users were loaded from file. [#814](https://github.com/clusterio/clusterio/issues/814)
```

### DIFF
--- a/packages/controller/controller.ts
+++ b/packages/controller/controller.ts
@@ -255,6 +255,7 @@ async function initialize(): Promise<InitializeParameters> {
 			default: "plugin-list.json",
 			type: "string",
 		})
+		.option("bypass-lock-file", { hidden: true, type: "boolean", nargs: 0, default: false })
 		.command("plugin", "Manage available plugins", lib.pluginCommand)
 		.command("config", "Manage Controller config", lib.configCommand)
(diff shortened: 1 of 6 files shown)


Danielv123:
Can I suggest a --force or something to allow stealing the lockfile? Its not uncommon for something to happen and a lockfile to get stuck after termination, ex power loss

Cooldude2606:
That particular case is already covered by the stale lock checks (read the file and checking the pid is an active process).

Although I will still add a bypass as there may still be other cases in which it could be useful.

PR #834, opened 2026-02-08
Fix mod dependency parser breaking with inconsistent spaces

Cooldude2606:
Following discussion in the Discord support channel, we found cases where mod dependency specifications used inconsistent spacing. This could prevent automatic downloads and stop mods from working on a host.

These cases are now handled by using the same regular expression as the game. Error handling has also been improved so that a failure in one dependency no longer prevents other dependencies from being resolved.

## Changelog
```
### Fixes
- Fixed mod dependency parser breaking with inconsistent spaces. [#834](https://github.com/clusterio/clusterio/issues/834)
```

### DIFF
--- a/packages/controller/src/ControlConnection.ts
+++ b/packages/controller/src/ControlConnection.ts
@@ -614,22 +614,21 @@ export default class ControlConnection extends BaseConnection {
 		return modInfos;
 	}
 
+	// eslint-disable-next-line complexity
 	async handleModDependencyResolveRequest(request: lib.ModDependencyResolveRequest) {
 		const {factorioVersion, checkForUpdates} = request;
 		const searchQueue = [...request.mods];
 		const localInfos = [...this._controller.modStore.mods()];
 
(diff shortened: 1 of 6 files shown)


packages/controller/src/ControlConnection.ts
```diff
@@ -708,14 +708,21 @@ export default class ControlConnection extends BaseConnection {
...
+								sha1: release.sha1,
+								version: release.version,
+								name: modReleases.name,
+								author: modReleases.owner,
+								title: modReleases.title,
+							});
+						} catch {
+							// Can error with invalid dependencies, in which case we highlight it as invalid
+							// We modify the name to expose the error to the user.
+							invalid.add(`${modReleases.name} (Has invalid dependency)`);
```
Danielv123: Here we are adding the mod release name + some text, but on line 744 it seems like the field assumes we should pass just the name (at least it does so for `incompatible`) - is this intended?

PR #838, opened 2026-02-25
Extended export data

solarcloud7:
### Features
- Extended export-data to produce separate spritesheets for recipes, signals, technologies, planets, qualities, entities, and static UI icons.
- Added Icon Reference page to web interface at `/icon-reference` for browsing all exported icons.

### Changes
- Refactored `useItemMetadata()` into a generic `useSpriteMetadata(category)` with shared caching.
- Export completion now logs per-category icon counts.

### Fixes
- Fixed bad mod paths in icon specifications crashing the export (now logs a warning and skips).
- Fixed prototypes with no icon data causing errors during export.

###Test plan:
1.  kick off a data export
- should now have completion message and error on individual icon will give error instead of crash

2.  Navigate to: 
http://localhost:8080/icon-reference
- All the icons are searchable with 0 duplicates (duplicated calculated by relative asset path not name)

3. New files in /static/ folder can be inspected by attached static.zip.

[static.zip](https://github.com/user-attachments/files/25554863/static.zip)

### DIFF
--- a/plugins/inventory_sync/module/inventory_sync.lua
+++ b/plugins/inventory_sync/module/inventory_sync.lua
@@ -328,9 +328,8 @@ inventory_sync.events[defines.events.on_player_joined_game] = function(event)
 	local player = assert(game.get_player(event.player_index))
 	local script_data = get_script_data()
 
-	-- It's possible Factorio doesn't invoke the on_player_created event when loading a save in single player
-	if not script_data.players[player.name] then
-		create_player(player, false)
+	if not game.is_multiplayer() then
+		return player.print("Inventory sync deactivated: game is single player")
 	end
 
 	-- Send acquire request even if an active download is currently in progress


packages/host/src/export.ts
```diff
@@ -142,6 +143,9 @@ async function loadIcon(
 	iconMipmaps: number,
 	iconCache: IconCache,
 ) {
+	if (typeof iconPath !== "string") {
+		return null;
+	}
```
Cooldude2606: I'm not sure this path can ever be hit. `iconPath` comes from `item.icon` or `layer.icon` both of which come from the export of a `Prototype` directly from lua, which has the engine check the value types. If it was not a string then the mod would fail to load and the export would never run.

packages/host/src/export.ts
```diff
@@ -514,12 +602,181 @@ export async function exportData(server: FactorioServer) {
...
+		// was already exported by a prior category (items, recipes, signals,
+		// etc.), and skip internal prefixed entities (dummy-, hidden-).
+		if (cat.name === "entity") {
+			items = items.filter(item => {
+				if (item.name.startsWith("dummy-") || item.name.startsWith("hidden-")) {
+					return false;
+				}
+				const iconPath = item.icons
+					? (item as LayeredIconSpecification).icons[0]?.icon
+					: (item as SimpleIconSpecification).icon;
```
Cooldude2606: This appears a few times, maybe should be pulled into a helper function `getPrimaryIconPath` ?

packages/host/src/export.ts
```diff
@@ -514,12 +602,181 @@ export async function exportData(server: FactorioServer) {
...
+ * @param alreadyExported - Set of __mod__/path strings already packed.
+ * @returns Spritesheet and metadata (may be empty if nothing found).
+ * @internal
+ */
+async function exportStaticIcons(
+	server: FactorioServer,
+	modVersions: Map<string, string>,
+	alreadyExported: Set<string>,
+) {
+	const builtinMods = ["core", "base", "space-age", "elevated-rails", "quality"];
```
Cooldude2606: Why not: `lib.ModPack.getBuiltinModNames(factorioVersion)`?
This would avoid hard coded values and allow for easier maintainence.

packages/lib/src/data/ExportManifest.ts
```diff
@@ -11,21 +11,44 @@ export default class ExportManifest {
...
+		 * Keys for static-metadata use the pattern `{prefix}-{basename}` for directory scans
+		 * (e.g. `tooltip-category-electricity`, `alert-no-fuel`) or a short name for
+		 * individual files (e.g. `add`, `cancel`, `clock`, `rename`).
 		 */
-		public assets: Record<string, string>
+		public assets: Record<string, string>,
+		/** Name of the mod pack used for this export. */
+		public modPackName?: string,
+		/** Name of the instance that triggered this export. */
+		public instanceName?: string,
```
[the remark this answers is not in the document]
solarcloud7: Your right about both modpacks and instances.    I originally added that data in case there were multiple instances with different mod packs on a host.  This could help in debugging at the time an export is completed, but in the grand scheme of things,  once mods or modpacks change, showing that info becomes not valuable at best.    


I'll remove Mod Pack, Instance, and the Icon's since the filter total is good enough.
image

packages/host/src/export.ts
```diff
@@ -280,19 +283,86 @@ const itemTypes = new Set([
...
+	["core", "graphics/import.png", "import"],
+	["core", "graphics/no-recipe.png", "no-recipe"],
+	["core", "graphics/questionmark.png", "questionmark"],
+	["core", "graphics/rename-icon.png", "rename"],
+	["core", "graphics/too-far.png", "too-far"],
+	["core", "graphics/multiplayer-waiting-icon.png", "multiplayer-waiting"],
+	["core", "graphics/player-force-icon.png", "player-force"],
+	["core", "graphics/green-circle.png", "green-circle"],
+	["core", "graphics/green-dot.png", "green-dot"],
+];
```
Hornwitser: This hardcoded list is incorrect as most of these are referenced by existing prototypes. In this list only favourite.png and player-force-icon.png is not referenced by any prototype. However favourite.png is just the factorio icon, and player-force-icon.png is an unused asset. Adding a whole static icon export system for only these two icons is of dubious value.

packages/controller/src/routes.ts
```diff
@@ -240,9 +240,23 @@ async function uploadExport(req: Request, res: Response) {
...
+		"export/technology-spritesheet.png",
+		"export/technology-metadata.json",
+		"export/planet-spritesheet.png",
+		"export/planet-metadata.json",
+		"export/quality-spritesheet.png",
+		"export/quality-metadata.json",
+		"export/entity-spritesheet.png",
+		"export/entity-metadata.json",
+		"export/static-spritesheet.png",
+		"export/static-metadata.json",
```
Hornwitser: Why are we adding a whole bunch of new spritesheets here where they are all 32x32 pixel icons? Shouldn't the item-spritesheet be renamed to icon-spritesheet and then all of the icons put into there?

packages/host/src/export.ts
solarcloud7: Your 100% right again here also.  I see loadFile() also uses that: `const builtinModNames = ["core", ...lib.ModPack.getBuiltinModNames(server.version!)];` so this makes

packages/host/src/export.ts
```diff
@@ -280,19 +283,86 @@ const itemTypes = new Set([
...
+
+// Non-prototype icon directories to scan for static UI assets.
+// Only include directories with small, reasonably square icons suitable for compositing.
+// Keys: directory name relative to graphics/icons/ → prefix used in metadata keys
+// Excluded: arrows (120x64), shapes (120x64), shortcut-toolbar (36x24/84x56 duplicates),
+//           category (doesn't exist in any builtin mod)
+// Note: alerts live in core/graphics/icons/alerts/, tooltips split across base/ and space-age/
+const staticIconDirs: Record<string, string> = {
+	"alerts":   "alert",
+};
```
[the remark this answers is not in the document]
solarcloud7: I added more info in the comment to be clearer:


// Non-prototype icon directories to scan across builtin mods (core + DLC) for static UI assets.

packages/host/src/export.ts
solarcloud7: This is true.   I shouldn't have added this dead code.

Hornwitser (changes requested):
The whole static-spritesheet makes no sense. Out of the 50 icons the code finds and puts in the static sheet there are only 5 that aren't part of the `utility-sprites` and `burner-usage` prototypes which are
- `__core__/graphics/icons/alerts/endangered-by-lightning-red.png`
- `__core__/graphics/icons/alerts/item-to-be-delivered-symbol.png`
- `__core__/graphics/icons/alerts/pipeline-disabled-icon.png`
- `__core__/graphics/favourite.png`
- `__core__/graphics/player-force-icon.png`

These are all unused assets of the game, no code references them, so we shouldn't expose them. It seems likely they could be deleted in a future cleanup.

One thing to note is that the way the icons are exposed in the web ui is not correct. It will not work with multiple mod packs, and we shouldn't assume web ui code only wants to have icons from the default mod pack. That was done that way as a holdover from the time before subspace storage was its own mod, and should be refactored. This doesn't need to be fixed in this PR, but as a note for the future the item_metadata.ts file is likely going to be rewritten.

packages/controller/src/routes.ts
solarcloud7: I'll remove these all to ensure we only have icons with a matching prototype in the game.

Hornwitser:
Looks like https://github.com/clusterio/clusterio/blob/5bea930d0560a6863234786b59edd4659209cbbc/test/integration/clusterio.js#L655 will have to be updated with the new content of the manifest as that test is failing.

Issue #855, opened 2026-03-03
A warning should be issued if player_online_auto_save_slots < settings.autosave_slots


Cooldude2606:
If the instance setting `factorio.player_online_autosave_slots` is less than `factorio.settings.autosave_slots` then when the server saves the game and players are online the the number of save slots becomes the player_online_autosave_slots number instead of the higher autosave_slots number. This is probably an error in the config, and should trigger a warning/error presented to the user.

PR #856, opened 2026-03-04
Add extended config validation

Cooldude2606:
This adds the ability to define configuration validators beyond just the type. Currently, validators have been added for values that must be integers, greater than zero, or valid file paths. Additionally, two special cases have been added for `factorio.version` and `factorio.player_online_autosave_slots`.

The controller will prevent values from being set if they fail validation. The web UI will also show the validation error before applying changes, providing immediate feedback.

A known limitation is that the validation order is undetermined, which could lead to unexpected interactions if there are many dependencies between configuration values. At present, only `factorio.player_online_autosave_slots` has a dependency, and it is simple. A more robust solution may be needed if this becomes an issue.

There is scope to add more validators in the future, but these are left for later implementation.

## Changelog
```
### Features
- Added config value validation beyond simple type checks. #856
- Added config staging to allow setting many values at once. #856
- Added a warning when attempting to set player autosave slots lower than autosave slots. #855
- Added `controller.authSecret` as the type of config field `controller.auth_secret` was inconsistent. #856
```

Closes: #855

### DIFF
--- a/packages/lib/src/config/classes.ts
+++ b/packages/lib/src/config/classes.ts
@@ -230,6 +230,8 @@ export class Config<
 	private fields: Fields;
 	private _unknownFields: Record<string, FieldValue> = {};
 
+	/** Set to true during validators to prevent edits */
+	private _readonly = false; // TODO replace with "using config.asReadonly()"
 	/** Set to true when a field in the config is changed. */
 	private dirty = false;
 	/** Set to true when a 'restart required' field in the config is changed. */
@@ -274,6 +276,7 @@ export class Config<
 	private _staging?: { config: Config<Fields>, changes: Set<string>};
 
 	/** Get the latest version of staging */
+	// TODO replace with "using config.startStating()"
 	get staging(): Config<Fields> {
 		if (this._staging) { return this._staging.config; }
 		const config = new this.constructor(this.location, this.fields);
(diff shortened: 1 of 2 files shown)


packages/web_ui/src/components/BaseConfigTree.tsx
```diff
@@ -391,6 +391,18 @@ function computeTreeData(
...
+						validator: function(_, fieldValue) {
+							if (def.extendedValidation) {
+								try {
+									def.extendedValidation(fieldValue, config);
+								} catch (err: any) {
+									return Promise.reject(err);
+								}
+							}
+							return Promise.resolve();
+						},
```
Hornwitser: Use an async arrow function instead of manually dealing with Promises:
```suggestion
						validator: async (_, fieldValue) => {
							if (def.extendedValidation) {
								def.extendedValidation(fieldValue, config);
							}
						},
```

packages/lib/src/config/classes.ts
```diff
@@ -513,12 +530,13 @@ export class Config<
...
 		this._checkAccess(name, def, remote, ConfigAccess.write, true);
 		if (remote !== this.location) {
 			this._checkAccess(name, def, this.location, ConfigAccess.write, true);
 		}
 
-		// Empty strings are treates as null
+		// Empty strings are treat as null
 		if (value === "") {
 			value = null;
 		}
```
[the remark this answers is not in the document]
Cooldude2606: https://github.com/clusterio/clusterio/pull/856/commits/014b18caf87f57f13065f282e79b5df9300cbd39

Cooldude2606:
Code should now be stable; however, all missed coverage looks reachable and so I will aim for 100% before marking this PR as ready. Early reviews welcome.

packages/lib/src/config/definitions.ts
```diff
@@ -151,30 +169,35 @@ export class ControllerConfig extends classes.Config<ControllerConfigFields> {
...
 			type: "number",
 			initialValue: 8,
+			validator: validators.greaterThanZero,
 		},
 		"controller.system_metrics_interval": {
 			title: "System Metrics Interval",
 			description: "Interval in seconds to collect and update system metrics for the Web UI",
 			type: "number",
 			initialValue: 10,
+			validator: validators.greaterThanZero,
```
Hornwitser: This value can be set to 0 to disable it.

PR #858, opened 2026-03-04
Upload test logs for debugging CI failures

Hornwitser:
The CI often fails with cryptic errors that would be much easier to debug if the logs from Factorio was provided.  Upload the cluster log that was written during the tests which includes Factorio logs so that these can be inspected after a test failure.

## Changelog
None. Internal change.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -40,6 +40,12 @@ jobs:
       with:
         files: ./coverage/lcov.info
         token: ${{ secrets.CODECOV_TOKEN }}
+    - name: Upload test logs for debugging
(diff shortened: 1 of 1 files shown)


Cooldude2606 (commented):
Any reason why we would want the logs of successful tests? i.e. why not `if: ${{ failure() }}`

Hornwitser:
Lots of reasons.
- Test succeeds but prints something you didn't expect.
- Successful output of a previous run to compare against when the test fail.
- Ability to inspect if the test did what you expected it to do in the CI.
- Ability to compare logs for different versions in the same run if only one version fails.
- Ability to check that the log uploading code works without having to make a failing CI run.

PR #882, opened 2026-03-21
Remove fs-extra and klaw

### LINKED ISSUE #860: Remove fs-extra
The use of fs-extra was primarily motivated by the lack of a promises based APIs, a lack of recursive filesystem operations, and a neat way to output files creating parent directories if needed.

Unfortunately fs-extra seems to be more or less abandoned and have knows bugs with APIs we use such as [createWriteStream](https://github.com/isaacs/node-graceful-fs/issues/248) and [createReadStream](https://github.com/isaacs/node-graceful-fs/issues/152).

Fortunately most of the things we need now exists in the node fs api with things such as the promises API, the recursive option to mkdir and readdir, and the use of fs-extra outputFile function was replaced long ago due to its lack of fsync making it unsafe.

Thus we should replace all usage of fs-extra with the promises API from fs (falling back the the callback API for things not supported by the promises API).

Hornwitser:
These packages are obsoleted by the promises API to the built-in Node fs module.  The fs-extra package is also causing problems with broken APIs that don't behave the same as Node's fs module, and throwing errors without code properties on them.

Resolves #860

### Changelog
```
### Changes
- Removed usage of fs-extra. #860

### Fixes
- Fixed headless download option in installer failing with `TypeError: callback is not a function`.
- Fixed controller crashing if a host attempted to download a mod that had been deleted from the controller's filesystem while the controller was running.
```


packages/create/create.js
```diff
@@ -97,11 +97,24 @@ class LineSplitter extends stream.Transform {
 
 class InstallError extends Error { }
 
+async function pathExists(filePath) {
```
Cooldude2606: This is only really used inside the migration code; so maybe its ok. But this can run into errors if other processes are working with the files and dirs as it may be created or removed. I believe the node docs say as much. But yeah maybe not something to care about inside of create migration code.

packages/ctl/src/commands_host.ts
```diff
@@ -87,7 +87,7 @@ hostCommands.add(new lib.Command({
 		} else {
 			logger.info(`Writing ${args.output}`);
 			try {
-				await fs.outputFile(args.output, content, { flag: "wx" });
+				await fs.writeFile(args.output, content, { flag: "wx" });
```
Cooldude2606: I beleive this requires the same `if (dir) mkdir` checks that other outFile file replacements have included.

packages/host/src/Host.ts
```diff
@@ -931,12 +931,21 @@ export default class Host extends lib.Link {
...
 		try {
-			await fs.move(
+			// Check if save with old name exists first
+			await fs.access(path.join(instance.path, "saves", oldName));
+			// Check if save with new name doesn't exist, this creates the file
+			await fs.writeFile(path.join(instance.path, "saves", newName), "", { flag: "wx" });
+			// Move save to new name overwriting the file that was created for it, this
+			// might leave an empty file at the new name if the save was (re)moved between
+			// the previous two calls and this one.
+			await fs.rename(
```
Cooldude2606: To remove the race condition, could we instead `link(src, dst)` then `unlink(src)` as link will error if the dst already exists. It also avoids writting a file.

packages/lib/src/load_plugin_list.ts
```diff
@@ -174,12 +174,14 @@ export async function loadPluginList(pluginListPath: string, findLocal = true):
...
-	}
 
 	// Read and parse package.json
-	const packageJson = JSON.parse(await fs.readFile(packageJsonPath, { encoding: "utf8" }));
+	let packageJson;
+	try {
+		packageJson = JSON.parse(await fs.readFile(packageJsonPath, { encoding: "utf8" }));
+	} catch (err: any) {
+		return false;
+	}
```
Cooldude2606: This silences the error for bad json content. Was that intended?

test/integration/server.js
```diff
@@ -44,12 +44,10 @@ describe("Integration of host/src/server", function() {
 
 		before(async function() {
 			// Delete result from previous run of these tests
-			if (await fs.pathExists(writePath)) {
-				await fs.remove(writePath);
-			}
+			await fs.rm(writePath, { force: true, recursive: true, maxRetries: 10 });
```
Cooldude2606: Do we know weather `rm` errors on the path not existing?

Cooldude2606 (changes requested):

packages/controller/src/BaseConnection.ts
```diff
@@ -97,7 +97,7 @@ export default class BaseConnection extends lib.Link {
 
 		let stream = await routes.createProxyStream(this._controller.app);
 		stream.filename = mod.filename;
-		stream.source = fs.createReadStream(modPath);
+		stream.source = (await fs.open(modPath)).createReadStream();
```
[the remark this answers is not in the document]
Hornwitser: There is a crucial difference between `fs.createReadStream` and `fs.promises.open(...).createReadStream`. The latter throws if opening the file fails while the former raises an error on the stream.

Which makes me realise this logic is slightly broken, it needs to open the file first so the error is thrown before the proxy stream is created.

packages/create/create.js
Hornwitser: Fixing bad fs interactions is out of scope for this PR. It's about removing fs-extra with minimal changes. And the migration and install code cannot function if other programs are writing to the files at the same time as it is running, so this usage is not a concern here.

packages/ctl/src/commands_host.ts
Hornwitser: I considered this, but most command line tools do not create folders for you if your file path is to a folder that doesn't exist. This is especially annoying if you typo the path.

packages/host/src/Host.ts
Hornwitser: That's a good idea. I'll try that.

packages/create/create.js
Cooldude2606: Fair enough.

Cooldude2606 (approved):

PR #894, opened 2026-04-12
Fix accessing main inventory when in map view

### LINKED ISSUE #892: Inventory sync issue
60362.014 Script @level/modules/inventory_sync/inventory_sync.lua:387: ERROR: Serializing player ILLISIS failed: level/modules/inventory_sync/serialize.lua:207: attempt to index local 'inventory' (a nil value)

Cooldude2606:
When in remote view, using `get_main_inventory` no longer proxies to the character and so a manual index into character is required. The absence of this caused issues with crafting queue serialisation which depended on using the player inventory to deconstruct the crafting queue.

### Changelog
```
### Fixes
- Incorrect acccessing of player inventory during crafting queue serialisation when in remote view. #892
- Use physical location and controller for syncing player to prevent teleporting with remote view. #894
- Fixed player exiting space platform hub on sync and other remote view edge cases.  #894
```

Fixes #892

### DIFF
--- a/plugins/inventory_sync/module/inventory_sync.lua
+++ b/plugins/inventory_sync/module/inventory_sync.lua
@@ -16,6 +16,8 @@ local get_script_data = require("modules/inventory_sync/get_script_data")
 local progress_dialog = require("modules/inventory_sync/gui/progress_dialog")
 local dialog_failed_download = require("modules/inventory_sync/gui/dialog_failed_download")
 
+local v2_remote_controller = compat.version_ge("2.0.0")
+
 -- Returns true if the player is currently in a cutscene
 local function is_in_cutscene(player)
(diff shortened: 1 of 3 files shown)


Hornwitser (changes requested):
Still broken. To reproduce queue up a bunch of crafts on the player, open the map screen, and then kick yourself from the game in the console.
Image

ILLISIS:
I have tested this, the players character is removed after connecting.

Cooldude2606:
This is now a larger piece of work which will require the physical controller type to be synced and restored if in remote view. Not supporting remote view is not an option because of space platforms, which also need work as it currently spawns your character on the platform.

Cooldude2606:
Spoke with @ILLISIS and confirmed that no further issues were found.

PR #895, opened 2026-04-12
Change config errors to be subclasses of RequestError

### LINKED ISSUE #888: Exception rasied during valid config validation appears in controller log
Config validation uses exceptions to signal that validation has failed and the config change should be rejected. This is an expected control flow and should not result in any log output. It should be investigated which log statement the error is propagating to and prevent it reaching that point. Example log below, taken from review on #885 

```
[error] Unexpected error responding to InstanceConfigSetFieldRequest:
Error: Failed validation of dependent statistics_exporter.load_plugin: Plugin statistics_exporter requires script commands
    at InstanceConfig.set (packages\lib\dist\node\src\config\classes.js:534:31)
    at ControlConnection.handleInstanceConfigSetFieldRequest (packages\controller\dist\node\src\ControlConnection.js:323:25)
    at ControlConnection._processRequest (packages\lib\dist\node\src\link\link.js:293:24)
    at ControlConnection._processMessage (packages\lib\dist\node\src\link\link.js:170:18)
    at WsServerConnector.<anonymous> (packages\lib\dist\node\src\link\link.js:74:22)
    at WsServerConnector.emit (node:events:508:28)
    at WsServerConnector._processMessage (packages\lib\dist\node\src\link\connectors.js:256:18)
    at WebSocket.<anonymous> (packages\controller\dist\node\src\WsServerConnector.js:160:22)
    at WebSocket.emit (node:events:520:35)
    at Receiver.receiverOnMessage (node_modules\.pnpm\ws@8.19.0\node_modules\ws\lib\websocket.js:1225:20)
```

Cooldude2606:
This prevents them from being logged when raised during an event handler. The alternative would be to have try catch blocks on all config related requests and to rethrow the error as a request error. But that would be combersome for little value (if any) in comparison to having them inherit from request error.

### Changelog
```
### Changes
- Config validation errors are no longer logged when raised during an request handler. #888
```

Fixes #888

### DIFF
--- a/packages/lib/src/config/classes.ts
+++ b/packages/lib/src/config/classes.ts
@@ -9,6 +9,7 @@ import * as libSchema from "../schema";
 import { StringEnum } from "../data/composites";
 import { safeOutputFile } from "../file_ops";
 import * as validators from "./validators";
+import { RequestError } from "../errors";
 
 const ConfigLocation = StringEnum(["controller", "host", "control"]);
 export type ConfigLocation = Static<typeof ConfigLocation>;
@@ -21,7 +22,7 @@ export type ConfigLocation = Static<typeof ConfigLocation>;
  *
  * @extends Error
  */
-export class InvalidAccess extends Error { };
+export class InvalidAccess extends RequestError { };
 
 /**
  * Invalid Value exception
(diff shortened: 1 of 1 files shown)


Hornwitser (approved):
This is an okay solution for now. In the future, renaming "RequestError" into something that better reflects its more general role as an expected error that does not require a stack traceback logged might be something to look into.

PR #896, opened 2026-04-12
Add double interrupt logic to ctl to force exit

Cooldude2606:
Add logic to ctl which mimics the double catching of sigint which the controller and host impliment. Namely, when a second sigint is caught it will immediately exit.

### Changelog
```
### Features
- Added double interrupt logic to ctl to force immediate exit. #887
```

Fixes #887


Hornwitser (approved):
Although perfectly servicable, for consistency it might be worthwile to implement it using the same wording and variable names as the same logic in packages/controller/controller.ts:449 and packages/host/host.ts:242.

If that wording is sub-par, then update the wording on all of them.

Cooldude2606:
As discussed on discord, the desire to have same wording and variable names prompted me to standise other asspects of the handler and move it into lib. This had the unexpected side effect of removing set blocking as a dependency in ctl and controller.

packages/lib/src/helpers.ts
```diff
@@ -51,6 +54,32 @@ export async function timeout<T>(promise: Promise<T>, limitMs: number, timeoutRe
...
+			logger.fatal(`Caught second ${signalName}, terminating immediately`);
+			process.exit(1);
+		}
+
+		called = true;
+		logger.info(`Caught ${signalName}, shutting down`);
+		callback().catch(err => {
+			setBlocking(true);
+			logger.error(err.stack);
+			logger.fatal("Error during shutdown, terminating immediately");
```
Hornwitser: The traceback should also be logged as fatal as it's the cause of the immediate termination. I see it wasn't done so before, but if we're going to refactor this anyway we might as well do it properly.

Hornwitser (approved):
Looks good appart from the one log level issue I've pointed out below. Can be merged if that is fixed.

packages/lib/src/helpers.ts
Cooldude2606: https://github.com/clusterio/clusterio/pull/896/commits/6f5b8ce68de0561de0b6527c2581877d8a7b70b8

PR #897, opened 2026-04-12
Allow custom versions to be selected as a target

Cooldude2606:
Add a custom version input field that allows users to enter any version number. This approach was chosen as a workaround because TreeSelect does not support custom values. I find that it is acceptable for this to add a couple of extra clicks for users selecting versions not listed on the Wube website, as this is a less common use case.

An autocomplete input was considered, but it made it difficult to distinguish between partial and full versions. It may be possible to modify the search behaviour so the entered value always appears as a selectable option, but this would require updating the tree data and trigger re-renders. Another idea was to simulate this with a button instead of an tree option, but that introduces additional complexity around validating the format. For now, this approach was not pursued, but it can be revisited later if needed.

<p>

</p>

### Changelog
```
### Features
- Allow selection of factorio versions not listed on the Wube website for instances and mod packs. #867
```

Fixes #867


Hornwitser (approved):
It has some usability problems. You can't tab/keyboard navigate to this field when you enter data into the config, you can't accept changes with enter in the field, you have to retype the whole version every time you want to change the custom version, it doesn't accept the "latest" string. And it's starting to be a lot of duplicated code in InputTargetVersion.tsx and InputPartialVersion.tsx.

Still better than not being able to input it all so in that sense it's better than what was before and "approved". I would like it to function better though.

Cooldude2606:
Your UX concerns should not be sorted. The tab key was the hardest of the bunch because antd prefers to use arrow keys for dropdowns, but does not let you insert custom components into the selection order for arrow keys.

I also took the time to dedupliate the code, I orignaly did it because the types weren't playing nicely. My solution to this was to revert to using "string" which isnt perfect but best I could come up with.

Danielv123 (approved):

Danielv123:
I am happy with this - a possible future improvement would be populating the versions list on the controller from the hosts when the API is unavailable.

PR #902, opened 2026-04-19
Improved RoleViewPage

Cooldude2606:
tl;dr Improved role permission page with grouped layout, search, and clearer change tracking.

This update refactors the role page to make permission management much nicer to work with. Permissions are now displayed in grouped sections based on their name (e.g. `core.instance.save.delete` -> `Core / Instance / Save`), with alphabetical ordering within groups. There is a search bar for quick filtering as the permission list has grown quite long. The permission descriptions are now always visible in a subdued style instead of being hidden behind tooltips.

A floating action bar has been introduced; this bar appears only when there are pending changes and provides buttons for “Apply” and “Revert”. We may want to extend this kind of behaviour to other pages.

Permissions now include visual indicators:
- Default permissions are marked with a star icon
- Dangerous permissions are marked with a warning icon
- Modified permissions are visually highlighted
- 
image

### Changelog
```
### Changes
- Updated role permission editor to use grouped layout with search and improved formatting. #902
```

### DIFF
--- a/packages/web_ui/src/components/RoleViewPage.tsx
+++ b/packages/web_ui/src/components/RoleViewPage.tsx
@@ -109,7 +109,7 @@ function PermissionGroup({
 	}
 
 	return (
-		<div style={{ border: "1px solid #f0f0f0", padding: 12, borderRadius: 6 }}>
+		<div style={{ border: "1px solid #424242", padding: 12, borderRadius: 6 }}>
 			<Space style={{ width: "100%", justifyContent: "space-between" }}>
 				<strong>{formatGroupTitle(groupName)}</strong>
 				<Space>
--- a/packages/web_ui/src/components/RoleViewPage.tsx
+++ b/packages/web_ui/src/components/RoleViewPage.tsx
@@ -113,8 +113,8 @@ function PermissionGroup({
 			<Space style={{ width: "100%", justifyContent: "space-between" }}>
 				<strong>{formatGroupTitle(groupName)}</strong>
 				<Space>
-					<Button size="small" onClick={() => setGroup(true)}>Select all</Button>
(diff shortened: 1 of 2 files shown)


packages/web_ui/src/components/RoleViewPage.tsx
```diff
@@ -13,20 +13,165 @@ import PageLayout from "./PageLayout";
...
+			return next;
+		});
+	}
+
+	return (
+		<div style={{ border: "1px solid #424242", padding: 12, borderRadius: 6 }}>
+			<Space style={{ width: "100%", justifyContent: "space-between" }}>
+				<strong>{formatGroupTitle(groupName)}</strong>
+				<Space>
+					<Button size="small" onClick={() => setGroup(true)}>Select all</Button>
```
Danielv123: Why aren't these disabled by canUpdate?

Danielv123: Otherwise I am happy with merging everything else

Danielv123:
Also, why is CI failing? Just flaky stuff?

packages/web_ui/src/components/RoleViewPage.tsx
Cooldude2606: Just an oversight, I will have these disabled with canUpdate then merge.

Cooldude2606:
> Also, why is CI failing? Just flaky stuff?

Yeah just flaking stuff, this PR does not modify anything in the tests or any code covered by tests. You can see how they passed on the previous commit on this PR despite only the board colour being changed.

packages/web_ui/src/components/RoleViewPage.tsx
Cooldude2606: f515b2db80b80de5409a50e0b732812b90530d2e

PR #909, opened 2026-04-25
Multiple small changes

### LINKED ISSUE #859: Instance kill button does not work
If you click on the right side of the kill instance button in the web UI then nothing happens. It only responds with a popup if you click on the "Kill" text.

Image

**Edit:** This also applies for the Host Stop button.

Cooldude2606:
All are self explanatory in the change log below. However, I should note that the use of caret versions can cause issues if an admin updates one package but not the others. As such, I will create a new feature request to have an "Update all" button on the webui.

### Changelog
```
### Changes
- Player auth now uses a scroll pane for listing instance names. #909

### Fixes
- Online user table on instance page now correctly defaults to online only filter. #909
- Fixed popup not opening when pressing kill within the instance control button. #859

### Meta
- Version dependencies now use `^` which enables updates to be applied one package at a time. #803 
- Updated change log creation to include new lines and user contributions. #909
```
Fixes #859

### DIFF
--- a/packages/web_ui/src/components/UsersTable.tsx
+++ b/packages/web_ui/src/components/UsersTable.tsx
@@ -199,7 +199,7 @@ export default function UsersTable({ instanceId, onlyOnline = false, pagination,
 			title: "Last seen",
 			key: "lastSeen",
 			filterMultiple: false,
-			defaultFilteredValue: onlyOnline ? ["online"] : undefined,
+			defaultFilteredValue: onlyOnline ? "online" : undefined,
 			filters: [
 				{ text: "Online", value: "online" },
 				{ text: "24h", value: "24h" },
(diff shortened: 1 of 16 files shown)


packages/web_ui/src/components/UsersTable.tsx
```diff
@@ -199,7 +199,7 @@ export default function UsersTable({ instanceId, onlyOnline = false, pagination,
 			title: "Last seen",
 			key: "lastSeen",
 			filterMultiple: false,
-			defaultFilteredValue: onlyOnline ? ["online"] : undefined,
```
Danielv123, line 202: According to https://ant.design/components/table defaultFilteredValue is a string[] - why did you change it?

Otherwise I am good to merge

Cooldude2606: Had a second look, the issue was an antd bug, so I have reverted this change.

PR #911, opened 2026-04-27
Fix localised info in mod settings is ignored by web UI

### LINKED ISSUE #768: Localisation is ignored in mod settings
For example, mod setting should display by its localised name, 
`{'phi-cl.combine', {'mod-setting-name-prefix.PHI-MB'}, {'mod-setting-name.PHI-MB'}}`

Image

Image

Image

Image

Image

gui-alb:
Should close #768 

### Changelog
```
### Features
- None.

### Changes
- exportData now creates locale.json files for other languages (previously it only created one in English)
- The WebUI now detects the browser language and uses it to load the corresponding locale-<language>.json

### Fixes
- Fixed an issue where the WebUI ignored localized information from mods

### Breaking Changes
- None
```

WebUI when browser is in Japanese:

image


packages/web_ui/src/components/ModPackViewPage.tsx
```diff
@@ -28,6 +28,7 @@ import SectionHeader from "./SectionHeader";
 import ModDetails from "./ModDetails";
 import { InputPartialVersion } from "./InputVersion";
 import Link from "./Link";
+import * as Console from "node:console";
```
Cooldude2606, line 31: Unused import?

packages/controller/src/routes.ts
```diff
@@ -226,29 +226,30 @@ async function uploadExport(req: Request, res: Response) {
...
+		let { name, ext } = path.posix.parse(filePath);
+		const allowedFiles = [
+			"settings.json",
+			"prototypes.json",
+			"locale.json",
+			"defines.json",
+			"spritesheet.png",
+			"metadata.json",
+		];
+		if (!allowedFiles.includes(`${name}${ext}`) && !name.startsWith("locale-")) {
```
Cooldude2606, line 245: This feels much more open than it was previously and than it needs to be. If I understand it correctly you are only needing to expand one dir deeper to allow multiple locales. But this allows any depth and also free traversal so long as the name matches.

Although I may be misunderstanding the attack vector, so if you could explain why this solution is sufficient then ill accept that also.

Cooldude2606 (commented):
I havent had the opportunity to test yet, but I do have two initial comments below.

PHIDIAS0303:
not related: looking at your photo make me see a locale issue, i have the same word twice xD

PHIDIAS0303:
However the actual locale look like this, it is a combined locale
Screenshot_208

Hornwitser:
I appreciate that you want to contribute and that you went to the effort of making a Pull request to try and solve an existing issue. Unfortunately there are big problems with this PR. The biggest is that it does not solve the issue it claims to solve.

Issue #768 is about the `localised_name` property in [Mod settings prototypes](https://wiki.factorio.com/Tutorial:Mod_settings#Creation) being ignored. This PR does not touch the the `localised_name` property and so doesn't fix the issue.

To implement rendering of `localised_name` you will have to implement the correct handling of the [LocalisedString](https://lua-api.factorio.com/latest/types/LocalisedString.html) type from Factorio. This is somewhat involved as it supports [quite a few options with parameters](https://wiki.factorio.com/Tutorial:Localisation#Localising_with_parameters). To get a nice looking result this will also need to deal with at least the commonly used parts of Factorio's [Rich text syntax](https://wiki.factorio.com/Rich_text).

You have implemented support for other languages in the mod settings of the admin panel. We have only used and supported English up to now for the simple practical reason that supporting multiple languages is a lot of extra work and it hasn't been needed. I fear that adding multi language support is going to lead to a lot of issues, feature requests, support burden and so on that we simply don't have the capacity to handle.

For instance there is no option to select the language used in this implementation inside of the admin interface. We are almost certainly going to get users who get the wrong language showing up on the page, and then it's our problem to figure out how to get it to show the right language.

Danielv123:
Agree om this not closing the issue, but I don't mind the contribution itself. Only thing I'd change is as Horn mentioned, add a dropdown or something where we can change language, because relying on the locale of the host computer is super annoying a lot of the time (at least it is for me)

PR #913, opened 2026-05-29
Refactor subscription handling and add new filter classes

Cooldude2606:
This change introduces several refactors to the subscriptions library aimed at improving performance, usability, and separation of concerns. The original `EventSubscriber` class has been split into three distinct classes: one for handling events, one for managing a single value, and one for maintaining a map of values. This separates subscription management from data handling responsibilities.

In addition, filter state has been introduced to `EventSubscriber`, along with a new subscription action that allows existing filters to be replaced. Together, these changes support common filtered subscription patterns without requiring additional boilerplate code.

### Changelog
```
### Features
- Added filter state to `EventSubscriber` allowing plugings to more easilly filter events. #913
- Added `replace` subscription action allowing filter state to be fully replaced with a single request. #913
- Added `ValueSubscriber` and `MapSubscriber` as subclasses of `EventSubscriber`. #913

### Changes
- `subscriptionUpdate` now fires for all subscription requests rather than just added subscriptions. #913

### Fixes
- Fixed permissions list not including `grantByDefault` in the json representation. #913
- Fixed subscription resources being `synced = false` if the remote data was empty. #913

### Breaking Changes
- Renamed `lib.Address.index` to `requestIndex` and added `addressIndex`. #913
- `EventSubscriber` no longer includes a values map, use `MapSubscriber` instead. #913
```


packages/lib/src/subscriptions.ts
```diff
@@ -1,13 +1,142 @@
...
+		if (this._filters.size === 0) {
+			return "[SubscriptionFilters Empty]";
+		}
+		return `[SubscriptionFilters Set<${this._filters.size}>]`;
+	}
+
+	/** Returns true if this filter accepts all the given filters */
+	accepts(value: string) {
+		return this._all || this._filters.has(value);
+	}
```
Hornwitser: Is the description supposed to say "if this filter accepts the given filter" here? The type signature only allows one string to be passed, not multiple.

Hornwitser (approved):
Very nice. I only have one question related to a doc comment, other than that this is good to merge.

packages/lib/src/subscriptions.ts
Cooldude2606: This is left over from when I was considering having an array of strings or another filter object. I decided against it because I struggled to think of a use case. I'll update then merge.

PR #914, opened 2026-05-31
Add viewable changelog

### LINKED ISSUE #907: Make change log visible in the WebUI when clicking the version number

Cooldude2606:
This change adds a changelog viewer modal to the web interface, accessible by clicking the version number in the header. The changelog is sourced from the root `CHANGELOG.md` removing any comments. This requires the addition of `react-markdown` to display the file correctly, although this does mean we can now support markdown else where, possibly in user added notes.

Also adds an "About" popup which contains: build info such as version, hash, and env; cluster info such as host, instance, and user count; support links to github and discord; and debug options to refresh the page and copy extra stats. The stats contain things such as page size, user agent, account permissions, enabled plugins.

### Changelog

```
### Features
- Added changelog viewer accessible from the version number in the web interface. #907
- Added about popup to user menu containing support links, build, clusterio, and debug info. #914
```

Closes #907


packages/web_ui/src/components/ChangeLogModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+				open={open}
+				onCancel={() => setOpen(false)}
+				footer={null}
+				width={800}
+			>
+				<div style={{ maxHeight: "70vh", overflow: "auto" }}>
+					<ReactMarkdown>
+						{cleanedChangelog}
+					</ReactMarkdown>
+				</div>
```
Hornwitser: Does this overflow wrapper make sense? Antd Modal already handles overflow with a scrollbar for the whole modal. The height set here also creates double scrollbars if open this on a phone or small tablet in landscape view.

Hornwitser (changes requested):
There doesn't appear to be any way to open the changelog modal on a narrow screen. The version text is not there, and there are no alternative ways to trigger it. This needs to be addresed. Other issues brought up here is for information only and does not block merge.

Image

Neat feature, but also kind of expensive, It's adding 167 kB of JavaScript to display a 53kB changelog, so I presume there's 100kB+ of libraries added into there. Size of all the .js assets grows from 2701.2 kB to 2868.6 kB.

packages/web_ui/src/components/ChangeLogModal.tsx
Cooldude2606: I thought it looked better to have the content scroll internally; but the double scrollbar is an issue I did not spot so will revert to the default antd behaviour. Probably more consistent that way as well/

Hornwitser: You could avoid the double scrollbars by having a height of something like `calc(100vh - Xrem)`, where X is however much margin to give for the top and bottom. But this has the problem of giving you a really bad space usage on a landscape mobile viewport size.

Another possibility is a media query so it only applies for large screens. But that's another can of worms.

Cooldude2606:
> There doesn't appear to be any way to open the changelog modal on a narrow screen. The version text is not there, and there are no alternative ways to trigger it. This needs to be addresed.

The first thought I had was to add a button in the user menu, but this lead me down a rabbit hole and now I have implimented an "About" page which displays general info such as version, build date, host count, and most importantly a button to view the change log. There is also a button to copy debug infomation which might be useful to us.

image

> Neat feature, but also kind of expensive, It's adding 167 kB of JavaScript to display a 53kB changelog, so I presume there's 100kB+ of libraries added into there. Size of all the .js assets grows from 2701.2 kB to 2868.6 kB.

That is the cost of supporting direct markdown rendering. Perhaps we could do server side rendering to reduce package size in the future. Also, I think having access to markdown will be useful for user created notes which many have asked for and I plan to ask for in the future. So overall, I consider it a worthwhile cost.

packages/web_ui/src/components/ChangeLogModal.tsx
Cooldude2606: Don't sound like great alternatives because of their drawbacks, so I will use your suggestion of removing the extra div. I wasn't too precious about it anyway.

packages/web_ui/src/components/AboutModal.tsx
```diff
@@ -0,0 +1,193 @@
...
+							<Button icon={<InfoCircleOutlined />} onClick={onOpenChangelog}>
+								Changelog
+							</Button>
+							<Button
+								target="_blank"
+								href="https://github.com/clusterio/clusterio/issues"
+								icon={<span style={{ display: "flex", alignItems: "center" }}>
+									<img
+										src="https://github.githubassets.com/favicons/favicon-dark.svg"
+										style={{ width: 16 }}
```
Hornwitser: We should not be loading tiny assets from external sources. This makes the UI fragile and allows random third parties to track usage of it.

Download the asset to packages/web_ui/src/images and reference that.

packages/web_ui/src/components/AboutModal.tsx
```diff
@@ -0,0 +1,193 @@
...
+							>
+								GitHub Issues
+							</Button>
+							<Button
+								target="_blank"
+								href="https://discord.gg/mzAsgnm"
+								icon={<span style={{ display: "flex", alignItems: "center" }}>
+									<img
+										// eslint-disable-next-line max-len
+										src="https://cdn.prod.website-files.com/6257adef93867e50d84d30e2/66e3d7f4ef6498ac018f2c55_Symbol.svg"
```
Hornwitser: Same here. Make the asset local.

Hornwitser (approved):
Very nice. There are a couple of externally loaded assets that needs to be made local, once that is fixed this is ready to merge.

I was a bit worried that the hosts/instances/users listing would break if you don't have the permission to list them. But it looks like it just says you've got 0 of those if you lack the permission for them.

Danielv123:
I like this a lot. Regarding the new About page - does it fully replace the old one, where it shows plugin versions etc?

Cooldude2606:
I did play around with having the plugins in the menu; but found it too busy and lacked the ability to display author and description etc. But plugins were too useful to exclude from the debug information so kept them there. So no it doesnt replace the plugin list; but perhaps it could be expanded later if we drop some details.

PR #915, opened 2026-05-31
Add version and connect button to player auth

### LINKED ISSUE #906: Player auth login screen should show which version of Factorio the servers are on.
The login screen for Player Auth shows which server to connect to in order to log in
Image

But this list doesn't tell you what version of the game it's on, or whether the game is on the public server list at all. This should be indicated.

Cooldude2606:
Adds additional server context to the Factorio login flow to improve usability. The server list now includes the running Factorio version and a direct connect option via Steam, reducing friction when using player auth.

The API has been extended to include the required data, with corresponding updates to the web interface. I also made the server list refresh every 15 seconds and refactored `verifyToken` effect hook to be more readable.

### Changelog
```
### Features
- Display Factorio server version in login server list. #906
- Add "Connect via Steam" button to login server list. #915
```
Closes: #906

### DIFF
--- a/plugins/player_auth/controller.ts
+++ b/plugins/player_auth/controller.ts
@@ -57,7 +57,7 @@ export class ControllerPlugin extends BaseControllerPlugin {
 				}
 
 				const host = this.controller.hosts.get(assignedHost);
-				if (!host || !host.publicAddress) {
+				if (!host) {
 					continue;
 				}
 
--- a/plugins/player_auth/controller.ts
+++ b/plugins/player_auth/controller.ts
@@ -3,11 +3,12 @@ import express, { type Request, type Response } from "express";
 import util from "util";
 import jwt from "jsonwebtoken";
 
+import { Static } from "@sinclair/typebox";
 import * as lib from "@clusterio/lib";
 import { BaseControllerPlugin } from "@clusterio/controller";
 const { basicType } = lib;
 
-import { FetchPlayerCodeRequest, SetVerifyCodeRequest } from "./messages";
+import { FetchPlayerCodeRequest, PlayerAuthServer, SetVerifyCodeRequest } from "./messages";
 
 
 async function generateCode(length: number): Promise<string> {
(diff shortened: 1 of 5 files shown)


plugins/player_auth/web/index.tsx
```diff
@@ -89,9 +124,35 @@ function LoginForm(props: LoginFormProps) {
...
+											<Text type="secondary"> ({server.factorioVersion})</Text>
+										)}
+									</span>
+
+									<Button
+										size="small"
+										onClick={() => {
+											window.location.href =
+												`steam://run/427520//--mp-connect=${server.address}`;
+										}}
```
Hornwitser: Is there a good reason to use onClick here instead of href?

Hornwitser (changes requested):
Code is good, but user experience wise this is strange and obtuse. You're presented with this:

Image

### Problems
- Connect button doesn't communicate that you have to be using the steam version of Factorio for it to work.
- Connect button doesn't communicate that it does nothing if you're already running the steam version of Factorio.
- The connect address for multiplayer servers that player_auth plugin is enabled on is now available to completely unauthenticated users. This should be documented.
- The connect address is available, but there's no way for you to see or make use of it via the interface if you're not running steam, or if Factorio is already running and you don't want to close and re-launch Factorio just to connect to the server.

plugins/player_auth/web/index.tsx
Cooldude2606: I copied the same logic which was used for the connect button on InstanceViewPage. If it works as a href then I will update both locations.

Cooldude2606: With the confirmation modal, the button's action is now to open the dialog rather than navigate, so onClick is now more appropriate. The actual navigation only occurs after the user confirms, which also performs other actions.

Cooldude2606:
The connect button now opens the following modal which should address all your concerns.

image

Additionally the sharing of connection addresses is now a config option which defaults to false. The description of the config option will act as the documention of it being available to unauthenticated users.

image

plugins/player_auth/controller.ts
```diff
@@ -46,12 +47,33 @@ export class ControllerPlugin extends BaseControllerPlugin {
...
-			let servers: string[] = [];
-			for (let instance of this.controller.instances.values()) {
-				if (instance.status === "running" && instance.config.get("player_auth.load_plugin")) {
-					servers.push(instance.config.get("factorio.settings")["name"] as string || "unnamed server");
+			const servers: Static<typeof PlayerAuthServer.jsonSchema>[] = [];
+
+			for (const instance of this.controller.instances.values()) {
+				const pluginLoaded = instance.config.get("player_auth.load_plugin");
+				const assignedHost = instance.config.get("instance.assigned_host");
+				if (instance.status !== "running" || !pluginLoaded || !assignedHost) {
```
Danielv123: Host ID can be 0

plugins/player_auth/messages.ts
```diff
@@ -1,5 +1,36 @@
...
+		public factorioVersion?: string,
+	) {}
+
+	toJSON(): Static<typeof PlayerAuthServer.jsonSchema> {
+		const json = { name: this.name } as Static<typeof PlayerAuthServer.jsonSchema>;
+		if (this.address !== undefined) {
+			json.address = this.address;
+		}
+		if (this.factorioVersion !== undefined) {
+			json.factorioVersion = this.address;
```
Danielv123: Typo I assume?

plugins/player_auth/controller.ts
```diff
@@ -46,12 +47,33 @@ export class ControllerPlugin extends BaseControllerPlugin {
...
+					continue;
+				}
+
+				const host = this.controller.hosts.get(assignedHost);
+				if (!host) {
+					continue;
 				}
+
+				const address = instance.gamePort !== undefined
+					? `${host.publicAddress}:${instance.gamePort}`
```
Danielv123, line 65: Do we want to handle publicAddress not being configured? Maybe hide the button? I don't think I have ever attempted to leave it unconfigured so not sure if we support that.

Danielv123 (changes requested):

plugins/player_auth/messages.ts
Cooldude2606: Yes, but makes me wonder how the tests are passing when they enumerate the field states... might to look into the test suite in more detail later. Or how any of it is working for that matter.

plugins/player_auth/controller.ts
Cooldude2606: It non-nullable, but an empty string could be set. I will cover that case by hiding the button as suggested.

PR #917, opened 2026-06-01
Enhance user filters and improve UI components

### LINKED ISSUE #905: Allow role and other filters on the user list
Allow filtering by instance, role, ban, admin, and whitelist status on the user list. If pages are added then it will also require the abilty to search by name.
When filtering by instance, it should show only players that have visited that instance. This can be used to for example see who was the last player to visit a certain instance, or when a certain player visited a certain instance.

Cooldude2606:
This PR enhances the usability of the users page by expanding filtering capabilities.

Key additions include user status filtering (admin, whitelist, banned) integrated into the existing name filter, as well as an instance selector to scope data per instance. Supporting changes were made to improve component reuse, loading behaviour, and consistency between instance-specific and global views.

image

image

### Changelog

```
### Features
- Added user status filtering (admin, whitelisted, banned) to the users table. #905
- Added instance selector to users page to scope user stats by instance. #905

### Changes
- Unified instance and global user table columns for consistent behaviour. #917
- Improved loading states for tables and inputs using synced state. #917
- Refactored username rendering into a reusable component. #917

### Fixes
- Fixed active filter icon behaviour in user search using empty string. #917
```
Closes: #905

### DIFF
--- a/packages/web_ui/src/components/InputInstance.tsx
+++ b/packages/web_ui/src/components/InputInstance.tsx
@@ -1,0 +1,42 @@
+import { Select } from "antd";
+
+import { InputComponentProps } from "../BaseWebPlugin";
+import { useInstances } from "../model/instance";
+
+interface BaseInputInstanceProps extends InputComponentProps {
+	includeAll?: boolean;
+}
+
+function BaseInputInstance(props: BaseInputInstanceProps) {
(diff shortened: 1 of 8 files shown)


Hornwitser (approved):
Looks good.

PR #919, opened 2026-06-02
Fix incorrect use of `path.parse` within create template

Cooldude2606:
As title. #877 introduced a bug which would prevent create from making templates for plugins.

### Changelog
```
### Fixes
- Fix incorrect use of `path.parse` within create template for plugins. #919
```


Hornwitser (approved):
Tested fine generating a basic plugin template.

PR #921, opened 2026-06-21
Refactor/plugin hooks

### LINKED ISSUE #730: Refactor how plugins work
The current way of defining plugins has problems leave a lot to be desired.

1. The current system invents its own `constructor` via the `async init` method. This causes two major problems:
    1. If the init method thrown you're left with a partially constructed plugin, and it's unclear if say the "exit" hook should be invoked or not.

    2. The that language support is built around the `constructor` method, for example in typescript special type inference logic is tied to code in the constructor, when the actions are done in `init` instead TypeScript complains that members are not set and a `!` has to be added to them to shut up the compiler.

2. Control is inverted. Instead of the programmer declaring exactly what they want to be called where and when in the plugin, the plugin system instead magically makes control flow into the plugin. This was one of the major pain points that the link refactor sought to remove entirely.

3. All plugins always participate in all hooks, even if a plugins doesn't need any of them. This is a result from hooks being methods on the plugin class with default implementations doing nothing.

I would like to simplify the mental model around how plugins work and be much less opinionated around how plugins should be made. In essence that means:
1. Using a simple async load function instead of a class as the entrypoint of a plugin. If a class is desired as the model for the plugin the entrypoint can instantiate the class.
2. Using hooks in the application to allow plugins to extend behaviour. This means calling app.foo.on("event") instead of having a onFooEvent method that's magically invoked.

For example the instance plugin code for global chat could look like this (simplified from original code).

```ts
import * as lib from "@clusterio/lib";
import type { Host, Instance } from "@clusterio/host"
import { ChatEvent } from "./messages";

class GlobalChatPlugin {
    constructor(public instance: Instance) {
    }

    async handleChatEvent(event: ChatEvent) {
        let content = `[${event.instanceName}] ${event.content}`;
        await this.instance.sendRcon(`/sc game.print('${lib.escapeString(content)}')`, true, "global_chat");
    }

    async sendChat(message: string) {
        this.instance.sendTo("allInstances", new ChatEvent(this.instance.name, message));
    }

    async onOutput(output: lib.ParsedFactorioOutput) {
        if (output.type === "action" && output.action === "CHAT") {
            await this.sendChat(output.message);
        }
    }
}

// helper defined elsewhere
function logError<T extends any[]>(logger: any, fn: (...args: T) => Promise<void>) {
    return (...args: T) => {
        fn(...args).catch(err => {
            logger.error(`Unexpected error:\n${err.stack}`);
        });
    }
}

// Exported by @clusterio/host
type InstanceContext = {
    instance: Instance;
    host: Host;
    logger: lib.Logger;
    info: lib.PluginNodeEnvInfo
    // ...
}

export async function loadInstancePlugin(context: InstanceContext) {
    const { logger, instance } = context;
    const plugin = new GlobalChatPlugin(instance)

    instance.handle(ChatEvent, plugin.handleChatEvent.bind(plugin));
    instance.server.on("output", logError(logger, plugin.onOutput.bind(plugin)));
}
```

This will of course require that the available hooks for plugins to use are clearly documented and discoverable.

Cooldude2606:
This PR introduces a typed hook system for plugins and refactors plugin loading to support it, while maintaining backwards compatibility with the old method of returning named classes.

* Adds `AsyncHook` collectors for controller, host, instance, web, and ctl plugins
* Plugins should now register handlers via hooks rather than relying on class methods
* Backwards compatibilty in maintained through the addition of `fromContext` on base plugin classes
* Improves type safety across plugin APIs, including component extension points and load context
* Web plugin fields (`extensionComponents`, `pages`, etc) are also collected via hooks once on load

### Changelog

```
### Features
- Added typed hook system for plugin registration. (#730)

### Changes
- Plugins should now return a function to register handlers via hooks. (#730)
- Improved typing for plugin APIs and webui extension points. (#921)
- Additional warnings have been added in various places. (#921)

### Breaking Changes
- Replaced `plugins` map with `loadedPlugins` set across core classes (except WebUi). (#921)
```

Closes: #730

TODO
[ ] Fix existing tests
[ ] Add new tests
[ ] Update templates


Cooldude2606:
Branch fell too far behind and accumulated too many conflicts to rebase. This will be revisted on a clean branch.

PR #922, opened 2026-06-22
Feature/pnpm overrides architecture

### LINKED ISSUE #728: Include `pnpm-lock.yaml` for reproducible builds
First, thank you for all the hard work maintaining this project!

I’m reaching out to request that the `pnpm-lock.yaml` file be included in the repository. Adding this lockfile can provide several benefits, especially for ensuring reproducible builds.

- The lockfile ensures that all contributors, CI environments, and deployment pipelines install the exact same dependency versions. This eliminates potential inconsistencies between environments.
- Tools like Nix rely on deterministic environments, and the lockfile is a critical piece for pinning dependencies effectively when using `pnpm`.
- Committing the lockfile allows pnpm to skip re-resolving dependencies, speeding up installation and improving the development experience (while decreasing CI runtimes 😉).

Please consider adding `pnpm-lock.yaml` to the repository and remove the lockfile line from the `.gitignore`. If there’s anything I can do to assist with this change, I’d be happy to help.

- [PNPM Docs (Lockfiles)](https://pnpm.io/git#lockfiles)
- [NPM Docs](https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json)

solarcloud7:
Resolves #728.

Commits `pnpm-lock.yaml` and removes it from `.gitignore`.    When anyone runs pnpm install, they get the exact same dependency versions every time, instead of pnpm picking whatever's newest.  To make a committed lockfile  *safe and useful* rather than just present, this PR also:

  - **Enforces singleton `webpack` / `react` / `@clusterio/lib`** 
  Used `pnpm.overrides`. A committed lockfile pins the entire dependency graph — so without this, any duplicate-dependency resolution would be frozen in too. The overrides force one copy across the workspace, which additionally fixes the long-standing Webpack "two React instances" error when building plugin web UIs.
  - **Switches CI to `pnpm i --frozen-lockfile`** 
  CI fails loudly if the lockfile drifts from the manifests instead of silently rewriting it (the old `--no-frozen-lockfile` hid drift, which would have undermined the whole point of committing the lock).
  - **Adds an `actions/cache` step** 
  Now that it's keyed on the manifests + lockfile, restoring pnpm's store between runs to cut dependency-download time (the CI-speed benefit noted in #728).

  ### External-plugin workflow change

 #### Problem: 
Previously, everything under external_plugins/ was automatically added to the pnpm workspace. Now that pnpm-lock.yaml is committed, that auto-incorporation would make the lockfile churn every time a developer adds or changes a private plugin — causing constant pnpm-lock.yaml merge conflicts between developers working on different external plugins.

#### Fix: 
external_plugins/* is no longer auto-added to the workspace. Using --dev-plugin with an external plugin now requires manually adding it to pnpm-workspace.yaml and marking it injected: true 
(documented in docs/writing-plugins.md and external_plugins/README.md).

#### Why injection? 
It's required specifically for external plugins because Webpack follows their symlink out of the workspace into their own node_modules, where it finds duplicate copies of shared deps. Marking them injected: true hard-links them inside the workspace boundary, so the pnpm.overrides apply and the duplicates collapse to one copy. Internal plugins don't need this — they stay symlinked (which preserves HMR / live-reload) and the overrides already cover them.

### Limitations 
 -  Only Directly tests the react and webpack overrides (falsifiable conflicts). @clusterio/lib uses the same mechanism and is additionally guarded at runtime by checkSingletonImport, which throws if two copies ever load.  The current limitation is fully simulating the external/injected path (a fixture with its own node_modules with webpack, react, clustorio deps so it can be tested  both directions).  I tested manually  but to fully automate it, it's a larger test deferred to the follow-up with our TAP tests.    

  ## How to verify
  ```
  # Lockfile is authoritative (matches manifests):
  pnpm install --frozen-lockfile        # succeeds; "Lockfile is up to date"

  # Overrides collapse the conflicting react@^17 and webpack@^4 to one copy each:
  grep -A8 'test/mock_external_plugin:' pnpm-lock.yaml
  #   react   specifier ^18.2.0  (rewritten from ^17)
  #   webpack specifier ^5.98.0  (rewritten from ^4)
  ls node_modules/.pnpm | grep -iE '^react@'                                    # one: react@18.3.1
  ls node_modules/.pnpm | grep -oE '^webpack@[0-9]+\.[0-9]+\.[0-9]+' | sort -u  # one: webpack@5.105.2

  # Falsifiable — remove '"react": "catalog:"' from pnpm.overrides, then:
  pnpm install --lockfile-only && grep -c 'react@17' pnpm-lock.yaml   # > 0: a second copy appears

  # Builds use the singleton:
  pnpm run build && pnpm run build-mod   # no duplicate-dependency errors
  ```

  
 ### Changelog
  ```
  ### Features
  - Commit `pnpm-lock.yaml` for reproducible installs across contributors and CI. #728
  - Cache the pnpm store in CI to speed up dependency installation. #728
  - Enforce a single copy of webpack, react and @clusterio/lib across the workspace with pnpm.overrides, fixing duplicate-dependency errors when building plugin web UIs. #728
  - Document developing and testing external plugins with `injected: true`. #728

  ### Changes
  - CI now installs with `--frozen-lockfile`, failing if the committed lockfile drifts from the manifests. #728

  ### Fixes
  - Stop mocha walking into installed `node_modules` and executing @clusterio/lib's browser bundle as a test spec. #922

  ### Breaking Changes
  - Plugins under `external_plugins/` are no longer automatically added to the pnpm workspace. To use `--dev-plugin` with an external plugin, add it to `pnpm-workspace.yaml` and mark it `injected: true` (see `docs/writing-plugins.md`). This prevents lockfile merge conflicts now that `pnpm-lock.yaml` is committed.
  ```


solarcloud7:
The failing check is a flaky Factorio integration test, not anything in this PR.

 `Clusterio Instance > whitelist creation > should always create a whitelist json even if unused` →
 `Error: Factorio server unexpectedly shut down with code 1`
 (the second "failing" is just its `after each` cascade). The other three `test` matrix jobs were cancelled by `fail-fast`, not independently failing, and `lint` + `build` pass.

 The same commit (`e2a50ca8`) passed **all** test jobs — including `test (22.x, 1.1.110)`, the exact combo that failed    
  here — on my fork:
  https://github.com/solarcloud7/clusterio/actions/runs/27924897622

Would anyone be able retrigger this test or I can make a dummy commit to re-trigger if that's easier.
Meanwhile i'll look into this bug since i'm about to push a TAP poc update very soon also!

Danielv123:
Why do we need to manually edit the .yml file to add the inject: true? Why can the --dev-plugin handle this?

Having to manually manage external plugins again seems like an unnecessary pain, because whatever I do manually will probably have the same issues as doing it automatically

solarcloud7:
Good push — you're right that re-introducing a manual inject step is the wrong direction. Two things were tangled together:

**Why `--dev-plugin` can't do it:** they're different layers. `--dev-plugin` is a flag to the already-running controller (it points webpack-dev-middleware at the plugin's `webpack.config.js` at startup); `injected: true` is a `pnpm install`-time decision about the node_modules layout. By the time the controller runs, install is already done.

**The bigger finding: if you *clone* the plugin into `external_plugins/`, you don't need `injected: true` at all.** As a normal in-repo workspace member, pnpm links its deps to the workspace singletons (the controller's webpack and the plugin's webpack end up the *same copy*), and it stays a plain symlinked member so HMR/live-reload keeps working — injection actually breaks that. So the inject dance was over-prescribed; I've simplified the docs.

**The case that genuinely still needs injection is symlinking a plugin from *outside* the repo** (what `external_plugins/` was originally for). There the realpath escapes the workspace into the plugin's own `node_modules` and the duplicates come back — an install-layout thing `--dev-plugin` can't retrofit at runtime. I added a helper (`pnpm dev-plugin <path>`) that detects which case you're in and only prints the inject snippet for the symlink-out one.

The one remaining manual bit — a line in `pnpm-workspace.yaml` — is unavoidable now that the lockfile is committed (auto-globbing `external_plugins/*` would churn it). You're right that a script hits the same constraint, so the helper just removes the typing, not the requirement.

To make this auditable, I added the build-capable fixture the *Limitations* section flagged as missing (`test/manual/external_plugin_fixture/`): `node test/manual/external_plugin_fixture/verify.js` shows the deliberately-conflicting `react@^17` / `webpack@^4` collapse to single workspace copies and the web bundle compile clean; removing the overrides makes it report the failure. (It reproduces resolution + build — a faithful replay of what `--dev-plugin` invokes — not a full live-controller boot.)

solarcloud7:
Let me test this a little more.  It's super hard to simulate these different external plugins  against npm install,  vs cloning git, vs me running in docker.    I'm in over my head at the moment.  And I've been wrong enough today that I should probably literally do a fresh clusterio install and walk each of these locally more thoroughly.

PR #924, opened 2026-06-22
Rework site layout into a fixed sidebar

bbassie:
## Summary
Removes the top header and folds its contents into the sidebar, which is now sticky and full height:
image

- Logo + version at the top (the version still opens the changelog)
- Navigation in the middle (scrolls on its own if it overflows)
- Account menu pinned to the bottom, opening to the right

The sidebar is rebuilt with antd `Flex`, `Typography` and `Menu`, so almost all of the styling now comes from antd and the custom css for this layout is reduced to a single sticky rule (`position: sticky` is the one thing antd doesn't provide).

The nav menus also use `theme="dark"` to match the dark `Sider`. Previously the sidebar menu had no theme, so under the app's `darkAlgorithm` it was treated as a light menu and coloured the active item with the primary blue on a faint tint, which read as low contrast on the dark background; a dark menu uses white text on a solid primary background instead.

## Notes
- No behaviour change to navigation or the account actions, only layout and styling.
- Net reduction of ~70 lines.

## Testing
Built (`tsc` + webpack) and viewed locally against a running controller.

### Changelog
```
### Changes
- Reworked the site layout into a fixed sidebar with a top bar. #924
```

### DIFF
--- a/packages/web_ui/src/components/SiteLayout.tsx
+++ b/packages/web_ui/src/components/SiteLayout.tsx
@@ -218,7 +218,7 @@ export default function SiteLayout() {
 				aria-label="Toggle navigation menu"
 			/>
 			<Link to="/" style={{ display: "flex", alignItems: "center" }}>
-				<img src={logo} width={32} height={32} alt="Clusterio logo" />
+				<img src={logo} width={56} height={56} alt="Clusterio logo" />
 			</Link>
 			<Flex vertical justify="center">
 				<Typography.Title level={4} style={{ margin: 0, lineHeight: 1.2 }}>Clusterio</Typography.Title>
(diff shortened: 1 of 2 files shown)


Danielv123 (approved):
Image

Image

I like it, works fine on mobile as well. One thing we could consider is adhering to normal hamburger menu conventions on phone screens - having the menu show on top of content instead of shifting it over and going away on navigation.

bbassie:
@Danielv123 Because the drawer did not have a natural position for the burger Icon I added the header back just more compact.

Cooldude2606:
Overall this is nicer, but the logo has been made much smaller when I think it should not have been. Can we see if it can be made larger before we merge this.

Cooldude2606 (approved):

psihius (approved):

PR #930, opened 2026-06-24
Persist table sort/filter state in the URL

### LINKED ISSUE #927: Tables and filters should remember their state on back-forward
Changing the sort order and/or adding a search filter to tables such as Hosts/Instances/Users/Roles/etc should modify the query part of the URL so that back-forward and linking to the page shows the same content when it's reloaded.

bbassie:
Closes #927.

Tables now write their **sort order, column filters and pagination into the URL query string**, so reload, link-sharing and browser back/forward all restore the same view.

## URL scheme

Each table namespaces its params with its name. **System keys** (`sort`, `page`, `size`) are reserved; **filterable columns** are `<table>.<column>` — so a filter field can never collide with a system key.

```
?user.sort=name.desc          sort: <field>.<asc|desc>
?user.page=2&user.size=20     pagination
?user.name=bad&user.admin=true  column filters
?instance.name=foo
```

- Sort packs the direction into the value (`name.desc`). `<table>.sort` is parsed tolerant of repeats, so multi-column sort can be added later without changing the URL shape (single-column for now).
- Multi-value filters use repeated params (`user.roles=1&user.roles=2`).
- The bundled user filter persists as separate readable params (`user.name`, `user.admin`, …) via a per-column `FilterCodec` rather than one opaque JSON blob.

## What's added

- **`useTableQueryState`** (`util/useTableQueryState.ts`) — makes an antd `<Table>` controlled and round-trips sort/filter/pagination through the URL via `useSearchParams`. Pure serialisation helpers live in `util/tableQuery.ts`.
- **`useColumnSearch`** (`components/useColumnSearch.tsx`) — a column-header text-search filter (Users-table style), persisted through the same mechanism.
- Both exported from `@clusterio/web_ui` so plugins can use them.

## Tables covered

Hosts, Roles, Plugins, the instance list (standalone Instances page + host detail page), Users (main table + online-players list on the instance detail page), instance saves, per-instance user stats, Mod Packs, Stored Mods, and the **subspace_storage** Storage page (its resource search moved into the Resource column header, keeping its regex matching).

## Verification

- Manually verified in the dev UI: changing sort/filter/pagination updates the query string (e.g. `user.sort=name.desc`, `user.size=20`), and reload / back-forward / opening the URL in a new tab restore the view.

### Changelog
```
### Features
- Tables now remember their sort and filters in the URL. #927
```


packages/web_ui/src/components/UsersTable.tsx
```diff
@@ -136,45 +150,53 @@ export default function UsersTable({ instanceId, onlyOnline = false, pagination,
...
+			sortOrder: tableState.sortOrder("firstSeen"),
 		},
 		{
 			title: "Last Seen",
 			key: "lastSeen",
 			filterMultiple: false,
 			filters: durationFilters,
-			defaultFilteredValue: onlyOnline ? ["online"] : undefined,
+			// Default to online-only when nothing for this column is in the URL yet.
+			filteredValue: tableState.filteredValue("lastSeen") ?? (onlyOnline ? ["online"] : null),
```
Cooldude2606: This comment is not correct. The code correctly depdents on the "onlineOnly" prop, but the comment makes no mention of this.

packages/web_ui/src/components/UsersTable.tsx
```diff
@@ -29,10 +30,20 @@ export interface UsersTableProps {
...
-export default function UsersTable({ instanceId, onlyOnline = false, pagination, size }: UsersTableProps) {
+export default function UsersTable(
+	{ instanceId, onlyOnline = false, pagination, size }: UsersTableProps
+) {
 	const [roles, rolesSynced] = useRoles();
 	const [users, usersSynced] = useUsers();
 	const navigate = useNavigate();
+	const callerPagination = pagination && typeof pagination === "object" ? pagination : undefined;
+	const defaultPageSize = (callerPagination as { defaultPageSize?: number })?.defaultPageSize ?? 50;
+	const tableState = useTableQueryState<lib.UserDetails>({
```
Cooldude2606: This looks like a bug fix that wasn't disucssed; where previously default page size was not extact. On the other hane it could be that table query state should accept pagiantion as given not not reuiqre this level of extraction. I would suggest reviewing how "pagination" within "useTableQueryState" is expected to act.

packages/web_ui/src/util/useTableQueryState.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface TableQueryStateOptions {
+	/** Table name used to namespace every query parameter, e.g. "user" -> `user.sort`, `user.name`. */
+	namespace: string;
+	/** Column key sorted by default (no `sort` param present). Omit for no default sort. */
+	defaultSortKey?: string;
+	/** Direction of the default sort. Default "ascend". */
+	defaultSortOrder?: SortOrder;
+	/** Pass an object to persist pagination; omit/false for `pagination={false}` tables. */
+	pagination?: false | { defaultPageSize?: number };
```
Cooldude2606: If default is "false" then why do many of the tables give it explictly? This should be cleaned up.

packages/web_ui/src/util/useTableQueryState.ts
```diff
@@ -0,0 +1,99 @@
...
+
+	const [params, setParams] = useSearchParams();
+
+	const sort = parseSort(params, prefix, defaultSortKey, defaultSortOrder);
+	const filters = parseFilters(params, prefix);
+	const page = paginationEnabled ? parsePagination(params, prefix, defaultPageSize) : undefined;
+
+	return {
+		onChange(pagination, tableFilters, sorter) {
+			const single = Array.isArray(sorter) ? sorter[0] : sorter;
```
Cooldude2606, line 81: It is not obivous to me why only the first is used, this should include a comment,

Cooldude2606 (commented):
All code appears to work and fits the requirements. Although I do have some comments I'd like comment on before I approve.

Cooldude2606 (changes requested):
Could you please add "Search" to the name columns for Mods and ModPacks as they are now the only tables without this feature.

I have discovered another inconsistency, but it is up to you if you wish to fix it within this PR or postpone because it feels complex for little gain. The users table currently uses a "live search" where it filters as you type, where as all other tables filter only on confirmation. Live search is of course better; however, the current implimention updates the URL params for every change (every single letter typed) this is not great. Idealy all tables would use live search and only update the url params on confirm / hide. Do let me know what you think, if you want to postpone we can create a feature ticket for it.

Cooldude2606 (changes requested):
Everything is working as expected which is great, except for the clear button which fails to update the url. This applies to both the user tables and all other tables. Once fixed I am happy to merge.

Image

Cooldude2606:
This inadvertently made it so clear doesnt auto close; but I think that is actully better. But now we need to make it so clear no longer updates the URL to be consistent with it no longer closing the search box.

Cooldude2606 (approved):

PR #931, opened 2026-06-25
Make navigation open in new tabs (real links)

### LINKED ISSUE #929: Make "open in new tab" work for all links
The "open in new tab" context menu option is not available in most cases. This has been noted to break a number of people's workflows.

bbassie:
Closes #929.

In-app navigation was done with JS `onClick` → `navigate(...)` handlers instead of real `<a href>` anchors, so the browser's right-click **"Open in new tab"** (and middle-click / ctrl-click) was unavailable in most places. This converts the link-like navigation to real anchors using the existing `Link` component (a real `<a>` via react-router `useHref`, basename-correct, SPA on plain left-click).

## Changes

- **New `useRowNavigation` hook** for table rows: left-click navigates in place; **middle-click / ctrl-click / cmd-click open the target in a new tab**; the middle-mouse autoscroll cursor is suppressed; clicks landing on a real link inside the row are ignored so the anchor handles them.
- **Table name cells become real `Link`s** (`color: inherit`, so they look the same) on Hosts, Roles, Plugins, Instances, Users, the user instance-stats table and Mod Packs. The name cell carries the right-click **"Open in new tab"** context menu; the rest of the row keeps the click/middle-click behaviour via `useRowNavigation`. (A `<tr>` can't be an anchor, so the context menu lives on the name cell.)
- **Sidebar nav items** and the **account "user"** entry render their labels as `Link`s, keeping the `Menu` `onClick` for plain left-clicks (`stopPropagation` avoids double navigation).
- The **"Go back to instances list"** alert action becomes a `Link`.

Genuine actions (download/delete/modify/log out), post-create navigation, and external links are intentionally left as JS/anchors as appropriate. Breadcrumbs were already `Link`s.

## Verification

- Manually verified in the dev UI: right-click a table name → "Open in new tab"; middle-click / ctrl-click anywhere on a row opens the right page in a new tab; left-click still navigates in place. Sidebar items and the account entry support right-click / middle-click new tabs. Opened tabs land on the correct route (basename preserved). Genuine actions and external links are unchanged.

### Changelog
```
### Features
- Navigation links can now be opened in a new tab. #929
```

### DIFF
--- a/packages/web_ui/src/components/InstanceList.tsx
+++ b/packages/web_ui/src/components/InstanceList.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import { useNavigate } from "react-router-dom";
 import { message, Button, Table, Space } from "antd";
 import CopyOutlined from "@ant-design/icons/CopyOutlined";
 import type { SizeType } from "antd/es/config-provider/SizeContext";
@@ -14,6 +13,7 @@ import InstanceControlButton, { InstanceControlButtonPermissions } from "./Insta
 import * as lib from "@clusterio/lib";
 import Link from "./Link";
 import { instancePublicAddress } from "../util/instance";
+import useRowNavigation from "../util/useRowNavigation";
 
 const strcmp = new Intl.Collator(undefined, { numeric: true, sensitivity: "base" }).compare;
 
@@ -25,9 +25,9 @@ type InstanceListProps = {
 
 export default function InstanceList(props: InstanceListProps) {
 	let account = useAccount();
-	let navigate = useNavigate();
 	let [hosts] = useHosts();
 	const [systems] = useSystems();
+	const rowNav = useRowNavigation();
 
 	function hostName(hostId?: number) {
 		if (hostId === undefined) {
@@ -49,6 +49,9 @@ export default function InstanceList(props: InstanceListProps) {
 			dataIndex: "name",
 			defaultSortOrder: "ascend",
 			sorter: (a, b) => strcmp(a["name"], b["name"]),
+			render: (_, instance) => <Link to={`/instances/${instance.id}/view`} style={{ color: "inherit" }}>
+				{instance.name}
+			</Link>,
 		},
 		{
 			title: "Assigned Host",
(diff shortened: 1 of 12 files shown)


Cooldude2606:
There is an issue where the link remains highlighted after middle clicking. In this image I middle clicked onto roles.
Also the right cllick is nice, but it only effects the name and not the whole row so it might be more confussing than useful.
Could you investigate these two issues before this is merged.

image

bbassie:
For the table there are 4 options:
A. Remove the link but then you don't have the right click menu (for tables like instances the host col is also a Link).
B. Remove the onRow click, and maybe add a action col at the end.
C. Wrap each non custom link cell (Most dirty fix)
D. Custom right click context menu on click of a table row.

My vote goes for B, this is the most consistent for UX I think.

If we go for B then I can also implement that the Link covers the full table cell because right now it does not.

Cooldude2606:
Having specific columns link else where is something I hadn't considered. In which case i think it makes sense as implemented. A custom context menu on row would be the prefered outcome but is overly complex to implement so I dont think should be pursued yet.

Happy to approve this after the highlight bug is fixed.

Cooldude2606 (approved):
Highlight issue is fixed and the full cell right click is a nice addition. I'm happy with this.

packages/web_ui/src/components/InstanceList.tsx
```diff
@@ -49,10 +49,15 @@ export default function InstanceList(props: InstanceListProps) {
...
 			sorter: (a, b) => strcmp(a["name"], b["name"]),
+			className: "table-link-cell",
+			render: (_, instance) => <Link to={`/instances/${instance.id}/view`} style={{ color: "inherit" }}>
+				{instance.name}
+			</Link>,
 		},
 		{
 			title: "Assigned Host",
 			key: "assignedHost",
+			className: "table-link-cell",
```
psihius, line 60: Possible issue: When an instance is unassigned (assignedHost === undefined), this class stretches the existing empty host <Link> over the whole Assigned Host cell. Clicking or opening that blank cell in a new tab now goes to /hosts/undefined/view instead of the instance row target, so the overlay/link needs to be skipped unless a host is actually assigned.

I'm not good at React, so this needs verification if it's applicable

bbassie: Fixed this in the latest commit so an unassigned host does not render the link.

bbassie:
In the latest commit, I also reverted the anchor tag overlay because otherwise clickable elements in that column always need to set position and zIndex. Better to have it display block then, and the user just needs to right-click on the name. If they want to copy the link.

psihius (approved):

PR #932, opened 2026-06-25
Remember the instance console chat/log toggle

bbassie:
The Console **Chat/Log** filter toggle on the instance page now persists its state to `localStorage`, so if you turn it off to view full logs, that choice is remembered across reloads (and other instance pages). It still defaults to **on** (chat only) when nothing is stored.

Small, contained change in `InstanceViewPage.tsx`: the `actionsOnly` state initialises from `localStorage` and writes back on toggle, matching the existing direct-`localStorage` style used in `App.tsx`.

## Verification
- Manually: toggle to "Log", reload → stays on "Log"; clear the key / fresh load → defaults to "Chat".

### Changelog
```
### Features
- Remember the instance console chat/log toggle. #932
```

### DIFF
--- a/packages/web_ui/src/components/InstanceViewPage.tsx
+++ b/packages/web_ui/src/components/InstanceViewPage.tsx
@@ -25,6 +25,7 @@ import { useHost } from "../model/host";
 import InstanceStatusTag from "./InstanceStatusTag";
 import Link from "./Link";
 import { instancePublicAddress } from "../util/instance";
+import useLocalStorage from "../util/useLocalStorage";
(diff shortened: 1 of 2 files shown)


Cooldude2606 (changes requested):
The code LGTM. Although there is a noticable delay in presing the toggle and it switching, but this is non-blocking. Please update the PR description to include a change log entry.

Cooldude2606 (approved):

Cooldude2606:
Thought about using `setImmediate` here; but it felt like it was very slightly less responsive. Not sure why that is, but I will leave it unchanged.

PR #933, opened 2026-06-25
Add cluster overview landing page

### LINKED ISSUE #928: Overview page
A landing page when opening the Web UI dashboard that shows a summary of the cluster operations. Ideas for content includes:
- Counts of instances/hosts running/connected
- Counts of users in total and online
- Counts of cpu/memory/disk total/in-use
- Controller uptime
- A window with errors recently logged.
- Link to Grafana dashboard if that's set up.

bbassie:
Closes #928.
image

Adds an **Overview** page at `/` — the top sidebar item and the dashboard landing page (the root route was previously blank when logged in). It summarises the cluster:

- **Hosts** connected / total, **Instances** running / total, **Users** online / total
- Aggregate **CPU / Memory / Disk** usage across host machines (used vs total + usage bar)
- **Controller uptime** (running since)
- **Recent errors** panel (reuses `LogConsole` with `all` + `maxLevel="warn"`)
- A **Grafana** link in the header when configured

Each card is gated by the relevant permission (`core.host.list`, `core.instance.list`, `core.user.list`, `core.log.follow`), so a limited user only sees what they can access.

### Backend
Adds a `controller.grafana_url` config field (optional string, mirrors `controller.public_url`). The page reads it via `ControllerConfigGetRequest` (gated on `core.controller.get_config`) and shows the Grafana button only when set.

### Notes
- CPU/mem/disk are summed across connected **host** systems; the controller system is used only for uptime.
- Built on existing model hooks (`useHosts`/`useInstances`/`useUsers`/`useSystems`), `system_metrics`, `lib.formatBytes`, and `LogConsole` — no new data plumbing.
- `webRoutes` already lists `/`, so no controller route change was needed.

### Verification
- `tsc --build` (lib + web_ui + controller) and `eslint` clean.
- Manually verified on a dev cluster: counts/metrics/uptime/recent-errors render and update; "Overview" is the landing page; setting `controller.grafana_url` shows the Grafana button (hidden when unset).

### Changelog
```
### Features
- Added a cluster overview landing page. #928
```

### DIFF
--- a/packages/lib/src/config/definitions.ts
+++ b/packages/lib/src/config/definitions.ts
@@ -18,6 +18,7 @@ export interface ControllerConfigFields {
 	"controller.bind_address": string | null;
 	"controller.trusted_proxies": string | null;
 	"controller.public_url": string | null;
+	"controller.grafana_url": string | null;
 	"controller.tls_certificate": string | null;
 	"controller.tls_private_key": string | null;
 	"controller.auth_secret": string | null;
@@ -136,6 +137,12 @@ export class ControllerConfig extends classes.Config<ControllerConfigFields> {
 			type: "string",
 			optional: true,
 			// TODO extendedValidation, valid url including the protocol
+		},
+		"controller.grafana_url": {
+			title: "Grafana URL",
+			description: "URL of a Grafana dashboard to link from the overview page, including the protocol.",
+			type: "string",
+			optional: true,
 		},
 		"controller.tls_certificate": {
 			title: "TLS Certificate",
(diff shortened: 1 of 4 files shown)


Cooldude2606 (changes requested):
Please swap the saves loaded card and the online time card so that cards of the same size are within the same row. Also you need to use `.subscribe` for the permission checks rather than `.list`. Otherwise the general code is good and happy to merge once fixed.

Cooldude2606 (approved):

PR #937, opened 2026-06-26
Add release channels (stable/experimental) as instance version targets

bbassie:
## Summary

Lets an instance's `factorio.version` target a **release channel** — `stable`, `experimental`, or any other key from `factorio.com/api/latest-releases` — in addition to the existing `latest` / `X.Y` / `X.Y.Z` values. When the instance starts, the channel is resolved to the concrete version that channel currently points to and then handled by the existing download/lookup path, so a channel auto-installs and runs just like a pinned version does.

## What changed

**lib**
- New `external/LatestReleases.ts`: `fetchLatestReleases()` (fetches the `latest-releases` JSON), `resolveReleaseChannel()` (reads a channel's `headless` build version), and schemas.
- `isReleaseChannel()` added; `isTargetVersion()` / `TargetVersionSchema` widened to accept channel names so they validate over the wire and on disk.
- `LatestReleasesRequest` (mirrors `FactorioVersionsRequest`) + registration, and a new `core.external.get_latest_releases` permission.
- `factorio.version` config description/validator updated to mention channels.

**controller**
- `latestReleases` `ValueCache` that proxies the API (re-fetched at most once per the request's `maxAgeMs`, 5 min by default) and persists the last successful response to `<database_directory>/latest-releases.json`, read back only when the API is unreachable. Request handler added in `BaseConnection`.

**host**
- `Instance.init()` resolves a channel target to a concrete version via the controller before `checkForUpdates`/`init`. If it can't be resolved (API down and nothing cached), the instance fails to start with a clear message.
- `FactorioServer.setTargetVersion()` to inject the resolved version.
- Clearer errors in `checkForUpdates`/`findVersion` when the host uses a direct (single-version) Factorio install, pointing at the versioned-layout requirement for automatic downloads.

**web_ui**
- `Control.latestReleases` cache; the version selector (`InputVersion`) now lists channels (e.g. `stable (2.0.77)`, `experimental (2.1.8)`) alongside `latest`.

## Notes / decisions

- **Offline behaviour:** the controller serves the disk-cached copy when the API is down; a channel target only fails if there's no cache at all.
- **Resolution semantics:** a channel resolves to the exact version the API reports for it (`headless` build) and reuses `checkForUpdates`, so it auto-downloads under the versioned layout on Linux — same as a pinned version.
- **`TargetVersion` typing:** the TS type is unchanged, but the schema/validator are intentionally permissive about channel names since the channel set is dynamic; resolution happens at start time.
- **web_ui inlines the channel→version lookup** instead of calling `resolveReleaseChannel`, because `@clusterio/lib` is a federation-shared module and an export only used inside a lazy component can be tree-shaken out.
- No automated tests added: the feature is primarily a thin proxy over an external API; happy-path resolution is exercised end-to-end manually.

### Changelog
```
### Features

- Instances can target a Factorio release channel (e.g. `stable` or `experimental`) as their version, resolved from `factorio.com/api/latest-releases`. #937
```

### DIFF
--- a/packages/controller/src/BaseConnection.ts
+++ b/packages/controller/src/BaseConnection.ts
@@ -33,6 +33,7 @@ export default class BaseConnection extends lib.Link {
 		this.handle(lib.ModPackGetDefaultRequest, this.handleModPackGetDefaultRequest.bind(this));
 		this.handle(lib.ModDownloadRequest, this.handleModDownloadRequest.bind(this));
 		this.handle(lib.FactorioVersionsRequest, this.handleFactorioVersionsRequest.bind(this));
+		this.handle(lib.LatestReleasesRequest, this.handleLatestReleasesRequest.bind(this));
 
 		this.handle(lib.SubscriptionRequest, this.handleSubscriptionRequest.bind(this));
 		this.connector.on("close", () => {
@@ -112,5 +113,9 @@ export default class BaseConnection extends lib.Link {
 
 	async handleFactorioVersionsRequest(request: lib.FactorioVersionsRequest) {
 		return await this._controller.factorioVersions.get(request.maxAgeMs);
+	}
+
+	async handleLatestReleasesRequest(request: lib.LatestReleasesRequest) {
+		return await this._controller.latestReleases.get(request.maxAgeMs);
 	}
 }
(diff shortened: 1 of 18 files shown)


Cooldude2606 (approved):
Add an integration test for starting an instance with a realease channel as the target, then we can merge.

PR #940, opened 2026-06-26
Fix antd Modal destroyOnClose deprecation and version TreeSelect duplicate-value warnings

bbassie:
## Summary

Fixes two Ant Design console warnings in the Web UI.

## What changed

- **`destroyOnClose` deprecation** — antd 5.25 deprecated the Modal `destroyOnClose` prop in favour of `destroyOnHidden` (`[antd: Modal] \`destroyOnClose\` is deprecated. Please use \`destroyOnHidden\` instead.`). Renamed it on all 10 `Modal` usages across the web UI (InstancesPage, RolesPage, NpmButton, UsersPage ×2, SavesList, LoadScenarioModal, AssignInstanceModal, ModPackViewPage, CreateSaveModal).
- **`TreeSelect` duplicate value** — the version selector (`InputVersion.tsx`) logged `Warning: Same \`value\` exist in the tree: 0.13` (and 0.14/0.15/…). A version group node uses the major.minor as its (selectable) value, and when the version list contains a bare major.minor entry — e.g. the `ApiVersions` fallback shown before the version fetch resolves — it was also added as a child with the same value. Children equal to their group key are now skipped, since the group node already represents that value.

No behaviour change beyond silencing the warnings.

### Changelog
```
### Fixes

- Silenced two Ant Design console warnings in the Web UI: the deprecated Modal `destroyOnClose` prop, and a duplicate-value warning from the instance version selector. #940
```

### DIFF
--- a/packages/web_ui/src/components/AssignInstanceModal.tsx
+++ b/packages/web_ui/src/components/AssignInstanceModal.tsx
@@ -67,7 +67,7 @@ export default function AssignInstanceModal(props: AssignInstanceModalProps) {
 			confirmLoading={applying}
 			onOk={handleAssign}
 			onCancel={handleCancel}
-			destroyOnClose
+			destroyOnHidden
 		>
 			<Paragraph style={{ maxWidth: "30em" }}>
 				Select a Host to assign this instance to.  Assignment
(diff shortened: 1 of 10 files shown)


Cooldude2606 (approved):
I seam to remember we didn't want to change to `destroyOnHidden` but Im not able to reproduce any kind of issue or discussion on it. So this is fine and its simple enough to revert if we do run into an issue later.

PR #941, opened 2026-06-26
Parse mod versions leniently like the game does

### LINKED ISSUE #926: Mods for Factorio 2.1 are impossible to use
If you try to upload or use a Factorio 2.1 mod in any way you'll be greeted with the following error

```
Error loading mod mod_name.zip: Mod's info.json is not valid
```

bbassie:
## Summary

Factorio reads version strings very leniently (leading whitespace is skipped, only the major and minor are required, the patch is optional, and anything after it is ignored), but clusterio validated a mod's `info.json` strictly against the hardcoded `ApiVersion` enum. That rejected mods the game itself loads fine. This makes clusterio normalise versions the same lenient way the game does, and stops using the enum to validate versions at all — the mod portal is now the source of truth.

## What changed

- **`version.ts`** — added `normaliseGameVersion()` (matches how the game reads versions: optional leading whitespace, required major.minor, optional patch, trailing content ignored, integers so leading zeros drop) backed by a regex-pattern `GameVersionSchema`, plus a `majorMinorVersion()` helper. Removed `ApiVersionSchema` / the `ApiVersion` type / `isApiVersion` / `normaliseApiVersion`. Kept the `ApiVersions` list used as a fallback in the web interface and added `2.1` to it.
- **`ModInfo.ts`** — a mod's `version` and `factorio_version` use the lenient `GameVersionSchema` and are normalised when read (`version` → `X.Y.Z`, `factorio_version` → `X.Y`); `factorioVersion` is now a `PartialVersion`. Unparseable versions throw a clear error.
- **`ModStore.ts`** — sends `majorMinorVersion(...)` as the portal `version` filter and, on a non-200, surfaces the portal's own error message instead of a generic status line.
- **`messages_mod.ts`, `ctl/commands_mod.ts`, web UI `ModsPage` / `ModPackViewPage`** — switched from `ApiVersion` / `normaliseApiVersion` to `PartialVersion` / `majorMinorVersion`.

## Tests

Replaced the `isApiVersion` / `normaliseApiVersion` tests with `normaliseGameVersion` / `majorMinorVersion` cases (including lenient inputs like `"1.1.x"`, `"  1.1"`, `"01.02.03"`, `"1.1.50-beta"`), updated the `ModStore` portal-error assertion, and added a `ModInfo` test proving a lenient `info.json` version normalises.

## Note

`majorMinorVersion` does not validate the version, so a Factorio version the portal doesn't support flows through and the portal's error is surfaced, rather than being pre-rejected against a fixed list. The serialised `ModInfo` schema uses the same lenient regex schema so portal-sourced data round-trips; normalised stored values still match it.

### Changelog
```
### Changes

- Mod `info.json` versions (a mod's version and its `factorio_version`) are now parsed leniently, matching how the game reads them, instead of being validated against a fixed list of Factorio versions. #941

### Fixes

- Surface the mod portal's own error when downloading mods for an unsupported Factorio version. #941
```

### DIFF
--- a/packages/controller/src/BaseConnection.ts
+++ b/packages/controller/src/BaseConnection.ts
@@ -80,7 +80,7 @@ export default class BaseConnection extends lib.Link {
 		return modPack;
 	}
 
-	getMod(mod: { name: string, version: lib.FullVersion, sha1?: string }) {
+	getMod(mod: { name: string, version: lib.GameVersion, sha1?: string }) {
 		let filename = lib.ModInfo.filename(mod.name, mod.version);
 		let modInfo = this._controller.modStore.files.get(filename);
 		if (!modInfo) {
--- a/packages/controller/src/BaseConnection.ts
+++ b/packages/controller/src/BaseConnection.ts
@@ -80,7 +80,7 @@ export default class BaseConnection extends lib.Link {
 		return modPack;
 	}
 
-	getMod(mod: { name: string, version: lib.GameVersion, sha1?: string }) {
+	getMod(mod: { name: string, version: lib.SourceVersion, sha1?: string }) {
 		let filename = lib.ModInfo.filename(mod.name, mod.version);
 		let modInfo = this._controller.modStore.files.get(filename);
 		if (!modInfo) {
--- a/packages/controller/src/ControlConnection.ts
+++ b/packages/controller/src/ControlConnection.ts
@@ -707,7 +707,7 @@ export default class ControlConnection extends BaseConnection {
 			const release = releases.releases
 				.filter(info => versionRange.testVersion(info.version))
 				.reduce<ModPortalReleaseType | undefined>((max, cur) => (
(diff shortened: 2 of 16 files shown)


Cooldude2606 (changes requested):
I'dd say this all looks good, but I notice some passing of known major minor strings into the major minor function which is duplicative work; so i think some of the internal data types can use a new `MajorMinorVersion` type with the current `majorMinorVersion` function being renamed to `normaliseMajorMinorVersion` and `isMajorMinorVersion` being added for future use. Also the game version regex is duplciated into the normalise game version function, we should use the same in both places.

Cooldude2606 (approved):
Made some small edits and renamed to `SourceVersion`, I have tested locally using the leading zeros mod from hornwitser. The test coverage was reduced due to a large number of small edits, particularly in CLI commands. It is fine not to cover these.

PR #942, opened 2026-06-26
Make SubscriberFilter extend Set and expose errors in EventSubscriber

Cooldude2606:
As title, subscriber filter now extends set to expose iterators and add. Then updated the snapshot to include any error which was rasied by the last network event, most commonly this will be permission errors.

### Changelog
```
### Changes
- SubscriptionFilter now directly extends Set<string>. #942
- EventSubscriber now exposes network errors, include permission errors. #942
```


Hornwitser (approved):
Looks good to me.

PR #943, opened 2026-06-26
OIDC Publish, Node version bump, and 2.1 tests working

Cooldude2606:
Node 20 has been EoL for a while now. Our support position is to test only LTS versions. All ci workflows have be bumped as required. [Node change log](https://nodejs.org/en/blog/announcements/v22-release-announce). Of note is native websocket module is no longer behind an experimental flag which may mean we no longer need `ws` as a dependency.

I have also added factorio 2.1.8 to the test matrix, but have not removed 1.1.110 because there is no stable 2.1 release yet.

Finally, I have enabled OIDC publishing for all packages. So now hopefully the CI will work as intended next release.

**This PR will be rebased and retested as needed to confirm 2.1 compatibility**

### Changelog
```
### Meta
- Node 20 is no longer officially supported as it went EoL in March 2026. Please update to v22 or v24.
```


Cooldude2606:
Current assumption is tests are failing due to #941 where 2.1 is not currently accepted within mod info json. Once 941 is merged this PR will be rebased and retested.

.github/workflows/publish.yml
Danielv123, line 46: Does this work when you have changed it to only have read permissions for the whole workflow above?

packages/lib/src/factorio/exchange_string.ts
```diff
@@ -427,6 +460,7 @@ export function readMapExchangeString(exchangeString: string) {
 	try {
 		const version = readVersion(state);
 		state.v2 = version[0] >= 2;
+		state.v2_1 = version[0] >= 2 && version[1] >= 1;
```
Danielv123, line 463: Is it confusing that both the v2 and v2_1 flag is set at the same time?

Danielv123 (approved):

.github/workflows/publish.yml
Cooldude2606: I have double checked and it does require `write` permissions. I will update. I did not realise that this included releases and not just commits / issues / prs.

packages/lib/src/factorio/exchange_string.ts
Cooldude2606: Maybe a bit, the intention is for it to be "greater than this version" as most changes are forward compatible. I will update the naming to `gt_v2_0` and `gt_v2_1`

PR #944, opened 2026-06-29
Add string encoding/decoding and update quick bar support

Cooldude2606:
This PR updates quick bar synchronization to support the new 2.1 API, including quality filters.

It also adds `encode_string` and `decode_string` to `compat`. These functions were originally introduced for quick bar slot serialization, but that approach was ultimately not used. They are still generally useful utilities, so they have been retained for future use.

Support for the additional quick bar slot types (`record`, `item`, and `remote`) has not been implemented in this PR. These slot types introduce significantly more complexity, but the updated data layout has been designed to remain forward-compatible, allowing support to be added in the future without requiring another format change.

### Changelog
```
### Changes
- Quick bar syncing now supports quality filters. #944
```

### DIFF
--- a/plugins/inventory_sync/module/serialize.lua
+++ b/plugins/inventory_sync/module/serialize.lua
@@ -197,7 +197,7 @@ function serialize.deserialize_personal_logistic_slots(player, serialized)
 end
 
 -- name is a custom type where quality is "normal" and comparator is "=" (most common case)
---- @alias QuickBarSlotEncoded.name { t: "n", n: string } 
+--- @alias QuickBarSlotEncoded.name string
 --- @alias QuickBarSlotEncoded.filter { t: "f", n: string, q: string, c: string }
 --- @alias QuickBarSlotEncoded QuickBarSlotEncoded.name | QuickBarSlotEncoded.filter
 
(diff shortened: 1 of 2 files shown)


plugins/inventory_sync/module/serialize.lua
```diff
@@ -194,6 +196,167 @@ function serialize.deserialize_personal_logistic_slots(player, serialized)
...
+--- @alias QuickBarSlotEncoded.name { t: "n", n: string } 
+--- @alias QuickBarSlotEncoded.filter { t: "f", n: string, q: string, c: string }
+--- @alias QuickBarSlotEncoded QuickBarSlotEncoded.name | QuickBarSlotEncoded.filter
+
+--- @param filter ItemFilter
+--- @return QuickBarSlotEncoded
+local function serialize_filter(filter)
+	if filter.quality == "normal" and filter.comparator == "=" then
+		return { t = "n", n = filter.name }
+	end
```
Hornwitser, line 209: You could potentially use `return filter.name` here and leave the simple string encoding intact for the simple case. That would reduce the raw payload size by 1.4 kb for a fully decked out hotbar with no quality filtered items on it.

Would of course also need to modify the return statement in the function below, the comment in deserialize_quick_bar_slot, and the types.

Hornwitser (approved):
I haven't tested the code, but it reads good. One thing to consider is to leave the string encoding intact to reduce the bytes sent over the wire.

PR #946, opened 2026-07-10
Add support for recommended mod dependencies

Cooldude2606:
Adds support for the new recommended mod dependecy type. Recommended dependencies will be downloaded and enabled if they can be satisfied but will be ignored if they are incompatible or unsatisfiable. All added and modified tests are passing locally but the suite will not pass due to other issues on the master branch.

There is one known edge case that may cause some confusion, although it is rare and easy to work around.

The dependency resolver does not maintain a full dependency tree. As a result, if a recommended dependency becomes unsatisfiable, any dependencies that were resolved through it may still be included in the final result, even though the recommended mod itself is omitted.

This only affects recommended dependencies and does not impact required dependency resolution. If the additional dependencies are not wanted, they can simply be disabled manually. Therefore, I consider this an acceptable appraoch at present.

### Changelog
```
### Changes
- Added support for recommended mod dependencies. #946
```

### DIFF
--- a/packages/controller/src/ControlConnection.ts
+++ b/packages/controller/src/ControlConnection.ts
@@ -675,6 +675,7 @@ export default class ControlConnection extends BaseConnection {
 		type ModPortalModType = InstanceType<typeof lib.ModPortalGetAllRequest.Response>["mods"][number];
 		type ModPortalReleaseType = NonNullable<ModPortalModType["releases"]>[number];
 		const dependencyRequirements = new Map<string, lib.ModVersionRange>();
+		const recommendedRequirements = new Map<string, lib.ModVersionRange>();
 		const optionalRequirements = new Map<string, lib.ModVersionRange>();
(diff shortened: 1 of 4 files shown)


Danielv123 (approved):
I am happy with this, but there is one thing I noticed you may want to fix - if a mod recommends a builtin like `quality`, it wont be enabled.

Cooldude2606:
> If a mod recommends a builtin like `quality`, it wont be enabled.

Neither will any mod that is already added to the mod back but is disabled.

I do not consider this a bug because it is not possible to gauge user intent for why a mod is disabled. They could have it intentionally disabled because they do not want to play with quality. Then it would be confussing behaviour to need to re-disable it after you add a new mod to the pack.

Perhaps there is room for better UX by displaying an "infomation" icon to indicate that it is recomended that the mod is enabled. And maybe this could be paired with other informative icons such as having an outdated version selected. But these are both out of scope of this PR because this would need to extend the existing error and warning icons that are displayed.

PR #947, opened 2026-07-11
Document Link method binding requirements

solarcloud7:
Link send and handle methods depend on their receiver. Document unsafe extraction forms and clarify that a directly invoked TypeScript cast retains its receiver.\n\n### Changelog\nNone. Documentation only.

### Changelog

None. Doc only change.

### DIFF
--- a/docs/writing-plugins.md
+++ b/docs/writing-plugins.md
@@ -455,6 +455,20 @@ async frobnicate() {
 }
 ```
 
+The send and handle methods depend on `this`, so call them directly on the connection object.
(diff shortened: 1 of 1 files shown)


Cooldude2606:
You should mention that `.bind` exists, and that this is most common when registering handlers. Such as `this.handle(MyEvent, this.handleMyEvent.bind(this))`

Cooldude2606:
Same comment as previously PR. Happy with conntent, needs to be worded more naturally and in line with existing content.

Cooldude2606 (approved):

PR #948, opened 2026-07-11
Warn against loading duplicate @clusterio libraries

solarcloud7:
Document how npm peer dependency installation can place a second copy of @clusterio/lib in a plugin, when the singleton guard fails, and how plugin developers can avoid the duplicate.\n\n### Changelog\nNone. Documentation only.

### DIFF
--- a/docs/writing-plugins.md
+++ b/docs/writing-plugins.md
@@ -43,6 +43,11 @@ After creating a directory for your plugin, the boilerplate can be generated aut
 To add it to `plugin-list.json` so that it gets loaded use the `plugin add <path>` sub-command to either clusteriocontroller, clusteriohost or clusterioctl.
 Note that it's important that the path starts with ./ or ../ (use .\ or ..\ on Windows).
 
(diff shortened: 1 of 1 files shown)


Cooldude2606:
This is written in a confussing way. Rather it should lead with the error and then explain why it happens, then end with the solution. This is because a dev normally would not intentionlly install it within the plugin and will only become aware of the issue when it fails the duplicate lib check.

Cooldude2606:
Im now happy with all the content here, but it is too obviously AIGen; the language needs to be more natural and matching the tone of the rest of the document.

solarcloud7:
I'd vote closing this as there is a fix that may be in the works:  https://github.com/clusterio/clusterio/pull/963

Cooldude2606:
963 is unrelated to duplicate clusterio libraries

Cooldude2606:
It would be create to get this merged before 2.0. If you still want this to be closed then thats fine also.

Issue #950, opened 2026-07-19
Prevent restart requests from performing unintended downgrades


Cooldude2606:
### Summary

When processing a **restart** request, detect whether the installed Clusterio version is older than the version currently running. If a downgrade is detected, reject the restart and require the controller / host to be stopped before the older version can be started manually.

### Motivation

Accidental downgrades can occur when package managers resolve an older Clusterio release due to runtime constraints. For example, users running Node.js 20 may receive an older release because newer versions require Node.js 22 or later.

Clusterio already records the version that is currently running in its configuration, while the installed version can be read from `package.json`. This makes it possible to detect a downgrade before a restart is performed.

### Proposed behaviour

When a **restart** request is issued:

1. Read the currently running version from the configuration.
2. Read the installed version from `package.json`.
3. If the installed version is older than the running version:

   * Reject the restart request.
   * Display a message explaining that the installed version is older than the running version.
   * Explain that downgrades require the controller / host to be stopped before the older version can be started.

No checks should be performed during startup.

### Why only check restart requests?

The purpose of this feature is to prevent **unintentional** downgrades without complicating the normal startup path.

Performing the version check during startup would add complexity to the startup flow and create unnecessary friction for service-managed installations, where the controller or hosts may be started automatically.

Instead, the safeguard is only applied to **restart** requests. This aligns with the existing administration model:

* A **restart** is intended for routine administration and allows a running controller or host to be restarted without requiring command-line access.
* A **stop** intentionally leaves the process offline. Starting it again requires command-line access, and the existing confirmation dialog already informs the user of this behaviour.
* Therefore, requiring a stop before a downgrade makes downgrading an explicit administrative action. Anyone who performs a stop either has the ability to start the process manually or is intentionally handing control to someone who does.
* A user who manually starts an older version is assumed to be intentionally performing the downgrade and understands its implications.

This approach protects against accidental downgrades in the common restart workflow while leaving manual startup and service-managed deployments unchanged.

### Benefits

* Prevents accidental downgrades during normal restart workflows.
* Leaves startup behaviour unchanged, avoiding complications for service-managed deployments.
* Requires deliberate user intervention before running an older version.
* Reuses the existing distinction between **restart** and **stop** operations without introducing additional permissions or configuration.

lntutor:
[...]

PR #951, opened 2026-07-19
Prevent restart requests from downgrading Clusterio

lntutor:
Implements #950.

Restart requests now compare the version recorded by the running controller or host with the currently installed package version. If the installed semantic version is older, the request is rejected with instructions to stop the process and start the older version manually. Equal versions and upgrades continue through the existing restart path, and startup remains unchanged.

Both controller and host read `package.json` at request time so package-manager changes made while the process is running are detected.

### Validation

- `pnpm run build`
- `pnpm run lint` (zero errors; four existing unused-disable warnings)
- focused controller/host restart tests: 10 passing
- `FAST_TEST=y pnpm test` with the bundled Factorio fixtures: 1,482 passing, 129 pending, 13 unrelated environment failures from incomplete Factorio fixtures and a network download timeout

### Changelog
```text
### Features
- Added a guard that prevents controller and host restart requests from unintentionally downgrading Clusterio. #950
```


Cooldude2606:
This LGTM although one quick change and discussion before merge. 1, can you update the pr desc to have this be a feature not fix. 2, semver feels like the correct solution, but we have an issue with "-alpha.22b" which is grater than "alpha-27" according to semver which is incorrect for us. We are currently discussing what the best solution for this would be over on our discord.

Cooldude2606:
From the discussion on discord, we will be making the next release 2.0 so this solution will be sufficient as it does not need to consider the broken alpha 22b release.

packages/host/src/Host.ts
```diff
@@ -2,6 +2,7 @@ import fs from "node:fs/promises";
 import path from "path";
 import events from "events";
 import pidusage from "pidusage";
+import semver from "semver";
```
Cooldude2606, line 5: The host package json has not been updated to include `semver`

packages/controller/src/ControlConnection.ts
```diff
@@ -189,6 +189,16 @@ export default class ControlConnection extends BaseConnection {
...
 			throw new lib.RequestError("Cannot restart, controller does not have a process monitor to restart it.");
 		}
+		const downgrade = await this._controller.checkRestartDowngrade();
+		if (downgrade) {
+			const { installedVersion, runningVersion } = downgrade;
+			throw new lib.RequestError(
+				`Cannot restart controller with installed Clusterio version ${installedVersion} because it ` +
+				"is older than " +
+				`running version ${runningVersion}. Stop the controller before starting the older ` +
+				"version manually."
```
Cooldude2606:
```suggestion
				`Cannot restart controller with installed Clusterio version ${installedVersion} because it ` +
				`is older than running version ${runningVersion}. ` +
				"Stop the controller before starting the older version manually."
```
It is better to break on start of a new sentence than how it is currently written.

test/controller/Controller.js
```diff
@@ -252,6 +252,57 @@ describe("controller/src/Controller", function() {
...
+				mockController.checkRestartDowngrade = async () => ({
+					installedVersion: "1.0.0",
+					runningVersion: "2.0.0",
+				});
+
+				await assert.rejects(restart, /Stop the controller before starting the older version manually/);
+				assert.equal(mockController.shouldRestart, false);
+				assert.equal(mockController.stopped, false);
+			});
+		});
```
Cooldude2606: Because this is a different class, it should be moved to its own new file `/tests/controller/ControlConnection.test.js`

Cooldude2606 (changes requested):

lntutor:
Updated the PR description and changelog section to classify this as a feature, as requested. The semver handling remains limited to the accepted 2.0 release scope.

lntutor:
@Cooldude2606 The requested PR-description classification is now updated to a feature, and the semver scope remains limited to the agreed 2.0 release. All existing checks are green. Could you please take another look when you have a moment?

packages/host/src/Host.ts
lntutor: Verified the current branch already declares `semver` and `@types/semver` in `packages/host/package.json`; no additional dependency change is needed.

packages/controller/src/ControlConnection.ts
lntutor: Updated the message to begin the manual-start instruction as a separate sentence.

test/controller/Controller.js
lntutor: Moved the ControlConnection restart tests into `test/controller/ControlConnection.test.js`; the new focused suite passes.

packages/host/src/Host.ts
Cooldude2606: Apologies, I had assumed it wasn't included and had been hoisted.

Cooldude2606 (approved):

PR #952, opened 2026-07-19
Show informational mod recommendations

### LINKED ISSUE #949: Improve mod status indicators with informational recommendations
### Summary

Extend the existing mod status indicator system to display informational notices. This would allow the UI to communicate non-critical recommendations, such as when a disabled mod is recommended by another enabled mod or when a newer compatible version is available.

### Motivation

The dependency resolver intentionally does not automatically enable mods that are already present in a mod pack but disabled, including built-in mods such as `quality`. (See #946)

This is deliberate because the resolver cannot determine user intent. A disabled mod may have been intentionally turned off, and automatically re-enabling it when adding another mod could be surprising behaviour.

However, the current UI provides no indication that a recommendation exists, making it easy for users to overlook.

### Proposed behaviour

Introduce a new informational status level alongside the existing errors and warnings. Unlike warnings, informational indicators would never imply that the mod pack is invalid. They simply communicate useful information to the user.

The existing status icon area already supports errors and warnings:

```tsx
<FileUnknownOutlined style={{ color: "#a61d24" }} />
<FileExclamationOutlined style={{ color: "#a61d24" }} />
<FileSyncOutlined style={{ color: "#dd5e14" }} />
```

This feature would extend that system with informational icons, using an appropriate informational colour (such as blue), for example:

```tsx
<LikeOutlined style={{ color: "#0958d9" }} />
<CloudSyncOutlined style={{ color: "#0958d9" }} />
```

Each icon would include a tooltip describing the recommendation, for example:

> This mod is recommended by another enabled mod but is currently disabled.

or

> A newer compatible version of this mod is available.

### Benefits

* Preserves existing behaviour by respecting user choices.
* Makes recommended dependencies discoverable.
* Avoids automatically enabling mods that users intentionally disabled.
* Creates a general framework for future informational messages instead of introducing one-off UI elements.

### Future extensions

Once an informational status category exists, it could also be used for other advisory messages without expanding the warning/error system, keeping the UI consistent as additional recommendations are introduced.

lntutor:
Adds reusable informational advisories to mod records and displays them alongside the existing error and warning indicators.

- Marks compatible disabled mods that are recommended by enabled mods.
- Records strictly newer compatible versions returned by Check for Updates.
- Shows both notices as non-blocking blue icons with descriptive tooltips.
- Clears stale update notices when the mod pack changes.
- Covers recommendation and update derivation with focused unit tests.

Closes #949

### Validation

- `pnpm build`
- `pnpm lint` (0 errors; 4 pre-existing warnings)
- `pnpm exec mocha --enable-source-maps --check-leaks -R spec test/lib/data/ModPack.js` (16 passing)

### Changelog
```text
### Features
- Mod packs now show informational notices for disabled recommended mods and available compatible updates. #949
```

### DIFF
--- a/packages/lib/src/data/ModPack.ts
+++ b/packages/lib/src/data/ModPack.ts
@@ -122,15 +122,10 @@ export function applyModRecordAdvisories(
 				continue;
 			}
 
-			if (!targetMod.advisories.some(
-				advisory => advisory.type === "recommended_dependency"
-					&& advisory.sourceModName === sourceMod.name
-			)) {
-				targetMod.advisories.push({
-					type: "recommended_dependency",
-					sourceModName: sourceMod.name,
-				});
-			}
+			targetMod.advisories.push({
+				type: "recommended_dependency",
+				sourceModName: sourceMod.name,
+			});
 		}
 	}
 
--- a/packages/lib/src/data/ModPack.ts
+++ b/packages/lib/src/data/ModPack.ts
@@ -12,7 +12,7 @@ import ModInfo, { ModDependencyUnsatisfiedReason } from "./ModInfo";
 
 import {
 	PartialVersion, PartialVersionSchema, integerPartialVersion,
-	SourceVersion, SourceVersionSchema, normaliseFullVersion,
+	SourceVersion, SourceVersionSchema, normaliseFullVersion, integerSourceVersion,
 } from "./version";
 
 
@@ -31,6 +31,10 @@ export interface ModSetting {
 	/** Value of the given mod setting. */
 	value: boolean | number | string | ModSettingColor;
 }
+
+export type ModRecordAdvisory =
+	| { type: "recommended_dependency", sourceModName: string }
+	| { type: "update_available", version: SourceVersion };
 
 const ModSettingJsonSchema = Type.Object({
 	"value": Type.Union([Type.Boolean(), Type.Number(), Type.String(), ModSettingColor]),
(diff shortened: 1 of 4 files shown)


packages/lib/src/data/ModPack.ts
```diff
@@ -56,6 +60,63 @@ export interface ModRecord {
...
+			const targetMod = modsByName.get(dependency.name);
+			if (
+				!targetMod
+				|| targetMod.enabled
+				|| dependency.version && !dependency.version.testVersion(targetMod.version)
+			) {
+				continue;
+			}
+
+			const targetAdvisories = targetMod.advisories ?? [];
```
Cooldude2606: This nor the check below are required. `modsByName` is constructed from `annotatedMods` which ensures the `advisories` field exists. For the check, a source mod will only ever be iterated once, and cant have duplicate dependencies, so this check will never fail.

packages/web_ui/src/components/ModPackViewPage.tsx
```diff
@@ -320,6 +322,16 @@ function DownloadDependenciesButton(props: DownloadDependenciesProps) {
 			lib.ModDependencyResolveRequest.fromModPackEnabled(props.modPack, props.checkForUpdates)
 		).then(response => {
 			if (canceled) { return; }
+			if (props.checkForUpdates && props.onUpdatesResolved) {
```
Cooldude2606: This doesn't achieve what is needed because when "check for updates" is enabled, the updates will be applied automatically and the advisory never seen. It is required for the advisory to appear before the button is pressed. At the moment this should be based on locally installed mod versions only.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ Many thanks to the following for contributing to this release:
 
 ### Features
 
+- Mod packs now show informational notices for disabled recommended mods and available compatible updates. [#949](<https://github.com/clusterio/clusterio/issues/949>)
```
Cooldude2606: Do not modify the change log directly.

Cooldude2606 (changes requested):

packages/lib/src/data/ModPack.ts
lntutor: Removed the redundant fallback and duplicate guard. The annotated local type now records that every clone has an advisory array.

packages/web_ui/src/components/ModPackViewPage.tsx
lntutor: Update advisories now derive from locally installed compatible `ModInfo` records during table rendering, before the button is pressed. The update dialog no longer drives advisory state.

CHANGELOG.md
lntutor: Removed the changelog entry; this change remains documented in the PR description and review history.

Cooldude2606 (approved):

PR #954, opened 2026-07-23
Redact credentials from URLs in download errors

### LINKED ISSUE #795: Factorio token exposed in logs
When a fetch to the mod portal fails, the factorio account token is exposed in the logs:
```
[info] Downloading mod warptorio-space-age_0.2.9.zip for Factorio 2.0 from portal.
[error] Error downloading file: Failed to fetch https://mods.factorio.com//download/warptorio-space-age/68bc7f318b01251f19df9426?username=cooldude2606&token=TOKEN_IS_HERE: 403 Forbidden
[error] Failed to download or load mod warptorio-space-age_0.2.9.zip (0.2.9): Failed to fetch https://mods.factorio.com//download/warptorio-space-age/68bc7f318b01251f19df9426?username=cooldude2606&token=TOKEN_IS_HERE: 403 Forbidden
[error] Error downloading mod warptorio-space-age_0.2.9.zip from portal: Failed to find mod warptorio-space-age_0.2.9.zip in store after download attempt.
```
(Example above intentionally creates a 403 by providing the wrong token)

bbassie:
Closes #795.

`ModStore` downloads mods from the portal by putting the Factorio account username and token into the download URL as query parameters. When `downloadFile` gets a non-ok response it interpolates the whole URL into its error message, and `ModStore.downloadModFromReleases` logs that message at error level — so any failed mod download writes the account token into the log verbatim.

This adds a `redactUrl` helper to `lib/helpers` that replaces the value of credential-carrying query parameters (`token`, `password`, `secret`, `api_key`, `apikey`, `access_token`) with `REDACTED`, and applies it where `downloadFile` builds its error message. Redacting at the point the URL is turned into text means any future caller passing a credentialed URL is covered too, rather than relying on each log site to remember.

The parameter names and the rest of the URL are kept intact so the message still says which request failed. `username` is deliberately *not* redacted: it is not a secret, and it is useful for diagnosing "wrong account configured" cases — which is the most common cause of the 403 in the issue.

Tests cover `redactUrl` directly (case-insensitive matching, repeated parameters, `URL` objects, non-credential URLs, unparsable input) and add a regression test asserting `downloadFile`'s error never contains the token value.

### Changelog
```
### Fixes
- Factorio account token no longer written to the log when a mod download fails. #795
```

### DIFF
--- a/test/lib/helpers.js
+++ b/test/lib/helpers.js
@@ -236,6 +236,45 @@ describe("lib/helpers", function() {
 		});
 	});
 
+	describe("redactUrl()", function() {
+		it("should redact the value of credential query parameters", function() {
+			assert.equal(
+				lib.redactUrl("https://mods.factorio.com/download/mod?username=user&token=secret"),
+				"https://mods.factorio.com/download/mod?username=user&token=REDACTED"
+			);
+			for (const name of ["access_token", "api_key", "apikey", "password", "secret", "token"]) {
+				assert.equal(
+					lib.redactUrl(`https://example.com/?${name}=secret`),
+					`https://example.com/?${name}=REDACTED`
+				);
+			}
+		});
+		it("should match credential parameter names case insensitively", function() {
+			assert.equal(lib.redactUrl("https://example.com/?Token=secret"), "https://example.com/?Token=REDACTED");
+			assert.equal(lib.redactUrl("https://example.com/?API_KEY=secret"), "https://example.com/?API_KEY=REDACTED");
+		});
+		it("should redact every occurrence of a repeated credential parameter", function() {
+			assert.equal(
+				lib.redactUrl("https://example.com/?token=first&token=second"),
+				"https://example.com/?token=REDACTED"
+			);
+		});
+		it("should accept a URL object", function() {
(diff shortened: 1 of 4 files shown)


test/lib/helpers.js
```diff
@@ -236,6 +236,45 @@ describe("lib/helpers", function() {
...
+		it("should match credential parameter names case insensitively", function() {
+			assert.equal(lib.redactUrl("https://example.com/?Token=secret"), "https://example.com/?Token=REDACTED");
+			assert.equal(lib.redactUrl("https://example.com/?API_KEY=secret"), "https://example.com/?API_KEY=REDACTED");
+		});
+		it("should redact every occurrence of a repeated credential parameter", function() {
+			assert.equal(
+				lib.redactUrl("https://example.com/?token=first&token=second"),
+				"https://example.com/?token=REDACTED"
+			);
+		});
```
Cooldude2606, line 261: It is rare to have repeated credential parameters, which means a repeat is likely an error. The resulting string needs to expose that multiple occurrences were present so this error is not hidden. This is a simple enough change so I will update then merge.

PR #955, opened 2026-07-23
Fix the instance console hiding Factorio diagnostics

### LINKED ISSUE #853: Instance console confusingly filters out Factorio log
By default the console view for an instance filters out log messages that might be vital to diagnose issues. This is toggled by a confusingly labelled toggle buttons that just says "chat" on it and defaults to on:

Image

For example you might input a command with a typo in it like `/command rcno.print("alive")`, run it and then be confused when it prints the command but no output:

Image

There's nothing about a toggle labled "chat" being on that suggest vital diagnostic have been hidden.

bbassie:
Closes #853.

Three separate commits, each can be dropped independently.

**Making the filter discoverable.** The control was a `Switch` with `checkedChildren="Chat"` / `unCheckedChildren="Log"`, which renders only the label of the mode you are already in. Defaulting to the filtered mode, it read as a button labelled "Chat" with nothing to suggest another mode existed. It is now a `Segmented` control listing both modes, so the one you are not in is always visible, with a tooltip describing what each covers. Per review the modes are named **Game** and **Instance** rather than Chat / Log.

**Not swallowing diagnostics.** The filter kept Factorio output only when it parsed as an action, which dropped the server's *replies* to those actions. The example in the issue — a typo'd `/command` echoing and then printing nothing — happens because `Cannot execute command. Error: ...` is a plain message, not an `[ACTION]` line. Factorio's own engine-level errors were dropped the same way, so e.g. a failed multiplayer connection left no trace in either mode.

Rather than special-casing command replies, the filter now selects by *where a line came from*. Factorio interleaves two streams on stdout: the date-stamped console log carrying what was done to the server and the replies to it, and the seconds-stamped engine log which is verbose diagnostic output. Game mode keeps the console log whole and takes only warnings and errors from the engine log. The same rule is applied to Clusterio's own entries, which were previously kept in full and drowned the Factorio output they were mixed with — those are now kept at `warn` or worse.

I replayed both the old and new predicates over a development cluster's full log history for one instance (5271 entries at the default max level). Game mode goes from 781 lines to 605: it loses 415 purely informational Clusterio entries, and gains the command replies plus engine errors like `0.911 Error ServerMultiplayerManager.cpp:743: Matching server connection failed` that were previously invisible in *both* modes.

**Labelling Factorio's output.** Also per review: the console rendered `server` level entries as bare parsed Factorio output while every other level was prefixed with its name, so nothing in a line distinguished what the game wrote from what Clusterio wrote about it. Those entries now carry the `[server]` prefix too, matching what `TerminalFormat` already does for the host and controller logs, styled with the same bold the terminal colourises the level with. This applies to every console using `LogConsole`, not just the instance one, which I read as the intent of "any server level log messages".

Two things I would still like your call on:

- **The default is unchanged.** The issue opens with "by default the console view for an instance filters out log messages that might be vital", and I have read that as being about the labelling and the over-filtering rather than a request to flip the default. Happy to flip it if that was the intent.
- **No test covers this.** `packages/web_ui` has no test directory, so `passesActionsOnly` is verified only by manual checking and the log replay above. It is a pure function and would be straightforward to test if you would like a web_ui test setup added — but that felt like a separate change.

### Changelog
```
### Changes
- Instance console filter is now a labelled Game/Instance control instead of a single-label toggle, and Factorio's output in the log consoles is prefixed with `[server]` like every other log level. #853

### Fixes
- Instance console no longer hides the Factorio server's replies to console commands, or Factorio's own errors, when filtering to game activity. #853
```

### DIFF
--- a/packages/web_ui/src/components/InstanceViewPage.tsx
+++ b/packages/web_ui/src/components/InstanceViewPage.tsx
@@ -1,7 +1,8 @@
 import React, { useContext, useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
 import {
-	Alert, Button, Descriptions, Dropdown, Flex, MenuProps, Modal, Space, Spin, Switch, Typography, message,
+	Alert, Button, Descriptions, Dropdown, Flex, MenuProps, Modal, Segmented, Space, Spin, Tooltip, Typography,
+	message,
 } from "antd";
 import CopyOutlined from "@ant-design/icons/CopyOutlined";
 import DeleteOutlined from "@ant-design/icons/DeleteOutlined";
@@ -34,7 +35,7 @@ import { MetricRelativeDate } from "./system_metrics";
 type MenuItem = Required<MenuProps>["items"][number];
 const { Title } = Typography;
 
-// Remembers the console "Chat/Log" toggle across reloads (defaults to on).
+// Remembers the console "Chat/Log" selection across reloads (defaults to chat).
 const consoleActionsOnlyKey = "instance-console-actions-only";
 
 type InstanceDescriptionProps = {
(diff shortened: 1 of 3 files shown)


Cooldude2606:
Can you make your comments more concise, a single line explaining why a check exists in `passesActionsOnly` would be sufficent. Also I think part of the confussion is resulting from the names of the sections, so lets change it from Chat / Log to Game / Instance. We should then have any "server" level log messages be prefixed correctly with `[server]`.

bbassie:
All three done, force-pushed as amendments to the existing commits plus one new commit.

Comments in `passesActionsOnly` are down to one line each, saying why that check is there and nothing else. Sections are now Game / Instance, in the control, the tooltip and the comment on the localStorage key. The `[server]` prefix is a separate commit: `formatLog` rendered `server` level entries as bare parsed output while every other level got its name prefixed, so it now emits the prefix for those too and only swaps the message body for the parsed rendering. That matches what `TerminalFormat` already does in `logging_utils.ts`, and I added a `.log-server` rule with the same bold the terminal colouriser uses, since no class existed for that level. It applies to every console built on `LogConsole`, not just the instance one — I read that as the intent of "any server level log messages", but say the word if you wanted it scoped to the instance console.

I left the `actionsOnly` prop, the `passesActionsOnly` name and the `instance-console-actions-only` storage key alone. Renaming the key would silently reset everyone's remembered choice, and the rest is internal, so it seemed better kept out of a naming change.

One thing worth flagging while the labels are open: Instance is a strict superset of Game — every line Game shows, Instance also shows. A two-way control reads as two disjoint categories, which is a little at odds with that, and it is arguably the same shape of confusion Chat / Log had. Something like **Game / All** would keep the Game half and say outright that the other side is just unfiltered. Entirely your call, and Game / Instance is in the branch as asked either way.

Cooldude2606 (approved):

PR #956, opened 2026-07-23
Reload the web interface when the controller's plugins change

### LINKED ISSUE #729: Error shown when adding a plugin to the controller
If a plugin is added to the controller and then the controller is restarted, the plugin list in the Web UI shows "Error loading module". This is incorrect as no attempt was made at loading the plugin. Instead the Web UI should indicate it has not been loaded and entice the user to refresh the page to load the plugin.

image

### LINKED ISSUE #804: Web UI not forced to refresh after controller plugin changes
## Description:
When controller plugins are added, removed, toggled (enabled/disabled), or when the controller enters recovery mode, connected Web UI clients are not forced to refresh their cached version. As a result, the Web UI continues operating as if the plugin is still enabled and sends events that the controller no longer expects, leading to inconsistent state or crashes.

## Proposed Fix:
Implement a mechanism to invalidate or force-refresh connected Web UI caches whenever plugin configuration changes or recovery mode is activated, ensuring the client and controller remain in sync.

bbassie:
Closes #804. Closes #729.

The web interface loads plugin bundles, calls `registerPluginMessages` and `addPluginConfigFields` once in `bootstrap()`, and has no way to undo any of it. When the controller is restarted to pick up a plugin being installed, enabled or disabled, open pages reconnect to it and carry on as though the old set were still loaded, sending messages the controller no longer handles. Starting the controller in recovery mode has the same effect from the other direction: `Controller.ts` force-disables every plugin, and since only `registerHost` refuses connections in recovery mode, control connections are still accepted — so the browser connects happily to a controller running none of the plugins it loaded.

The controller already states the plugins it is running in the `hello` handshake, built from `controller.plugins` in `WsServer.handleConnection` and sent on every WebSocket connection. Hosts have used this for a long time — `Host` keeps it as `serverPlugins` and consults it to decide which events to forward and which instance plugins to load — and the web interface was the only client ignoring it. The first commit compares it against the set the page loaded, and on a mismatch stops rendering and asks to be reloaded rather than acting on plugins that are no longer there.

Since the handshake is sent on every connection, this does not rest on a session never surviving a controller restart, and it needs no protocol change, no new endpoint, and no extra request on reconnect.

The comparison is against `loaded` rather than `enabled`, taken from the `/api/plugins` response bootstrap already fetches. `enabled` is `loaded && config.get(load_plugin)`, so a `load_plugin` change that has not been restarted into yet would otherwise latch a mismatch the page could never clear.

A plugin rebuilt in place without its version changing is not covered here, since the handshake reports names and versions. That is the web asset fingerprint discussed in the thread below, which follows as a separate PR once this lands.

The second commit fixes #729, which is the same situation seen from the plugin list. That page reported "Error loading module" for anything without a loaded package, which includes a plugin the controller picked up after the page was opened — nothing had been attempted for it, so reporting a failure was wrong, and the remedy is a reload rather than an investigation. Note that a plugin the controller cannot serve at all is dropped during startup *before* it reaches the loaded plugin map, so absence from that map alone does not mean "new"; the web error is checked first and still reports as an error. The first commit makes that state hard to reach in practice, since the reload card intercepts it before the plugin list can be opened, but the label is wrong on its own terms and the check is three lines.

### Testing

Against a running dev cluster of 14 plugins, reading the plugin set off the handshake on each restart:

| Controller | plugins reported | reload card |
| --- | --- | --- |
| baseline | 14 | — |
| restarted, nothing changed | 14 | **no** card, reconnected silently |
| restarted into `--recovery` | 0 | card shown |
| Reload clicked | 14 | re-bootstraps to the normal UI |

The second row is the one that matters — without it the third would only be showing that any reconnect prompts a reload.

`packages/web_ui` still has no test directory, so this is covered by the runs above. The follow-up PR puts its fingerprint in `packages/controller`, where it is unit tested.

### Changelog
```
### Fixes
- Web interface now asks to be reloaded when the controller's plugins change, instead of continuing against plugins that are no longer loaded. #804
- Plugin list no longer reports a plugin added since the page was opened as having failed to load. #729
```

### DIFF
--- a/packages/web_ui/src/bootstrap.tsx
+++ b/packages/web_ui/src/bootstrap.tsx
@@ -11,6 +11,7 @@ import InputRole from "./components/InputRole";
 import InputModPack from "./components/InputModPack";
 import { InputTargetVersion, InputPartialVersion, InputFullVersion } from "./components/InputVersion";
 import { Control, ControlConnector } from "./util/websocket";
+import { fetchPluginSet, pluginSetFingerprint } from "./util/pluginSet";
 
 const { ConsoleTransport, WebConsoleFormat, logger } = lib;
 
@@ -29,16 +30,16 @@ async function loadScript(url: string) {
 	return result;
 }
 
-async function loadPluginInfos(): Promise<lib.PluginWebpackEnvInfo[]> {
-	let response = await fetch(`${webRoot}api/plugins`);
+async function loadPluginInfos(): Promise<[lib.PluginWebpackEnvInfo[], string]> {
 	let pluginList: lib.PluginWebApi[];
-	if (response.ok) {
-		pluginList = await response.json();
+	try {
+		pluginList = await fetchPluginSet();
 
-	} else {
-		logger.error("Failed to get plugin data, running without plugins");
+	} catch (err: any) {
+		logger.error(`Failed to get plugin data, running without plugins: ${err.message}`);
 		pluginList = [];
 	}
+	const fingerprint = pluginSetFingerprint(pluginList);
 
 	let pluginInfos: lib.PluginWebpackEnvInfo[] = [];
 	await __webpack_init_sharing__("default");
@@ -73,7 +74,7 @@ async function loadPluginInfos(): Promise<lib.PluginWebpackEnvInfo[]> {
 			}
 		}
 	}
-	return pluginInfos;
+	return [pluginInfos, fingerprint];
 }
 
 async function loadPlugins(pluginInfos: lib.PluginWebpackEnvInfo[], control: Control) {
@@ -132,13 +133,14 @@ export default async function bootstrap() {
 		level: "verbose",
 		format: new WebConsoleFormat(),
 	}));
-	let pluginInfos = await loadPluginInfos();
+	let [pluginInfos, pluginFingerprint] = await loadPluginInfos();
 	lib.registerPluginMessages(pluginInfos);
 	lib.addPluginConfigFields(pluginInfos);
 
 	let wsUrl = new URL(webRoot, document.location.href);
 	let controlConnector = new ControlConnector(wsUrl.href, 120);
 	let control = new Control(controlConnector, new Map(pluginInfos.map(p => [p.name, p])));
+	control.pluginFingerprint = pluginFingerprint;
 	control.plugins = await loadPlugins(pluginInfos, control);
 	control.inputComponents = inputComponentsFromPlugins(control.plugins);
 
--- a/packages/web_ui/src/bootstrap.tsx
+++ b/packages/web_ui/src/bootstrap.tsx
@@ -11,6 +11,7 @@ import InputRole from "./components/InputRole";
 import InputModPack from "./components/InputModPack";
 import { InputTargetVersion, InputPartialVersion, InputFullVersion } from "./components/InputVersion";
 import { Control, ControlConnector } from "./util/websocket";
+import { loadedPluginSetKey } from "./util/pluginSet";
 
 const { ConsoleTransport, WebConsoleFormat, logger } = lib;
 
@@ -29,7 +30,7 @@ async function loadScript(url: string) {
 	return result;
 }
 
-async function loadPluginInfos(): Promise<lib.PluginWebpackEnvInfo[]> {
+async function loadPluginInfos(): Promise<[lib.PluginWebpackEnvInfo[], string]> {
 	let response = await fetch(`${webRoot}api/plugins`);
 	let pluginList: lib.PluginWebApi[];
 	if (response.ok) {
@@ -39,6 +40,7 @@ async function loadPluginInfos(): Promise<lib.PluginWebpackEnvInfo[]> {
 		logger.error("Failed to get plugin data, running without plugins");
 		pluginList = [];
 	}
+	let pluginSetKey = loadedPluginSetKey(pluginList);
 
 	let pluginInfos: lib.PluginWebpackEnvInfo[] = [];
 	await __webpack_init_sharing__("default");
@@ -73,7 +75,7 @@ async function loadPluginInfos(): Promise<lib.PluginWebpackEnvInfo[]> {
 			}
 		}
 	}
-	return pluginInfos;
+	return [pluginInfos, pluginSetKey];
 }
 
 async function loadPlugins(pluginInfos: lib.PluginWebpackEnvInfo[], control: Control) {
@@ -132,13 +134,13 @@ export default async function bootstrap() {
 		level: "verbose",
 		format: new WebConsoleFormat(),
 	}));
-	let pluginInfos = await loadPluginInfos();
+	let [pluginInfos, pluginSetKey] = await loadPluginInfos();
 	lib.registerPluginMessages(pluginInfos);
 	lib.addPluginConfigFields(pluginInfos);
 
 	let wsUrl = new URL(webRoot, document.location.href);
 	let controlConnector = new ControlConnector(wsUrl.href, 120);
-	let control = new Control(controlConnector, new Map(pluginInfos.map(p => [p.name, p])));
+	let control = new Control(controlConnector, new Map(pluginInfos.map(p => [p.name, p])), pluginSetKey);
 	control.plugins = await loadPlugins(pluginInfos, control);
 	control.inputComponents = inputComponentsFromPlugins(control.plugins);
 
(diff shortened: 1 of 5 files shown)


Cooldude2606:
This looks like a great framework which can also include the web ui core. For example, when the controller is updated to a new version and the content of `controller\dist\web\static\main` is changed, we should prompt to reload. Could you please look into how this would fit into your feature.

bbassie:
Yes, this fits — and the core bundle is the easier half, because it is already content hashed the same way plugin bundles are.

The mechanism would be: the *loaded* value comes from the DOM (`#script-bundle`'s `src`), which is what the browser actually executed, so there is no build-time constant and no race if the controller is updated between the HTML being served and bootstrap running. The *current* value has to come from the controller, and that is where there is a design choice I would rather you make than guess at.

**1. Consolidate into one endpoint.** Something like `/api/cluster-info` returning `{ name, webBundle, plugins }`, replacing `/api/cluster-name` and `/api/plugins`. This is the only option that *reduces* requests rather than adding one, and it is a natural home for anything else describing what the controller is serving. It is a breaking change to two documented endpoints, and it makes `LoginForm` pull the whole payload when it only wants the cluster name. Worth noting `/api/plugins` is currently fetched in five separate places across the UI (bootstrap, PluginsPage, PluginViewPage, HostsPage, AboutModal), so a single endpoint plus a client-side cache would remove more redundancy than just this feature's share.

**2. A small dedicated endpoint,** e.g. `/api/web-bundle` returning `{ main }`, resolved from `app.locals.mainBundle` with the same dev-mode webpack fallback `serveWeb` already uses. No breaking change, but it makes three endpoints where the client already fetches two, which is the wrong direction if consolidation is on the table.

**3. Parse `index.html` on reconnect.** No server change at all: fetch the page, read `#script-bundle` from it with `DOMParser`. `serveWeb` already resolves the bundle name for both production and `--dev`, so this gets the dev path for free. `index.html` is ~1.5 KB, smaller than the `/api/plugins` response, so the transfer cost is not the objection I first assumed it was.

My lean: if you are open to the breaking change, **1** is the better end state and I am happy to do it. If you would rather not reshape the API right now, **3** is the option that adds no API surface at all and leaves the consolidation decision open for later — at which point the bundle check moves to the new endpoint with a one-line change.

Two smaller things either way:

The card currently says the *plugins* changed. A core update needs different wording, so I would track which of the two differs and vary the sentence.

In `--dev`, webpack rebuilds on every edit and the main bundle hash changes each time, so the card would appear often while working on the web UI. Same tension as the plugin bundles, which I flagged in the description. My inclination is to accept it since the page genuinely is running stale code, but it is more noticeable for core than for plugins and I will follow your preference.

Finally, scope: this is new work on top of what #804 asked for, so I would suggest landing it as a separate PR once this one is merged, unless you would rather see them together.

Cooldude2606:
Happy for the core part to be a separate PR, but we should make sure the architecture is aligned so we don't end up rewriting it later. Also, in development mode it is expected that this appears after every edit, since a reload is required for changes to take effect.

For the implementation, parsing the `index.html` DOM feels like the simplest approach. It also reuses the logic we will require for obtaining the loaded value on the initial fetch, which, as you mentioned, is important for avoiding race conditions.

If we are already parsing the DOM, we could extend it to expose additional metadata, such as a fingerprint hash representing the combined state of core and its plugins. This would be an opaque value that the client can compare quickly to determine whether a reload is required. It only needs to be computed once on controller startup (except in development mode), since changes to a plugin's enabled state do not affect whether a reload is necessary.

This approach would also allow us to use the complete manifest rather than only the main bundle, giving us broader coverage of the package.

bbassie:
I went looking at how the core bundle would fit and found that the handshake already does half of this.

`WsServer.handleConnection` sends `MessageHello` on every connection with `{ version, plugins }`, where `plugins` is built from `controller.plugins` — the plugins actually loaded. `Host` already keeps that as `serverPlugins` and uses it to decide what to forward and which instance plugins to load. So "which plugins is the controller running" is already in the protocol and already load-bearing; the web interface was the only client not using it. I have rewritten the first commit to listen to the handshake instead of re-fetching `/api/plugins`, which drops the fingerprint helper, the extra request on reconnect, and the dependency on a session never surviving a restart — the handshake is sent on every connection, resumed or not.

That leaves the part manifests genuinely have to answer, and I would keep it as a second value rather than folding both into one opaque hash. `Controller.loadPlugins` loads each plugin's `dist/web/manifest.json` *before* the `load_plugin` and recovery-mode `continue`s, so the manifests are byte-identical in recovery mode. Measured against my dev cluster:

| Controller | hello.plugins | webFingerprint |
| --- | --- | --- |
| normal | 14 | 45344af103215258 |
| restarted, nothing changed | 14 | 45344af103215258 |
| `--recovery` | 0 | 45344af103215258 |
| core bundle rebuilt | 14 | f80cba708f306092 |

Row three is the one that matters. A fingerprint over the complete manifests cannot see recovery mode, so folding the plugin set into it would leave #804 open — which is also why I would not treat enabled state as irrelevant to whether a reload is needed. Keeping the two apart additionally lets the card name which of them changed, which a single opaque value cannot.

The shape I have implemented, with the core half on a branch stacked on this one:

**Which plugins are running** — `hello.plugins`, no server change at all. The page's baseline comes from the `/api/plugins` response bootstrap already fetches, compared on `loaded` rather than `enabled`, since `enabled` also folds in `load_plugin`, which only applies from the next restart and would otherwise latch a permanent mismatch.

**What is being served** — sha256 over the complete core manifest plus every plugin manifest, computed once at startup and per call under `--dev`, where webpack's compilation hash stands in for the manifest. This is your "complete manifest rather than only the main bundle". Content hashing also means an identical rebuild yields an identical value, so redeploying unchanged assets does not prompt anyone.

On where the *current* value comes from, I ended up not re-fetching `index.html`. The value the page *loaded* still has to be stamped into the document — the browser knows the entry points it loaded but not the complete manifest, and stamping it closes the race if the controller is updated between the HTML being served and bootstrap running, which was your point. But for the current value the handshake is strictly better: it arrives with the connection, so there is no second request to fail — the description already notes that a failed `/api/plugins` re-fetch leaves no card — and no cache or reverse proxy in front of `index.html` to serve a stale copy. It is one optional field alongside `version` and `plugins`, which are the same kind of statement about the running controller. It degrades cleanly both ways: a newer interface against an older controller sees no fingerprint and skips that half, and hosts and ctl ignore the field.

`webFingerprint` also lands in `packages/controller`, which has a test suite, so the "no automated test" question from the description answers itself — six cases cover plugin ordering, a core change, a plugin change, add/remove, an unservable plugin, and that it is computed once outside dev mode.

I have not pushed the rewritten first commit yet, since it changes the mechanism you reviewed and I would rather have your read on the shape first. Happy to move the current value back onto HTTP if you prefer it there — it is a small change on the client either way.

Cooldude2606:
On the stacking, please wait until this PR has been merged into master before you open the second PR. This will avoid needing to rebase.

Which plugins are running - This looks good, happy for the handshake to be used alongside the plugins api endpoint. And correct to exclude `enabled` as I had previously mentioned.

What is being served - This looks correct and helps prevent unnecessary reloads. And happy for you to do as described so that you can name which plugins changed and even better that we found a way to do it without modifing the html. Amd good job being able to impliment test cases for this.

Please push your changes for my review.

bbassie:
Pushed, and understood on holding the second PR until this one is merged. I have rewritten the description too, since it still described the old mechanism.

One correction so it does not surprise you in review later: the follow-up does still add one line to `index.html`, a `<meta name="clusterio-web-fingerprint">` that `serveWeb` fills in next to `__MAIN_BUNDLE__`. What it avoids is *parsing* the HTML at runtime — the page reads that tag once at bootstrap and the current value then rides the handshake, so there is no re-fetch of `index.html` on reconnect. The tag is what makes the loaded value describe the assets the browser was actually served, which is what closes the race if the controller is updated between the HTML being served and bootstrap running.

Also to set expectations on the card wording: it names whether the core interface or the plugin set changed, not which individual plugin. Per-plugin detail is already available from `/api/plugins` if we ever want it on that screen, but the reload decision itself does not need it.

Cooldude2606 (approved):

Cooldude2606:
There is on issue with plugins still sending events on reconnect (such as subscription requests) but this will be hard to fix and not in scope for this PR. This should be revisited later, if it is even worth it as the only effect is getting some extra log lines.

PR #958, opened 2026-07-23
Stop addressing subscription events to a request id

### LINKED ISSUE #847: Subscriptions logic send events to request addresses
The addresses in the link layer consists of 2 or 3 components: The type of address, the id for the type and an optional request id. The request id component is generated and set when sending a request so that if the same source sends multiple of the same request type responses get routed back to the correct requester in the code even if they arrive out of order. These request ids are not intended to be used with events.

While investigating another issue I discoreverd that events sent out by the subscription system have request ids set in the address.

```
WS out {"type":"event","seq":559,"src":[0,0],"dst":[3,3,213],"name":"UserUpdatesEvent","data":{"updates":[{"name":"player2","roles":[1],"updated_at_ms":1772456328905}]}}
                                                         ^
                              requestId of the request the subscription was made with.
```

This might cause issues in the future, especially if any code looks at the subscription destination. Addresses with requestId are considered distinct from addresses without them, and don't compare equal to same destination places with a different requestId. Given that the requestId updates every time the consumer changes its subscription that's a recipe for surprises.

bbassie:
Closes #847.

A subscribe arrives as a request, so the `src` address `SubscriptionController.handleRequest` is called with carries the request id used to route the response back to the caller. That address was stored verbatim as the subscription's destination:

```ts
eventData.subscriptions.set(addressIndex, { link, dst: src, filters: request.filters });
//                                               ^^^ keeps the request id forever
...
link.sendTo(src, eventReplay);   // the replay event goes to it too
```

So every event broadcast to that subscriber, and the replay event sent when the subscription is created, went out addressed to a request that had long since been answered — the `"dst":[3,3,213]` in your WS dump.

A request id identifies one request in flight and has no meaning on an event. `link.ts` keys response routing off it (`_forwardedRequests.delete(message.dst.requestIndex())` when a message is not addressed to the receiver), so events carrying one get put through paths meant for response traffic.

The fix strips the request id when taking the destination from the request that created the subscription, via a new `Address.withoutRequestId()` since the same stripped address is wanted in three places.

**One correction to your diagnosis:** you pointed at line 219, but `subscriptions.ts` was rewritten in the batch that landed with alpha.27 — the sites are now 361, 374 and 387.

**Good news on the blast radius:** the subscription map is keyed on `addressIndex()`, which is `` `${type}:${id}` `` and already excludes the request id. So resubscribing never created duplicate entries and no subscription state was corrupted; the only thing wrong was the address events were sent to.

### Testing

Three tests added under a new `subscriber addressing` block, covering the destination stored on `subscribe`, the destination stored when `replace` creates the subscription, and the replay event. Each asserts the sent message's `dst` has no `requestId` and equals the plain subscriber address.

They need their own `SubscriptionController`: a subscription is only addressed when it is first created, and the existing `.broadcast()` setup has already subscribed at the same `addressIndex`, so a request made against it takes the "subscriber already exists" branch that only updates filters. My first attempt at these tests passed against unfixed code for exactly that reason — worth knowing if you extend them.

Confirmed all three fail against unmodified `subscriptions.ts` and pass with the fix. Full `test/lib` and `test/messages` run clean at 1020 passing.

### Changelog
```
### Fixes
- Events sent to subscribers are no longer addressed to the request id of the request that created the subscription. #847
```

### DIFF
--- a/packages/lib/src/data/messages_core.ts
+++ b/packages/lib/src/data/messages_core.ts
@@ -115,6 +115,24 @@ export class Address {
 	}
 
 	/**
+	 * Returns this address with any request id removed
+	 *
+	 * A request id identifies one request in flight and is set on the
+	 * source address of a request so that its response can be matched back
+	 * up with it.  It is meaningless on anything else, so an address taken
+	 * from a request has to be stripped of it before it is used as the
+	 * destination of an event.
+	 *
+	 * @returns this address without a request id.
+	 */
+	withoutRequestId() {
+		if (this.requestId === undefined) {
+			return this;
+		}
+		return new Address(this.type, this.id);
+	}
+
+	/**
 	 * Returns true if this address targets the given destination
 	 * @param dst - Destination to check
 	 * @returns true if this address is addressed to the given destination
--- a/packages/lib/src/subscriptions.ts
+++ b/packages/lib/src/subscriptions.ts
@@ -345,6 +345,9 @@ export class SubscriptionController {
 		}
 		const addressIndex = src.addressIndex();
 		const subscriber = eventData.subscriptions.get(addressIndex);
+		// src is the source of a request and carries its request id, which
+		// has no meaning on the events sent back to the subscriber.
+		const subscriberDst = src.withoutRequestId();
 		switch (request.action) {
 			case "unsubscribe":
 				if (!subscriber) {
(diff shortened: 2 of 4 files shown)


Cooldude2606 (approved):
The change is good. I have made the following changes before merge:
1) Used this PR as an opportunity to rename the internal `subscriber.dst` to `subscriber.address`
2) Moved the newly added tests under `.handleRequest()`
3) Added tests for new `.withoutRequestId()`

PR #963, opened 2026-07-26
Discover transitive NPM plugins

### LINKED ISSUE #953: Detect plugins through transitive dependencies during automatic discovery
### Summary

Extend the automatic plugin discovery mechanism to recursively scan the dependencies of detected plugins for additional plugins. Recursion should only continue through valid Clusterio plugins, while non-plugin packages should be ignored.

### Motivation

Plugin discovery currently only considers the initial set of installed plugins. This means plugins that are installed as dependencies of other plugins are not automatically discovered unless they are also installed explicitly.

Supporting transitive plugin discovery would allow plugin packages to depend on other plugins without requiring users to install or register every plugin manually.

### Proposed behaviour

When scanning for plugins:

1. Discover the initial set of installed plugins using the existing mechanism.
2. For each valid Clusterio plugin:

   * Inspect its package dependencies.
   * Check whether each dependency is itself a valid Clusterio plugin.
   * If it is, register it and continue scanning its dependencies.
3. If a dependency is not a valid Clusterio plugin:

   * Ignore it.
   * Do not recurse into its dependencies.
4. Track visited packages to prevent duplicate processing and avoid infinite loops caused by circular dependencies.

The result is that the discovery process traverses the graph of **Clusterio plugins**, rather than the entire `node_modules` dependency tree.

### Why only recurse through plugins?

Most package dependencies are ordinary npm libraries and cannot contain discoverable Clusterio plugins. Recursing into every package would significantly increase discovery time while providing little benefit.

By only recursing through packages that are confirmed to be valid Clusterio plugins, discovery remains efficient while still supporting plugin-to-plugin dependencies.

### Benefits

* Allows plugins to declare dependencies on other plugins.
* Removes the need for users to explicitly install or register every plugin in a dependency chain.
* Keeps plugin discovery efficient by avoiding traversal of non-plugin packages.
* Naturally handles nested plugin dependencies while avoiding duplicate work and circular dependency issues.

lntutor:
Closes #953.

Plugin discovery now traverses dependencies only after confirming the parent package is a Clusterio plugin. Package paths are tracked to prevent duplicate work and dependency cycles; ordinary packages remain leaves of the traversal.

Tests cover a transitive plugin, a dependency cycle, a non-plugin package with a hidden plugin dependency, and a plugin that restricts package exports.

Validation:
- `pnpm run build`
- `pnpm exec mocha --enable-source-maps --check-leaks -R spec test/lib/plugin_loader.js` (12 passing)
- `pnpm exec eslint packages/lib/src/load_plugin_list.ts test/lib/plugin_loader.js`

### Changelog
```
### Changes
- Plugin discovery now traverses dependencies only after confirming the parent package is a Clusterio plugin. #963
```

### DIFF
--- a/packages/lib/src/load_plugin_list.ts
+++ b/packages/lib/src/load_plugin_list.ts
@@ -74,42 +74,16 @@ function getPluginName(requireSpec: string) {
 	return context.pluginInfo.name;
 }
 
-async function resolveNpmPackage(
-	packageName: string,
-	parentPath: string,
-): Promise<{ packagePath: string; requirePath: string } | undefined> {
-	const packageRequire = createRequire(path.join(parentPath, "package.json"));
-	let entryPath;
+function resolvePackagePath(packageName: string, fromPath: string): string | undefined {
+	const packageRequire = createRequire(path.join(fromPath, "package.json"));
 	try {
-		entryPath = packageRequire.resolve(packageName);
+		return path.dirname(packageRequire.resolve(`${packageName}/package.json`));
 	} catch (err: any) {
-		if (err.code === "MODULE_NOT_FOUND") {
+		if (err.code === "MODULE_NOT_FOUND" || err.code === "ERR_PACKAGE_PATH_NOT_EXPORTED") {
 			return undefined;
 		}
 		throw err;
 	}
-
-	const resolvedEntryPath = await fs.realpath(entryPath);
-	for (const searchPath of packageRequire.resolve.paths(packageName) ?? []) {
-		const packagePath = path.join(searchPath, packageName);
-		try {
-			await fs.access(path.join(packagePath, "package.json"));
-			const resolvedPackagePath = await fs.realpath(packagePath);
-			const relativeEntryPath = path.relative(resolvedPackagePath, resolvedEntryPath);
-			if (
-				relativeEntryPath !== ".."
-				&& !relativeEntryPath.startsWith(`..${path.sep}`)
-				&& !path.isAbsolute(relativeEntryPath)
-			) {
-				return { packagePath, requirePath: entryPath };
-			}
-		} catch (err: any) {
-			if (err.code !== "ENOENT") {
-				throw err;
-			}
-		}
-	}
-	return undefined;
 }
 
 /**
@@ -143,6 +117,13 @@ async function findNpmPlugins(pluginList: Map<string, string>): Promise<boolean>
 		parentPath: process.cwd(),
 		isRootDependency: true,
 	}));
+	const corePackages = new Set([
+		"@clusterio/controller",
+		"@clusterio/ctl",
+		"@clusterio/host",
+		"@clusterio/lib",
+		"@clusterio/web_ui",
+	]);
 	const visitedPackages = new Set<string>();
 
 	while (pendingPackages.length) {
@@ -150,15 +131,16 @@ async function findNpmPlugins(pluginList: Map<string, string>): Promise<boolean>
 		if (visitedPackages.has(packageName)) {
 			continue;
 		}
-		const npmPackage = await resolveNpmPackage(packageName, parentPath);
-		if (!npmPackage) {
+
+		const npmPackagePath = resolvePackagePath(packageName, parentPath);
+		if (!npmPackagePath) {
 			continue;
 		}
-		const { packagePath, requirePath } = npmPackage;
+
 		visitedPackages.add(packageName);
 
-		if (await checkPackageJson(packagePath)) {
-			const pluginName = getPluginName(requirePath);
+		if (await checkPackageJson(npmPackagePath)) {
+			const pluginName = getPluginName(npmPackagePath);
 			if (pluginList.has(pluginName)) {
 				logger.warn(
 					`${pluginName} provided by ${packageName}@${packageVersion} is already in the plugin list, ` +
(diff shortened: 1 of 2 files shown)


test/lib/plugin_loader.js
```diff
@@ -165,10 +177,12 @@ describe("lib/plugin_loader", function() {
...
 			await fs.mkdir(notAPluginPath, { recursive: true });
-			await fs.writeFile(
-				path.join(notAPluginPath, "package.json"),
-				JSON.stringify({ name: "not-a-plugin", version: "1.0.0" })
-			);
+			await fs.writeFile(path.join(notAPluginPath, "package.json"), JSON.stringify({
+				name: "not-a-plugin",
+				version: "1.0.0",
+				dependencies: { "hidden-plugin": "^1.0.0" },
+			}));
```
Cooldude2606: This needs to be `await writePlugin(notAPluginPath, "not-a-plugin", { "hidden-plugin": "^0.0.1" }, { keywords: [] });` because it is important the the index file exists for module resolution.

test/lib/plugin_loader.js
```diff
@@ -124,12 +124,16 @@ describe("lib/plugin_loader", function() {
 		const npmPluginPath = path.join(baseDir, "node_modules", "npm-plugin");
 		const monorepoPluginPath = path.join(baseDir, "external_plugins", "monorepo", "monorepo-plugin");
 		const monorepoPluginPathAbs = path.resolve(monorepoPluginPath);
+		const transitivePluginPath = path.join(baseDir, "node_modules", "transitive-plugin");
+		const hiddenPluginPath = path.join(
+			baseDir, "node_modules", "not-a-plugin", "node_modules", "hidden-plugin"
+		);
```
Cooldude2606: The hidden plugin should be installed in the same `node_modules` folder as the others, not a nested folder. It is hidden because we do not enumerate the node modules folder but instead rely on the root package json.

packages/lib/src/load_plugin_list.ts
```diff
@@ -98,13 +119,27 @@ async function findNpmPlugins(pluginList: Map<string, string>): Promise<boolean>
...
+		packageName,
+		packageVersion,
+		parentPath: process.cwd(),
+		isRootDependency: true,
+	}));
+	const visitedPackages = new Set<string>();
+
+	while (pendingPackages.length) {
+		const { packageName, packageVersion, parentPath, isRootDependency } = pendingPackages.shift()!;
+		const packagePath = await findNpmPackagePath(packageName, parentPath);
```
Cooldude2606: Is it possible to use `require.resolve` here rather than using our own method?

test/lib/plugin_loader.js
```diff
@@ -150,7 +156,13 @@ describe("lib/plugin_loader", function() {
 			// Create external plugin
 			await writePlugin(externalPluginPath, "external_plugin");
 			// Create npm plugin
-			await writePlugin(npmPluginPath, "npm");
+			await writePlugin(
+				npmPluginPath,
+				"npm",
+				{ "transitive-plugin": "^1.0.0" },
+				{ exports: "./index.js" },
```
Cooldude2606: We should probably have this as part of the defaults above as it is more correct.

packages/lib/src/load_plugin_list.ts
```diff
@@ -73,6 +73,27 @@ function getPluginName(requireSpec: string) {
 	return context.pluginInfo.name;
 }
 
+async function findNpmPackagePath(packageName: string, parentPath: string): Promise<string | undefined> {
```
Cooldude2606: If `require.resolve` does not work, please add a doc comment here explaining what the function does.

packages/lib/src/load_plugin_list.ts
```diff
@@ -98,13 +119,27 @@ async function findNpmPlugins(pluginList: Map<string, string>): Promise<boolean>
...
+	}));
+	const visitedPackages = new Set<string>();
+
+	while (pendingPackages.length) {
+		const { packageName, packageVersion, parentPath, isRootDependency } = pendingPackages.shift()!;
+		const packagePath = await findNpmPackagePath(packageName, parentPath);
+		if (!packagePath) {
+			continue;
+		}
+		if (visitedPackages.has(packagePath)) {
```
Cooldude2606: Using the path can result in duplicates if linking is used (such as with pnpm). So we should be using the package name as the unique identifier.

lntutor: Changed traversal deduplication to use the package name and added a nested duplicate-package regression in dfb60377.

test/lib/plugin_loader.js
lntutor: Hoisted hidden-plugin into the shared root node_modules directory; the regression still verifies that dependencies of non-plugin packages are not traversed, in dfb60377.

test/lib/plugin_loader.js
lntutor: Moved exports: ./index.js into the writePlugin defaults in dfb60377.

test/lib/plugin_loader.js
lntutor: Updated the non-plugin fixture to use writePlugin with keywords: [], so it has a resolvable entrypoint while remaining a non-plugin, in dfb60377.

packages/lib/src/load_plugin_list.ts
lntutor: The helper now uses native require.resolve semantics via createRequire; added coverage for aliases and nested dist/package.json boundaries in dfb60377.

packages/lib/src/load_plugin_list.ts
lntutor: Switched package discovery to native createRequire resolution and retained the owning package root for metadata checks in dfb60377.

lntutor:
All requested review changes are in dfb60377: native createRequire resolution, package-name deduplication, corrected shared fixtures/entrypoints, and the added alias/nested-boundary regressions. The fresh CI matrix is now fully green (8/8 checks). Ready for re-review.

Cooldude2606 (approved):
Thanks for the help on this PR. I have added a commit which drastically simplfies the path resolution function and continues to pass all tests you have introduced. The only material change to the tests is that the `package.json` is always in the root of plugin, this is an expectation of node modules.

PR #974, opened 2026-08-18
Filter dependency resolution by the mod pack's Factorio version

### LINKED ISSUE #965: Dependency download ignores the mod pack's Factorio version
### Summary

The download dependencies action ignores the mod pack's configured Factorio version when resolving dependency releases. As a result, it downloads the latest compatible dependency regardless of the mod pack's target Factorio version.

### Steps to reproduce

1. Create or open a mod pack targeting Factorio 2.0.
2. Add a mod that has different dependency releases for Factorio 2.0 and 2.1 (for example, a mod from the Py mod pack).
3. Select download dependencies.
4. Observe that the downloaded dependencies target Factorio 2.1 rather than Factorio 2.0.

### Expected behaviour

The dependency resolver should only consider releases matching the mod pack's configured Factorio version. In the example above, it should download the Factorio 2.0 versions of the dependencies.

### Actual behaviour

The dependency resolver ignores the mod pack's configured Factorio version and downloads the latest dependency releases, even when they target a different version of Factorio.

### Impact

This results in incompatible dependencies being added to the mod pack, requiring the user to manually replace them or resolve the resulting compatibility issues.

bbassie:
Fixes #965

The dependency resolver (`handleModDependencyResolveRequest`) received the mod pack's Factorio version but only used it for the built-in mod list. Both the mod portal release selection and the local mod store lookup filtered solely on the dependency's version range, so a 2.0 mod pack would resolve (and download) 2.1 dependency releases whenever they were newer.

Both lookups now also require the release's `factorio_version` to match the mod pack's. A dependency with no release for the pack's Factorio version is reported as `notFound` instead of pulling in an incompatible one.

Added unit tests for the resolver covering portal filtering, local mod filtering (with and without `checkForUpdates`), and the not-found case; all three fail without the fix.

Tested manually with a Factorio 2.0 mod pack containing Pyanodons Coal Processing — the resolved dependencies are now the 2.0 releases:

Resolved dependencies for a 2.0 mod pack

### Changelog
```
### Fixes
- Mod dependency resolution now only picks releases made for the mod pack's Factorio version. #965
```

### DIFF
--- a/packages/controller/src/ControlConnection.ts
+++ b/packages/controller/src/ControlConnection.ts
@@ -713,9 +713,14 @@ export default class ControlConnection extends BaseConnection {
 				return null;
 			}
 
-			// Select the latest matching version
+			// Select the latest matching version for the mod pack's Factorio version
 			const release = releases.releases
-				.filter(info => versionRange.testVersion(info.version))
+				.filter(info => (
+					versionRange.testVersion(info.version)
+					&& lib.normaliseMajorMinorVersion(
+						lib.normaliseSourceVersion(info.info_json.factorio_version)
+					) === factorioVersion
+				))
 				.reduce<ModPortalReleaseType | undefined>((max, cur) => (
 					max && lib.integerSourceVersion(max.version) > lib.integerSourceVersion(cur.version) ? max : cur
 				), undefined);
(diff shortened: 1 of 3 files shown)


Cooldude2606:
The fix looks good, however the tests should be placed along side the others in `tests/messages/mods.test.js`

bbassie:
Done, moved the tests to `test/messages/mod.test.js` (sorry, missed that harness). Two side effects of the fix in that file are worth calling out:
- The existing "prefers the locally installed version" fixtures had no `factorio_version` (so defaulted to 0.12) and now need one matching the pack; updated them.
- The live pymodpack 3.0.0 / 2.0 test no longer resolves `enable-all-feature-flags`. That was only being pulled in because the 2.1 releases of `pyalternativeenergy` were selected, which *recommend* it; the 2.0 releases only list it as optional. Dropped it from the expected set.

Cooldude2606 (approved):

PR #977, opened 2026-08-18
Support enemy_expansion.build_base_unit_dispatch_cooldown in exchange strings

bbassie:
Factorio 2.1.13 added `build_base_unit_dispatch_cooldown` to `enemy_expansion` in the map exchange string, and rejects map settings JSON that lacks the key. This made `readMapExchangeString` fail with "data after end" on 2.1.13+ strings and broke `instance load-scenario --map-exchange-string` on current Factorio (the `load-scenario` integration test failed on 2.1.14 with `Key "build_base_unit_dispatch_cooldown" not found in property tree at ROOT.enemy_expansion`).

To pin down position, type and version I generated real exchange strings from headless 2.1.7, 2.1.8, 2.1.12, 2.1.13 and 2.1.14 (created a map from the existing modified 2.1 settings and read `game.get_map_exchange_string()` over RCON) and diffed the inflated buffers using a distinctive value. The new field is an optional `uint32` that comes directly after `max_expansion_cooldown`, i.e. the last field of `enemy_expansion`; everything after it realigns exactly. It is present in 2.1.13 (which also rejects settings without the key) and absent in 2.1.12, so the reader gates it on `>= 2.1.13`.

The reader state now keeps the full version with an `atLeast(major, minor, patch)` helper, replacing the TODO comment about exposing the version once finer gating was needed; `gt_v2_0`/`gt_v2_1` are derived from it. Adds a `modified_v2_1_13` fixture string generated with 2.1.14 (same settings as `modified_v2_1` plus `build_base_unit_dispatch_cooldown: 1234`), a unit test asserting the full decoded output, and makes the `load-scenario` integration test use the new string on 2.1.13+.

Verified the decoded JSON matches the existing modified 2.1 settings apart from the new key on 2.1.13/2.1.14 while older strings decode unchanged, and the `load-scenario` and `instance kill` integration tests pass again on 2.1.14.

### Changelog
```
### Fixes
- Fixed reading map exchange strings from Factorio 2.1.13 and later, which broke `instance load-scenario --map-exchange-string` on current Factorio. #977
```

### DIFF
--- a/packages/lib/src/factorio/exchange_string.ts
+++ b/packages/lib/src/factorio/exchange_string.ts
@@ -5,16 +5,12 @@ class MapReaderState {
 	last_position = { x: 0, y: 0 };
 	/** Version of Factorio the string was created with */
 	version = [0, 0, 0, 0];
-	/** True when a version greater than 2.0.0 is detected */
-	gt_v2_0 = false;
-	/** True when a version greater than 2.1.0 is detected */
-	gt_v2_1 = false;
 	constructor(
 		public buf: Buffer
 	) { }
 
 	/** True when the version is at least the given major.minor.patch */
-	atLeast(major: number, minor: number, patch = 0) {
+	versionAtLeast(major: number, minor: number, patch = 0) {
 		const [vMajor, vMinor, vPatch] = this.version;
 		if (vMajor !== major) { return vMajor > major; }
 		if (vMinor !== minor) { return vMinor > minor; }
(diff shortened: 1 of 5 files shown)


Cooldude2606 (changes requested):
The code looks good, but as you have now added the `atLeast` helper could you refactor the other scetions to use this method rather than the the flags. Also for consitency can you name it `versionAtLeast`

bbassie:
Done: renamed to `versionAtLeast` and dropped the `gt_v2_0`/`gt_v2_1` flags in favour of it throughout the reader. Unit tests pass and older fixture strings still decode identically.

Cooldude2606 (approved):

PR #978, opened 2026-08-27
Add CI check for controller and host version compatibility

bbassie:
Adds a CI check that catches protocol breakage between the controller and host across releases, as discussed with Cooldude. Rather than the full four-way matrix, this tests the current code against the latest release published to npm in both directions, relying on releases being pairwise compatible: if each release is compatible with the previous one, checking against the latest transitively covers the rest.

## What it does

`scripts/version-compat-test.js` boots a minimal cluster with one side from the repo and the other installed from npm into `temp/compat-test/`. It configures and starts the controller, generates a host config through ctl, starts the host, and then verifies the parts that actually cross the version boundary: the host completes the handshake ("Started host" only logs after the connection is established), shows as connected in `host list`, and accepts an instance assignment. No Factorio install is needed. The ctl version always matches the controller it talks to.

Before booting the cluster the script also diffs the link messages registered by the local and published lib, since serialization changes are the most common way compat breaks (per Cooldude's review). It dumps `Link._requestsByName` and `Link._eventsByName` from each lib in a separate child process (the lib refuses to load twice in one process) and compares src, dst, data schema and response schema per message. A common message whose serialization changed fails the test with a diff of the schemas; messages that only exist on one side produce a warning annotation instead, since breaking compat is unavoidable when adding or removing a message.

The new `version-compat` job in `node.js.yml` runs the script both ways: local controller with published host, and published controller with local host.

## Notes

- `PUBLISHED_VERSION` is an env var on the job, defaulting to the `latest` dist-tag (currently 2.0.0-alpha.27). Pin it there to test a specific release instead. Testing the oldest still-supported release would avoid relying on compatibility being transitive.
- Since the repo is currently also at alpha.27, the check is near-tautological today; it starts pulling its weight as soon as either side drifts from the last release.
- Both directions pass locally against the published alpha.27. One subtlety: npm resolves the published host's lib dependency to the newest matching release, so the schema diff is labeled with the lib version actually installed, which can be newer than the pinned package version.

### Changelog
```
### Meta
- Added a CI check that the controller and host stay compatible with the latest published release. #978
```

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -47,6 +47,40 @@ jobs:
         name: logs-${{ matrix.node-version }}-${{ matrix.factorio-version }}
         path: temp/test/logs
 
+  version-compat:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    env:
+      # npm version spec for the published side of the compatibility test.
+      # If each release is compatible with the one before it then testing
+      # against the latest release transitively covers older versions too.
+      PUBLISHED_VERSION: latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - controller: local
+            host: published
+          - controller: published
+            host: local
+
+    steps:
+    - uses: actions/checkout@v4
+    - uses: pnpm/action-setup@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 24.x
+    - name: Prepare and install
+      run: pnpm i --no-frozen-lockfile
+    - name: Build
+      run: pnpm run build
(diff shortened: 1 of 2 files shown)


Cooldude2606 (changes requested):
I like this smoke test, but it misses one of our biggest failure modes. Most often compat will break because the searialision of link messages changed between versions.

Therefore as an addition can we include a check of all registered link messages that are common between the host and controller. Where a message was added or removed, compatibility breakage is unavoidable and should be logged but not fail the test annotated if possible.

Cooldude2606 (approved):
Im happy with this now, it will give us a clear indication when compatibility breaks between versions which is important going into 2.0.0.

PR #981, opened 2026-08-28
Make staticRoot include the static/ prefix and use it for plugin bundles

### LINKED ISSUE #980: Fix staticRoot and webRoot mixup and make static/ prefix part of staticRoot

bbassie:
Fixes #980

`staticRoot` was set to the same value as `webRoot`, so the `static/` prefix ended up scattered across consumers. `index.html` relied on the manifest path containing it, the export stores and `ModPackViewPage` hardcoded `${staticRoot}static/`, and `bootstrap.tsx` loaded plugin bundles from `webRoot` instead of `staticRoot`, which only worked because the manifest path happened to carry the prefix.

Now `staticRoot` is `${webRoot}static/`. The controller strips the `static/` prefix from bundle paths (main bundle and `/api/plugins` `web.main`) before handing them to the page, and the web UI drops the hardcoded `static/` segments. The helper lives in `routes.ts` rather than on `Controller` because `routes.ts` only used `Controller` as a type before and I did not want to add a value use across the existing circular import.

Tested by running a controller with player_auth added and checking `/`, `/hosts`, `/instances/1/view` (with and without trailing slash) and `/api/plugins`. The main bundle and plugin bundle both resolve under `/static/` with the new relative roots. There were no existing tests covering `serveWeb` or `/api/plugins`, so none were changed.

### Changelog
```
### Fixes
- Plugin web bundles are now loaded relative to `staticRoot` instead of `webRoot`. #980

### Breaking Changes
- `staticRoot` in the web interface now includes the `static/` prefix. Plugins that fetch assets with `${staticRoot}static/...` should drop the `static/` part. #980
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,16 @@ Many thanks to the following for contributing to this release:
 [@username](https://github.com/username)
 -->
 
+## Version 2.0.0-alpha.28
+
+### Fixes
+
+- Plugin web bundles are now loaded relative to `staticRoot` instead of `webRoot`. [#980](<https://github.com/clusterio/clusterio/issues/980>)
+
+### Breaking Changes
+
+- `staticRoot` in the web interface now includes the `static/` prefix. Plugins that fetch assets with `${staticRoot}static/...` should drop the `static/` part. [#980](<https://github.com/clusterio/clusterio/issues/980>)
+
 ## Version 2.0.0-alpha.27
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,16 +31,6 @@ Many thanks to the following for contributing to this release:
 [@username](https://github.com/username)
 -->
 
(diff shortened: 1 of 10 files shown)


Cooldude2606 (changes requested):
Code looks good but you should not modify the change log. Include the changelog entry in the pr description as per the template.

bbassie:
Reverted the CHANGELOG.md edit and moved the entries into the PR description per the template.

bbassie:
Added unit tests for `stripStaticPrefix` and `/api/plugins` to cover the new lines flagged by codecov/patch.

Cooldude2606 (approved):

Hornwitser (approved):
This works as I intended it to. There's one more feature I will make based on this, and that's a config option to set a custom static root so that you could have them hosted via a CDN while leaving the API outside of the cdn.

PR #983, opened 2026-08-31
Configurable static assets URL

Hornwitser:
Add option in the controller config to set where static assets are loaded from in order to support deployments where the static assets are served via a separate CDN while the API is directly accessed on the controller.

This is useful because some CDNs put low limits on how large requests bodies can be, which causes issues with uploading large mods and saves.

With the many locations where static assets can be located, I've also added a `copy-static` command to copy them out from those places into a folder, in case you want to host the assets from a separate web server like Nginx, instead of proxying and caching the requests to the the controller (which would be the simplest deployment).

### Changelog
```
### Features
- Added option to load static assets from a separate URL from where the controller's web API is hosted at. #983
- Added `clusteriocontroller copy-static` command to copy static web assets to a folder to ease hosting assets on a separate server. #983
```

### DIFF
--- a/packages/controller/controller.ts
+++ b/packages/controller/controller.ts
@@ -211,6 +211,40 @@ async function handleBootstrapCommand(
 	}
 }
 
+async function handleCopyStaticCommand(
+	args: any,
+	pluginInfos: lib.PluginNodeEnvInfo[],
+) {
+	await fs.cp(
+		path.join(__dirname, "..", "web", "static"),
+		args.target,
+		{
+			recursive: true,
+			mode: fs.constants.COPYFILE_FICLONE,
+		},
+	);
+	logger.info("Copied main web interface files");
+	for (const pluginInfo of pluginInfos) {
+		try {
+			await fs.cp(
+				pluginInfo.webStaticPath,
+				args.target,
+				{
+					recursive: true,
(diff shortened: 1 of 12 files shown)


packages/controller/controller.ts
```diff
@@ -211,6 +211,37 @@ async function handleBootstrapCommand(
...
+				{
+					recursive: true,
+					mode: fs.constants.COPYFILE_FICLONE,
+				},
+			);
+		} catch (err: any) {
+			if (err.code === "ENOENT") {
+				logger.info(`Skipped web interface files for ${pluginInfo.name}: ${err.message}`);
+			}
+		}
```
Cooldude2606, line 244: Should we re-throw errors which are not `ENOENT` so that they get output to the user correctly?

Cooldude2606 (approved):
Just the one comment, very minor tho. Also seeing you edit config.md makes me realise how many config options I have added but completely forgot to document; i'll need to revisit those at some point.

packages/controller/controller.ts
Hornwitser: Yes and no. It should output the error to the user, but it should not throw because you might put it in a deploy script and you don't want this to fail completely if one plugin is badly behaved.

PR #984, opened 2026-08-31
Migrate to ESM

Hornwitser:
The majority of our code in TypeScript is already using the ESM syntax, but TypeScript is configured to transpile it to CommonJS `require()` calls, which can create some interoperability problems and confusion. We also have lib which is imported as CommonJS modules in Node.js and ESM in the Web UI. Given that ESM is the future it makes sense to migrate the codebase to it, but this does pose some challenges with making the plugin loading still work.

The most notably issue is that you can't import a path like `./foo` and have it load `./foo.js`, or import a directorly like `./bar/` to have it import `./bar/index.js`. This means that all of our file relative imports have to be changed to include full filename with the `.js` extention (even if they import .ts files, because Node runs the transpiled .js files and TypeScript deliberately does not convert .ts to .js in imports).

### Changelog
```
### Changes
- Converted the project to using ES Modules. #984

### Breaking Changes
- Plugin entrypoints must now specify the full filename including the `.js` extension. #984
```


packages/lib/src/plugin_loader.ts
```diff
@@ -26,11 +26,15 @@ export async function loadPluginInfos(pluginList: Map<string, string>) {
...
 		let pluginInfo: libPlugin.PluginNodeEnvInfo;
 		let pluginPackage: { name?: string, version: string, main?: string, private?: boolean };
+		let packagePath;
 
-		// Check if plugin path exists, otherwise remove it
+		// Check if plugin has a package.json file, otherwise remove it
 		try {
-			require.resolve(pluginPath);
-		} catch {
+			const absolute = path.isAbsolute(pluginPath);
```
Cooldude2606, line 34: This is unused.

plugins/player_auth/controller.ts
```diff
@@ -6,9 +6,9 @@ import jwt from "jsonwebtoken";
 import { Static } from "@sinclair/typebox";
 import * as lib from "@clusterio/lib";
 import { BaseControllerPlugin } from "@clusterio/controller";
-const { basicType } = lib;
+import { basicType } from "@clusterio/lib";
```
Cooldude2606: We can also include `RequestError` here and then we dont need line 7.

Cooldude2606 (approved):
Two small things, happy for it to be merged once fixed.

packages/lib/src/plugin_loader.ts
Hornwitser: It's used in the migrate to ESM commit.

Cooldude2606: Oh I see, yeah it is. I was going commit by commit so missed that.

PR #989, opened 2026-09-05
Resolve numeric names in ctl id lookups

### LINKED ISSUE #886: Ctl interface breaks on numeric instances names
If you name an instance by a pure number like "10" you can create it with `clusterioctl instance create 10`, but you can't delete it with `clusterioctl instance delete 10` as the latter interprets 10 only as the id of the instance, not the name of it.

bbassie:
Fixes #886. `resolveHost`, `resolveInstance`, `resolveModPack` and `retrieveRole` treated any integer like argument as an id and never looked it up as a name, so an instance created as `10` could not be deleted with `clusterioctl instance delete 10`. The lookup now treats the argument as an id only if an entry with that id exists and otherwise falls back to a name match. When both an id and a name match, the id wins, which keeps the old behaviour for the common case.

The four functions now share one helper. `resolveModPack` also had a regex without an end anchor, so `10abc` used to resolve to id 10. Numeric ids that do not exist now fail in ctl with "No instance named 123" instead of a controller error, since the list is fetched either way.

### Changelog
```
### Fixes
- Fixed clusterioctl not being able to reference hosts, instances, mod packs and roles by an integer like name. #886
```

### DIFF
--- a/packages/lib/src/command.ts
+++ b/packages/lib/src/command.ts
@@ -147,6 +147,22 @@ export class CommandTree {
 
 
 /**
+ * Resolve a string with either an id or a name to the id of an entry
+ *
+ * An integer like string is treated as an id if an entry with that id
+ * exists, otherwise it is looked up as a name.
+ */
+function resolveId(entries: { id: number, name: string }[], idOrName: string) {
+	if (/^-?\d+$/.test(idOrName)) {
+		const id = parseInt(idOrName, 10);
+		if (entries.some(entry => entry.id === id)) {
+			return id;
+		}
+	}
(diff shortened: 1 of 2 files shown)


Cooldude2606 (approved):
Code and tests LGTM. @Hornwitser can I confirm that you intended for a "fallback" behaviour as the fix. I could see an alternative where we reject numeric names, but fallback feels like the correct solution to me.

Hornwitser:
Yes, falling back to using the numeric id if the name doesn't match by string is a good fix. It does have the edge case of being confusing if you have an instead with the name 10 and another instance with the id 10, but that seems like a too contrived case to care about.

PR #990, opened 2026-09-05
Sync shortcut toggles and game view settings in inventory_sync

### LINKED ISSUE #898: toggle personal roboports setting not preserved
toggle personal roboports setting not preserved when changing instances

bbassie:
Fixes #898. The personal roboport and equipment movement bonus toggles reset on every instance switch because their state was never part of the serialized player data. The built-in shortcuts do not set `toggleable` in their prototypes and `set_shortcut_toggled` only affects custom Lua shortcuts, so the fix syncs the state behind them instead: `allow_dispatching_robots` on the character (added to the character stat keys), `inhibit_movement_bonus` on the character's armor grid (restored after the armor inventory so the grid exists), and every field of `game_view_settings`, which covers alt mode (`show_entity_info`) and tall entity visibility (`hide_tall_entities`).

The game view settings key list is version gated: `show_surface_list` from 2.0.22, `hide_tall_entities` and `show_pins_gui` from 2.1.7. On restore only keys present in both the record and the target version are written, so records from another Factorio version are a no-op for the rest. `map_view_settings` is write-only in the API, so there is nothing to read on the source instance and it is left out.

Shortcuts with `toggleable` set are still synced through `is_shortcut_toggled` and `set_shortcut_toggled` for mods that define Lua shortcuts. Records without the new fields deserialize as before.

Lua lint: 0 findings over 28 files with the CI config. `tsc --noEmit` on the plugin passes.

### Changelog
```
### Fixes
- Fixed inventory_sync not preserving the personal roboport toggle, the equipment movement bonus toggle, game view settings such as alt mode, and toggleable mod shortcuts between instances. #898
```

### DIFF
--- a/plugins/inventory_sync/messages.ts
+++ b/plugins/inventory_sync/messages.ts
@@ -13,6 +13,7 @@ export type IpcPlayerData = {
 	cheat_mode: boolean,
 	flashlight: boolean,
 	shortcuts?: Record<string, boolean>,
+	game_view_settings?: Record<string, boolean>,
 	ticks_to_respawn?: number,
 	character?: any,
 	inventories?: any,
--- a/plugins/inventory_sync/messages.ts
+++ b/plugins/inventory_sync/messages.ts
@@ -12,6 +12,7 @@ export type IpcPlayerData = {
 	force: string,
 	cheat_mode: boolean,
 	flashlight: boolean,
+	shortcuts?: Record<string, boolean>,
 	ticks_to_respawn?: number,
 	character?: any,
 	inventories?: any,
--- a/plugins/inventory_sync/module/define_game_view_settings.lua
+++ b/plugins/inventory_sync/module/define_game_view_settings.lua
@@ -1,0 +1,30 @@
+local compat = require("modules/clusterio/compat")
+
+local keys = {
+	"show_alert_gui",
+	"show_controller_gui",
+	"show_crafting_queue",
+	"show_entity_info",
+	"show_entity_tooltip",
+	"show_hotkey_suggestions",
+	"show_map_view_options",
+	"show_minimap",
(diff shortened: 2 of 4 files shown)


Cooldude2606 (changes requested):
While these two shortcuts are the most important. I think we should sync all shortcuts that are `toggleable`

bbassie:
Done, it now iterates the shortcut prototypes and syncs every one with `toggleable` set. Names missing or not toggleable on the target are skipped on restore. Updated the PR title and description to match.

Cooldude2606:
It looks like the shortcuts do not work as expected as none of them are "toggleable" within their prototype. Instead we need to look elsewhere for our settings. For example `LuaEquipmentGrid::inhibit_movement_bonus` `LuaEntity::allow_dispatching_robots` `GameViewSettings.hide_tall_entities` and `GameViewSettings.show_entity_info` (although we likely want to sync all game view settings, as well as map view settings).

This doesnt mean we should remove the shortcuts syncing as that will still be useful for some mods, but we will need to add all of the above in order to close the linked issue.

bbassie:
Good catch, `set_shortcut_toggled` is documented as only affecting custom Lua shortcuts and none of the base game shortcuts set `toggleable`. Added syncing for `allow_dispatching_robots` (via the character stat keys), `inhibit_movement_bonus` on the character's armor grid (restored after the armor inventory so the grid exists), and every `game_view_settings` field, with version gates for the 2.0.22 and 2.1.7 additions. `map_view_settings` is write-only in the API, so there is nothing to read on the source instance and I left it out. The shortcut sync stays as is for mods with Lua shortcuts. Title and description updated.

Cooldude2606 (approved):
Tested locally and everything now syncs as expected.

PR #991, opened 2026-09-05
Add save copying and start options when creating an instance

### LINKED ISSUE #405: Feature Request: Add the ability to clone an existing Factorio instance to the same worker node
One of the most satisfying things about Clusterio is that when a given "problem" is solved, it can be packaged up as a Factorio save and scaled significantly further than any single map can reasonably handle. The mechanism for doing this right now is to:

* Download the save of the instance you'd like to duplicate (or save from in-game).
* Create a new instance.
* Assign the new instance.
* Manually duplicate any configuration changes like visibility, password, etc.
* Upload the save from the first step.

A simplification of this process could include "cloning" an existing instance to the same worker node, that worked similar to the following.

* A button to "clone" is added to the instance UI
* Clicking this button sends a command to make a save to the running instance
* An instance is created that is assigned to the same worker node, with the same configuration as the copied instance, with the save uploaded.

There's a couple of edge cases in configuration that I imagine shouldn't be exact clones. For example, the name should be modified by appending `(clone)` to the end or something. There's also some question in my mind about whether or not the save should happen, or just the last save should be taken - The instance size is effectively unbounded, so waiting for a save in the critical path is possibly annoyingly or prohibitively slow, unless the clone operation is some async op.

bbassie:
Closes #405. Creating an instance now offers a host, a server name, save copying and a start option.

The host dropdown shows for accounts with `core.instance.assign` and defaults to unassigned. Picking a host adds "Start after creation"; when no save is copied the instance creates a new one on first run. "Copy Config" is independent of the host, and when it points at an instance a "Server Name" field appears, prefilled with the source's `factorio.settings.name`, so the clone does not turn up in the server browser under the name of the source. With both a host and a config source set, "Copy Save" can copy the active save of the source (the loaded save if it is running, otherwise the save that would load by default) or first save the running game and then copy it. Source and target no longer have to share a host, the controller proxies the transfer. The modal shows a loading state while the steps run and navigates to the new instance once it has been created, even if a later step fails, so the error notification points at an instance that exists.

Saving a running game had no request before, so this adds `InstanceSaveGameRequest`. The host handler sends `/server-save` over RCON and resolves when the server reports the save finished, or rejects if the server exits first. It reuses the `core.instance.save.create` permission. The request is also exposed as `clusterioctl instance save-game <instance>` with a docs entry and an integration test next to the existing `/server-save` one.

Not included: a "save game" button on the saves list for running instances. The request makes that a small follow-up if wanted.

Verified with `pnpm test` (1671 passing, the port 8080 timeout is local to my machine and the one other failure passes when run on its own) and the full lint. The modal itself I drove in a browser against a controller and host: each field appears and disappears as described, creating with a changed server name gives an instance on the chosen host with the new `factorio.settings.name` and the rest of the settings intact, and copying a save with "Start after creation" ticked ends with the new instance running on the copied save.

### Changelog
```
### Features
- Added host, server name, save copy and start options to the create instance dialog in the web UI. #405
- Added `InstanceSaveGameRequest` and `clusterioctl instance save-game` to save the running game of an instance. #405
```

### DIFF
--- a/docs/managing-a-cluster.md
+++ b/docs/managing-a-cluster.md
@@ -65,6 +65,12 @@ Terminates the running Factorio server without giving it a chance to save or cle
 
 Note: This may cause loss of data.
 
+### Save game
+
+    ctl> instance save-game <name>
+
+Saves the running game of the instance to the save it was started from and waits for the save to finish.
+
 ### Extract players from instance
 
     ctl> instance extract-players <name>
--- a/packages/ctl/src/commands_instance.ts
+++ b/packages/ctl/src/commands_instance.ts
@@ -447,6 +447,16 @@ instanceCommands.add(new lib.Command({
 }));
 
 instanceCommands.add(new lib.Command({
+	definition: ["save-game <instance>", "Save the running game of an instance", (yargs) => {
+		yargs.positional("instance", { describe: "Instance to save the game on", type: "string" });
+	}],
+	handler: async function(args: { instance: string }, control: Control) {
+		let instanceId = await lib.resolveInstance(control, args.instance);
+		await control.sendTo({ instanceId }, new lib.InstanceSaveGameRequest());
+	},
+}));
+
+instanceCommands.add(new lib.Command({
 	definition: ["extract-players <instance>", "Extract players from running save into the cluster.", (yargs) => {
 		yargs.positional("instance", { describe: "Instance to extract players and online time from", type: "string" });
 	}],
--- a/packages/host/src/Instance.ts
+++ b/packages/host/src/Instance.ts
@@ -288,6 +288,7 @@ export default class Instance extends lib.Link {
 		this.handle(lib.InstanceLoadScenarioRequest, this.handleInstanceLoadScenarioRequest.bind(this));
 		this.handle(lib.InstanceSaveDetailsListRequest, this.handleInstanceSaveDetailsListRequest.bind(this));
 		this.handle(lib.InstanceCreateSaveRequest, this.handleInstanceCreateSaveRequest.bind(this));
+		this.handle(lib.InstanceSaveGameRequest, this.handleInstanceSaveGameRequest.bind(this));
 		this.handle(lib.InstanceExportDataRequest, this.handleInstanceExportDataRequest.bind(this));
 		this.handle(lib.InstanceRestartRequest, this.handleInstanceRestartRequest.bind(this));
 		this.handle(lib.InstanceStopRequest, this.handleInstanceStopRequest.bind(this));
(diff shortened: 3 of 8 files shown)


Cooldude2606:
This works very well, but I think we can extend this a little for evern greater benefit.
1. Firstly, if the user has `core.instance.assign` permission, then we should show a hosts drop down to select from. (Default: Unassigned)
2. When a host is selected, the "Start after creation" option is shown. A new save is created on first run if none are copied.
3. Regardless of if a host is selected, the "Copy config" option remains available.
4. When "Copy config" is selected, a "Server name" option is shown. This will overwrite `instance.server.name` on clone. Default is the value currently in the config.
5. If both a host and config are selected, the "Copy save" option becomes available.

A small nit, for "save and copy" can we change the text to not include the save name.

bbassie:
All five plus the nit are in 9b17fa2.

1. Host dropdown for accounts with `core.instance.assign`, default Unassigned.
2. Selecting a host shows "Start after creation". With no save copied it sends `InstanceStartRequest` without a save name, so the host creates one on first run.
3. Copy Config no longer depends on the host.
4. Picking an instance to copy adds a "Server Name" field, prefilled from the source's `factorio.settings.name` and applied as part of the create request. There is no `instance.server.name`, so it goes in as the whole `factorio.settings` object with only `name` replaced, which keeps description, tags and the rest of the cloned settings. Reading the default needs `core.instance.get_config`; without that permission the field stays hidden and the clone keeps the server name of the source. Say the word if you would rather have it always shown with an empty default.
5. Copy Save now needs a host and a config source. The two do not have to be on the same host, the controller proxies the transfer between hosts.
6. "Save and copy" now reads "Save the running game and copy it".

Still no browser on this machine, so the modal is typechecked rather than clicked through. The one part I could test headlessly is the server name: sending `factorio.settings` from control through `toRemote` and the controller's `update` keeps every other setting intact through the rename.

bbassie:
Correction to my last comment: I did have a browser available, I was wrong about that. I have now driven the modal in Chromium against a controller and host, and all of it holds up.

What I checked, with the source instance assigned to a host and `factorio.settings` carrying `name: "Source Server"` and `tags: ["keep-me"]`:

- With nothing selected the form is Name, Host (Unassigned), Copy Config.
- Picking a host adds "Start after creation".
- Picking a config adds "Server Name", prefilled with "Source Server", and "Copy Save", whose options are None, "Copy copy-me.zip" and "Save the running game and copy it".
- Setting the host back to Unassigned removes Copy Save and Start after creation again, and Server Name stays.
- Creating with the server name changed to "Renamed Server" produces an instance assigned to the chosen host with `factorio.settings.name` of "Renamed Server" and `tags` still `["keep-me"]`.
- Copying a save with "Start after creation" ticked ends with the new instance running on the chosen host with the copied save loaded.

The test spec was throwaway since Playwright is not a dependency of the repo, so nothing new is committed for it.

One thing the browser run turned up that predates this PR: the tooltip on "Copy Config" hangs over the field above it while the pointer is on the select. Not touching it here.

Cooldude2606 (approved):
Everything works as expected and allows for a quick way of creating new instances.

Issue #969, opened 2026-08-07
Add UDP communication support for instances


Cooldude2606:
### Summary

Add UDP communication to instances by exposing a `sendUDP` API and handling incoming UDP messages from Factorio through the existing IPC processing flows. UDP support should be opt-in through instance configuration and guarded by a plugin feature flag for plugins that require it.

### Motivation

Some plugins may benefit from low-latency, fire-and-forget communication where UDP is more appropriate than the existing RCON communication.

Currently, instances do not expose a way for plugins to send or receive UDP messages. Adding UDP support would allow plugins to use UDP-based protocols while keeping the functionality disabled by default for instances that do not require it.

### Proposed behaviour

Add a new plugin feature flag for UDP support, similar to existing plugin feature requirements such as `SavePatching` or `ScriptCommands`

Add UDP support to instances with the following capabilities:

* Provide a `sendUDP` API for plugins to send UDP packets to the Factorio server.
* Extend the existing IPC handling to detect incoming UDP messages from the Factorio server.

UDP support should be disabled by default and require it to be enabled in the instance config.

When enabled:

* The instance must be assigned a UDP port, following similar rule as assinging an RCON port.
* The instance startup command should include the required command-line option to enable UDP support on the Factorio server.
* Incoming and outgoing UDP communication should be available to plugins that require it.

### Benefits

* Enables plugins to use UDP communication where appropriate.
* Keeps additional networking functionality disabled unless explicitly required.
* Reuses the existing IPC processing model for incoming messages.

PR #996, opened 2026-09-09
Add UDP communication support for instances

bbassie:
Implements #969.

Adds opt-in UDP messaging between the host and the Factorio server, using the `--enable-lua-udp` option that Factorio 2.0.59 added. When `factorio.enable_lua_udp` is set the instance binds a localhost UDP socket, starts Factorio with `--enable-lua-udp` on `factorio.lua_udp_port` (random if null, like the RCON port) and passes the host port to the clusterio module through `clusterio_private.update_instance`.

Data out of Factorio goes through `clusterio_api.send_udp(channel, data)`, which uses the same IPC framing as `send_json`, so plugins receive it as the usual `ipc-<channel>` events. The host only accepts packets sent from the Lua UDP port, since Factorio always sends from that socket and this is otherwise the only unauthenticated way into the IPC handler. Data into Factorio goes through the new `sendUdp` method on instances and instance plugins and is dispatched in-game as `on_udp_packet_received` events, polled by the clusterio module every tick.

Plugins that need it declare the new `LuaUdp` feature flag, which refuses to load the plugin on instances without the option, the same way `SavePatching` and `ScriptCommands` work. In-game `send_udp` also needs script commands, because the host port only reaches the module over RCON. This is documented.

Verified against a headless 2.1.17 server: packets flow both ways, and `send_udp` / `recv_udp` are silent no-ops when the option is off, so a stale port in a save moved to an instance without UDP does no harm. The integration matrix now starts the test instance with UDP enabled on Factorio 2.0.59+ and round trips a player event over it. The 1.1.110 CI job skips it.

### Changelog
```
### Features
- Added opt-in UDP messaging between instances and the Factorio server via `factorio.enable_lua_udp`, with a `sendUdp` plugin API, `clusterio_api.send_udp` in-game and a `LuaUdp` plugin feature flag. [#969](<https://github.com/clusterio/clusterio/issues/969>)
```

### DIFF
--- a/docs/developing-for-clusterio.md
+++ b/docs/developing-for-clusterio.md
@@ -251,6 +251,20 @@ Parameters:
 - `channel`: string identifying which channel to send it on.
 - `data`: table that can be converted to JSON with game.table_to_json
 
+#### clusterio_api.send_udp(channel, data)
+
+Send JSON data to Clusterio over the given channel using UDP instead of stdout.
+Clusterio plugins receive it the same way as data sent with `send_json`.
+Requires `factorio.enable_lua_udp` to be enabled on the instance and Factorio 2.0.59 or later, otherwise the data is silently dropped.
+The port to send to is passed to the module with a script command on startup, so `factorio.enable_script_commands` must be enabled as well.
+See the [Communicating over UDP section](writing-plugins.md#communicating-over-udp) in the Writing Plugins document for more information.
+
+**Note**: Delivery is not guaranteed and payloads must be less than 64 kB.
+
+Parameters:
+- `channel`: string identifying which channel to send it on.
+- `data`: table that can be converted to JSON with game.table_to_json
+
 
 ### serialize library
 
--- a/docs/developing-for-clusterio.md
+++ b/docs/developing-for-clusterio.md
@@ -253,17 +253,16 @@ Parameters:
 
 #### clusterio_api.send_udp(channel, data)
 
-Send JSON data to Clusterio over the given channel using UDP instead of stdout.
-Clusterio plugins receive it the same way as data sent with `send_json`.
+Send data to Clusterio over the given channel using UDP instead of stdout.
+Clusterio plugins receive it as `udp-channel_name` events, see the [Communicating over UDP section](writing-plugins.md#communicating-over-udp) in the Writing Plugins document.
 Requires `factorio.enable_lua_udp` to be enabled on the instance and Factorio 2.0.59 or later, otherwise the data is silently dropped.
 The port to send to is passed to the module with a script command on startup, so `factorio.enable_script_commands` must be enabled as well.
-See the [Communicating over UDP section](writing-plugins.md#communicating-over-udp) in the Writing Plugins document for more information.
 
-**Note**: Delivery is not guaranteed and payloads must be less than 64 kB.
+**Note**: Delivery is not guaranteed and payloads should be kept well below 32 kB.
 
 Parameters:
 - `channel`: string identifying which channel to send it on.
-- `data`: table that can be converted to JSON with game.table_to_json
+- `data`: string with the payload to send, any bytes are allowed.
 
 
 ### serialize library
--- a/docs/developing-for-clusterio.md
+++ b/docs/developing-for-clusterio.md
@@ -255,7 +255,7 @@ Parameters:
 
 Send data to Clusterio over the given channel using UDP instead of stdout.
 Clusterio plugins receive it as `udp-channel_name` events, see the [Communicating over UDP section](writing-plugins.md#communicating-over-udp) in the Writing Plugins document.
-Requires `factorio.enable_lua_udp` to be enabled on the instance and Factorio 2.0.59 or later, otherwise the data is silently dropped.
+Requires `factorio.enable_lua_udp` to be enabled on the instance and Factorio 2.1.12 or later, otherwise the data is silently dropped.
 The port to send to is passed to the module with a script command on startup, so `factorio.enable_script_commands` must be enabled as well.
 
 **Note**: Delivery is not guaranteed and payloads should be kept well below 32 kB.
--- a/docs/developing-for-clusterio.md
+++ b/docs/developing-for-clusterio.md
@@ -255,7 +255,7 @@ Parameters:
 
 Send data to Clusterio over the given channel using UDP instead of stdout.
 Clusterio plugins receive it as `udp-channel_name` events, see the [Communicating over UDP section](writing-plugins.md#communicating-over-udp) in the Writing Plugins document.
-Requires `factorio.enable_lua_udp` to be enabled on the instance and Factorio 2.1.12 or later, otherwise the data is silently dropped.
+Requires `factorio.enable_lua_udp` to be enabled on the instance and Factorio 2.1.10 or later, otherwise the data is silently dropped.
 The port to send to is passed to the module with a script command on startup, so `factorio.enable_script_commands` must be enabled as well.
 
 **Note**: Delivery is not guaranteed and payloads should be kept well below 32 kB.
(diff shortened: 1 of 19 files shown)


Danielv123:
Quick question: Why do we need this? What advantages does it bring?

bbassie:
Fair question, the issue is light on motivation. The short version is that this is about latency and keeping non-critical high-frequency data off the two channels we have today. It is not a bandwidth feature.

Today everything out of Factorio goes through stdout. Every line passes through the output parser and the log handling, and on Windows payloads over 4 kB get reordered. Everything into Factorio goes through RCON as a script command. On a headless 2.1.17 I measured that Factorio runs one RCON command per tick no matter how small, each one waiting for a reply and a slot in the five command queue.

UDP gives a separate channel for data that is fine to lose. The example I have in mind is player positions. A plugin that shows where players are on a map, or feeds a live view, wants a position per player every few ticks. Over stdout that is thousands of lines an hour going through the log handler and output parser for data that is stale the moment the next one arrives. Over UDP it goes straight to the plugin and nothing else sees it.

Other data with the same shape: evolution and pollution samples, entity or train counts for graphs, research progress, per force production stats for a dashboard. Anything sampled on a timer where only the latest value matters and a dropped sample does not hurt.

Into the game it helps with frequent small messages. One recv_udp poll dispatches every queued packet, so a burst lands in the same tick instead of one per tick, and nothing queues behind a slow command or blocks a save or stop that also needs RCON. Cross server player positions or presence updates fit here.

What it does not do: payloads still enter game state as input actions, so they are replicated to connected clients like a script command. Delivery is not guaranteed, packets above roughly 32 kB are dropped, it needs Factorio 2.0.59 or later, and the module needs script commands to learn the host port. It is off by default and costs nothing when unused. Plugins that need it declare the LuaUdp feature so they refuse to load without it.

Cooldude filed the issue so he may have a specific plugin in mind as well.

Cooldude2606:
I have no plans at the moment for this feature. But this is a command line flag which is currently inaccessible and could be useful for plugins which wish to use it. Therefore just looking for a bare bones IPC to make it available for use. I am unlikely to merge this before 2.0

docs/developing-for-clusterio.md
```diff
@@ -251,6 +251,20 @@ Parameters:
 - `channel`: string identifying which channel to send it on.
 - `data`: table that can be converted to JSON with game.table_to_json
 
+#### clusterio_api.send_udp(channel, data)
+
+Send JSON data to Clusterio over the given channel using UDP instead of stdout.
```
Cooldude2606: We should not limit udp to json data, it is most useful for sending any arbitrary blob.

docs/developing-for-clusterio.md
```diff
@@ -251,6 +251,20 @@ Parameters:
...
+Clusterio plugins receive it the same way as data sent with `send_json`.
```
Cooldude2606: Plugins should have a seperate hook for udp as the data is likely to be different to regular ipc messages. But it should still have a channel.

packages/host/lua/clusterio_lib/api.lua
```diff
@@ -64,5 +67,16 @@ function api.send_json(channel, data)
...
 end
 
+function api.send_udp(channel, data)
+	local port = call_api("get_host_udp_port")
+	if not port then
+		return -- Lua UDP is not enabled for this instance
+	end
+
+	channel = escape_channel(channel)
+	data = compat.table_to_json(data)
```
Cooldude2606: As above, we should not limit this to json.

packages/host/src/server.ts
```diff
@@ -1061,6 +1128,7 @@ export class FactorioServer extends events.EventEmitter<FactorioServerEvents> {
 		this._server = null;
 		this._rconClient = null;
 		this._rconReady = false;
+		this._closeUdpSocket();
```
Cooldude2606: This should align with how we handle the rcon client.

Cooldude2606 (changes requested):

docs/developing-for-clusterio.md
bbassie: Changed in 7fcb463. send_udp now takes any string and the packet is just channel?payload, no JSON framing.

docs/developing-for-clusterio.md
bbassie: Changed in 7fcb463. Packets now come out as a udp-<channel> event with the payload as a Buffer, plus a handleUdp(channel, handler) helper that logs handler errors the same way handle does for ipc events. The channel escaping is shared with the ipc path.

packages/host/lua/clusterio_lib/api.lua
bbassie: Same change as in the docs thread, it takes a plain string now.

packages/host/src/server.ts
bbassie: Changed in 7fcb463. The socket is now opened from the same output heuristic that starts the RCON client and closed in the exit, error and stop timeout handlers next to rconClient.end(). _resetState only nulls the reference like it does for the RCON client, and a bind failure is logged like a failed RCON connect rather than thrown.

bbassie:
The 2.0.77 and 2.1.8 jobs failed on the last push because Factorio itself exited with code 1 in the new to-game UDP test. I reproduced it on headless 2.0.77 and 2.1.8 outside the test suite: whenever helpers.recv_udp dispatches a received packet the server dies with "TickClosure.cpp: Trying to add invalid input action to the closure". It does not matter if recv_udp runs from on_tick or from a command, or whether the server is paused or has auto_pause off. 2.1.12, 2.1.14 and 2.1.17 all work, and I could not find the fix in the changelog.

So the minimum for factorio.enable_lua_udp is now 2.1.12. FactorioServer refuses to start with it enabled on older versions, the docs say so, and the integration tests only turn it on from 2.1.12 up, which means the 2.0.77 and 2.1.8 CI jobs skip the round trip.

Danielv123:
It was actually fixed in 2.1.10.

The problem is that helpers.recv_udp used an inputaction for synchronization, which failed because the headless server doesn't have a playerindex. There is a whitelist for inputactions that are allowed to be sent without a playerindex and it wasn't in there.

Danielv123:
Regarding the minimized comment: There is no limit to the number of commands per tick. You can run a dozen of them if you wanted.

For output, we usually write to files when the output is initiated from the game. It works well.

bbassie:
Thanks, that explains it. I pulled the 2.1.9 and 2.1.10 headless builds and confirmed it: 2.1.9 still crashes, 2.1.10 works on both the on_tick and command paths. The floor is 2.1.10 now in the docs, the config description, the start check and the test gates, and the docs say headless server rather than paused since that is the actual trigger.

bbassie:
Fair correction. What I measured was sequential round trips, each command waiting for its reply before the next is sent, which comes out as one per tick from the client's point of view. Pipelined commands do land in the same tick, and Cooldude already pointed out that pushing more per tick is not a win anyway since every command is replicated to clients. And yes, the file path covers large output from the game well. The remaining case for UDP is small, frequent, droppable data that should not go through the output parser or the log at all, and it is opt-in for exactly that.

bbassie:
Note on the coverage drop: with the floor at 2.1.10 none of the Factorio versions in the CI matrix (1.1.110, 2.0.77, 2.1.8) enable Lua UDP any more, so the integration round trip through a real server is skipped in CI and only runs locally. The remaining uncovered lines in server.ts are the start command flag and the socket opening on a live server. The host side is covered by unit tests. Once the matrix moves to a 2.1.10 or later build the round trip runs again.

Cooldude2606 (approved):
There were one or two nits in the tests, but nothing which impacts behaviour so I'm happy as is. This will be merged after 2.0

Issue #964, opened 2026-08-04
Move recipe notification merging from Lua to the instance plugin


Cooldude2606:
### Summary

Move the merge logic for recipe notifications from Lua into the instance plugin. By comparing the imported recipe notifications against the player's current state before sending data back to Factorio, the amount of TypeScript → Lua IPC traffic can be significantly reduced while preserving the existing behaviour.

### Motivation

Inventory synchronization currently sends the complete set of recipe notifications back to Lua, where they are merged with the player's existing state.

On heavily modded servers, `recipe_notifications` can grow very large despite changing infrequently. As a result, importing an otherwise empty inventory may still require hundreds of IPC chunks simply to resend recipe notifications that already exist on the destination instance.

The IPC channel from Lua to TypeScript is effectively unconstrained for this operation, whereas communication from TypeScript back to Lua must be chunked. Performing the merge in the instance plugin therefore allows redundant data to be removed before the expensive direction of communication.

### Proposed behaviour

Update `ipc-inventory_sync_download` and `handleDownload` so that the download request includes the player's current recipe notification state from Lua.

The synchronization flow would become:

1. Lua initiates `ipc-inventory_sync_download`, including the player's current `recipe_notifications`.
2. The instance plugin requests the imported inventory from the controller.
3. The instance plugin compares the imported recipe notifications against the current state using the same merge behaviour that is currently implemented in Lua.
4. Any recipe notifications that already exist are discarded.
5. Only the remaining recipe notifications are sent back to Lua as part of the inventory import.

The resulting behaviour should be functionally identical to the current implementation, with only the location of the merge logic changing.

### Why move the merge to the instance plugin?

The primary motivation is to reduce the amount of data sent from TypeScript to Lua.

* Lua → TypeScript communication is effectively unconstrained for this operation.
* TypeScript → Lua communication must be chunked, making large payloads significantly more expensive.
* Recipe notifications are largely append-only, so most of the data being transmitted already exists on the destination instance.

Moving the merge to the instance plugin allows redundant data to be discarded before transmission, reducing both IPC traffic and import time while preserving the existing behaviour.

### Alternatives considered

* **Allow disabling recipe notification synchronization.** This would benefit users who do not care about recipe notifications, but it does not improve performance for those who do.
* **Compress or serialize the data more efficiently.** This would likely require introducing a shared lookup table or another custom serialization format, increasing implementation complexity while still transmitting largely unchanged data.
[...]

PR #997, opened 2026-09-09
Send only recipe notification changes on inventory download

bbassie:
Implements the proposal in #964. The download request from Lua now includes the player's current recipe notification state, encoded the same way as the stored data. The instance plugin decodes both, computes the names to add to and remove from the cleared list, and sends that delta in the `recipe_notifications` field in place of the full list. The Lua deserializer applies the delta directly: names to add get their notification cleared, names to remove get it back, and names that don't exist on this instance go to the failed list. The state Lua sends is only its own cleared recipes, so any stored name unknown here shows up in the add list on every download and the failed list is recomputed from scratch each time without depending on the previous round.

A few decisions worth a look:

- The delta is always sent when the stored data has notifications. Without a snapshot from Lua, which only happens on Factorio versions before the notification API, the delta is computed against an empty list and adds everything. That's the same size as today and Lua ignores the field on those versions anyway.
- If the stored data can't be decoded the plugin drops the field with a warning rather than failing the whole download. The next upload from the instance replaces the broken value.
- Storage on the controller and the upload path are unchanged, so there's no migration.

While reading the serializer I found that `serialize_crafting_notifications` appended cleared names into the stored failed list in place. A player who rejoins the same server repeatedly without a download in between would grow that list on every leave. The serializer now copies the list instead.

Measured on a headless 2.1.17 server: the encoded list is about 7.4 bytes per recipe, so a fully researched Space Age game is 3 chunks at the default chunk size and a ten thousand recipe modpack would be around 75. With this change the field is one chunk or less in the common case. Each chunk costs at least one tick, and since script commands are replicated to every connected client the chunk size is a per-tick bandwidth budget rather than something to raise, so bytes saved translate directly into ticks saved.

Unit tests cover the delta computation and the host-side encode path. The Lua rebuild was checked inside Factorio against a delta produced by Node.

### Changelog
```
### Changes
- Inventory sync sends only the recipe notifications that differ from the destination instance's state on download. #964

### Fixes
- Fixed failed recipe notification names being duplicated on every upload from the same instance in inventory sync. #964
```


plugins/inventory_sync/module/serialize.lua
```diff
@@ -533,17 +533,61 @@ function serialize.serialize_crafting_notifications(player, failed)
...
+--- Current notification state, sent with download requests so the instance plugin only returns the difference
+--- @param player LuaPlayer
+--- @param failed FailedDeserializationPlayerData?
+--- @return string?
+function serialize.crafting_notification_snapshot(player, failed)
+	if not recipe_notifications_api then
+		return nil
+	end
+	return serialize.serialize_crafting_notifications(player, failed and failed.recipe_notifications)
+end
```
Cooldude2606: What is the purpose of this function? It would be better to have the if statement visible at the call site rather than hidden here.

plugins/inventory_sync/module/inventory_sync.lua
```diff
@@ -111,6 +111,13 @@ function inventory_sync.deserialize_player(player, finished_record)
...
 	local serialized_player = compat.json_to_table(finished_record.data)
 	assert(type(serialized_player) == "table", "wrong type for serialized_player")
+	if serialized_player.recipe_notifications_delta then
+		serialized_player.recipe_notifications = serialize.apply_crafting_notification_delta(
+			assert(finished_record.recipe_notifications, "recipe notification delta without snapshot"),
+			serialized_player.recipe_notifications_delta
+		)
+		serialized_player.recipe_notifications_delta = nil
+	end
 	script_data.failed_deserialization[player.name] = serialize.deserialize_player(player, serialized_player)
```
Cooldude2606: Why do not need to support full deseialise and delta. So we should reuse `recipe_notifications` and change deserialise to use the delta format.

Cooldude2606 (changes requested):

plugins/inventory_sync/module/serialize.lua
bbassie: Removed in bb36af08, the version check is now at the call site in initiate_inventory_download.

plugins/inventory_sync/module/inventory_sync.lua
bbassie: Done in bb36af08. The delta goes in recipe_notifications and deserialize_crafting_notifications applies it directly, taking the previous failed names as a parameter.

Cooldude2606 (changes requested):
Had to fix an quick error with the `failed` array but now the good works as expected, so I'm happy to merge. Although, needing to provide the previous failed deserialise to the current one feels very messy, especially because it is persistent from a previous round. Could we please have it so the "current state" given on download init does not depent on previous state.

bbassie:
Thanks for the fix. b1e79cf8 removes the dependency on the previous round: Lua now sends only its own cleared recipes on download init, so every stored name unknown on the instance lands in the add list each time and the deserializer recomputes the failed list from that. The previous_failed parameter is gone, and with it the case your fix covered.

Cooldude2606 (approved):

PR #999, opened 2026-09-09
Rename RequestError to ExpectedError

### LINKED ISSUE #900: Rename `RequestError` to match its more general purpose
When `RequestError` was first introduced, its purpose was to signal to requests that an error should be passed to the requester rather than logged as an unexpected error. This gradually expanded as more errors would set to inherit from the class. Therefore, the name "Request" is no longer meaningful and should be updated to represent any error that is part of normal program control flow. Example case where this is used include `PermissionError`, `InvalidAccess`, `InvalidValue`, and `InvalidField` (see #895)

bbassie:
`RequestError` began as the marker for errors that should be passed back to the requester instead of logged, but it has become the base class for any error that is part of normal control flow. `PermissionError`, `InvalidAccess`, `InvalidValue` and `InvalidField` all extend it. This renames it to `ExpectedError`, which reads naturally next to the "Unexpected error responding to ..." log line that the check in `Link` guards.

I picked `ExpectedError` over alternatives like `OperationalError` or `HandledError` because it describes the contract directly: this error is expected, do not log it as a bug. Happy to change the name if there is a preference.

`RequestError` is kept as a deprecated alias (both value and type) so plugins keep compiling and running. The default `code` sent in error responses changes from `RequestError` to `ExpectedError`; nothing in the repo compares against that string.

The alpha 14 migration doc still shows `lib.RequestError` in its examples. I left it alone since it documents that specific migration, but can update it if wanted.

### Changelog
```
### Changes
- Renamed `RequestError` to `ExpectedError` to reflect its use as the base class for all errors that are part of normal control flow. `RequestError` remains as a deprecated alias. #900
```

Closes #900

### DIFF
--- a/packages/controller/src/BaseConnection.ts
+++ b/packages/controller/src/BaseConnection.ts
@@ -63,7 +63,7 @@ export default class BaseConnection extends lib.Link {
 		let { id } = request;
 		let modPack = this._controller.modPacks.get(id);
 		if (!modPack) {
-			throw new lib.RequestError(`Mod pack with ID ${id} does not exist`);
+			throw new lib.ExpectedError(`Mod pack with ID ${id} does not exist`);
 		}
 		return modPack;
 	}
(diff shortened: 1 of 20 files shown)


Cooldude2606:
Marked this as draft as discussion will be required around what name to use.

PR #1014, opened 2026-09-10
Add create-scenario command to clusteriohost

bbassie:
Adds `npx clusteriohost create-scenario <output>`, which writes a scenario directory with the Clusterio modules patched into it. The idea came up on Discord: a way to get a "patched scenario" so the modules can be reused outside of Clusterio without running a cluster. Putting it on clusteriohost lets it reuse the save patching as is.

The command zips the base scenario into a temp file, runs the existing `patch()` on it and extracts the result, so there is no second patching code path to keep in sync. The base scenario defaults to freeplay from the Factorio install in `host.factorio_directory` (`--factorio-version` picks the install, default latest). `--scenario <path>` accepts a scenario directory, a zipped scenario or a save instead, with the same restriction as save patching: it has to be a known event_handler scenario or something Clusterio has patched before. Modules come from every plugin enabled in the host config, or from the plugins named with `--plugins`. The command refuses to write over an existing output directory.

The module gathering in `Instance.prepareSave` moved to `patch.loadModules` so the instance and the command share it, and `findVersion` in server.ts is now a regular export instead of a test-only one.

`--zip` keeps the output as a zip file for moving it to another machine. Factorio does not pick up zips from the scenarios folder (`--scenario2map` reports the scenario as not found), so the default is to extract, and the README says to extract before use. The directory form does load: `--scenario2map` on the output with headless 2.1.17 ran control.lua and produced a save.

Tests cover directory and zip input, that the input is left untouched, the existing-output and unknown-scenario errors, and `loadModules` with the bundled module and a plugin module. `test/host` and `test/integration` pass apart from the controller bypass-users timeout that is local to my machine. Docs added to the host README and a pointer in how-it-works.md.

### Changelog
```
### Features
- Added `clusteriohost create-scenario` to create a scenario with the Clusterio modules patched in for use outside of Clusterio. #1014
```


packages/host/src/create_scenario.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Runs the save patching on the scenario at scenarioPath and writes the
+ * result to the directory at outputPath.  The scenario can be a directory,
+ * a zipped scenario or a save.
+ *
+ * @param scenarioPath - Path to the scenario to patch.
+ * @param outputPath - Directory to write the patched scenario to.
+ * @param modules - Description of the modules to patch.
+ */
+export async function createScenario(scenarioPath: string, outputPath: string, modules: SaveModule[]) {
+	if (await fs.access(outputPath).then(() => true, () => false)) {
```
Cooldude2606: This does not account for the case of not having permission to access an existing dir.

packages/host/src/create_scenario.ts
```diff
@@ -0,0 +1,149 @@
...
+		throw new Error(`${outputPath} already exists`);
+	}
+
+	const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), "clusterio-scenario-"));
```
Cooldude2606, line 41: Why is there a `-` on the end of this name.

packages/host/src/create_scenario.ts
```diff
@@ -0,0 +1,149 @@
...
+	try {
+		const tempPath = path.join(tempDir, "scenario.zip");
+		if ((await fs.stat(scenarioPath)).isDirectory()) {
+			await zipDirectory(scenarioPath, tempPath);
+		} else {
+			await fs.copyFile(scenarioPath, tempPath);
+		}
+
+		await patch(tempPath, modules);
+		await extractZip(tempPath, outputPath);
```
Cooldude2606: I think we should have an option to leave the file zip'd as the user might want to transfer it to another device. Extracting the files should remain the default.

packages/host/src/create_scenario.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+export function createScenarioCommand(yargs: any) {
+	yargs
+		.positional("output", {
+			describe: "Directory to write the scenario to",
+			type: "string",
+		})
+		.option("scenario", {
+			nargs: 1,
+			describe: "Scenario directory, zipped scenario or save to patch, defaults to freeplay",
```
Cooldude2606: It is not clear that the default "freeplay" is a special case, and you cannot use scenario names as the input.

docs/how-it-works.md
```diff
@@ -5,6 +5,7 @@ The save patching is based on the built in [event_handler](https://github.com/wu
 Mods on the other hand work without modifications.
 
 This can be disabled with the `factorio.enable_save_patching` config on a per instance basis, however disabling it means that modules are not loaded into the game and most plugins will not function.
+To use the modules outside of Clusterio the `npx clusteriohost create-scenario` command creates a scenario with the modules patched into it.
```
Cooldude2606, line 8: We should make it clear that IPC is not supported when run outside of a clusterio context, so many modules may not work as expected as a standalone scenario.

Cooldude2606 (changes requested):
Some changes required, plus fixing the conflict resulting from the ESM pr.

packages/host/src/create_scenario.ts
bbassie: Changed to fs.stat and only ENOENT counts as not existing, anything else (EACCES included) is rethrown.

packages/host/src/create_scenario.ts
bbassie: mkdtemp appends six random characters to the prefix, so the dash separates them: clusterio-scenario-Ab3xYz. Same convention as the node docs example (`foo-`).

packages/host/src/create_scenario.ts
bbassie: Added `--zip`. Extracting stays the default. The README notes that Factorio does not load zipped scenarios from the scenarios folder (checked with --scenario2map), so it is for transfer only.

packages/host/src/create_scenario.ts
bbassie: Reworded to "Path to a scenario directory, zipped scenario or save to patch. Defaults to the freeplay scenario of the Factorio install in host.factorio_directory".

docs/how-it-works.md
bbassie: Added a sentence to both the docs and the README that IPC is not available outside Clusterio, so modules that depend on it will not work as expected.

Cooldude2606 (approved):

Cooldude2606:
This will be merged after 2.0

PR #1018, opened 2026-09-10
Add plugin install command to controller, host and ctl

### LINKED ISSUE #1008: Rework plugin installation
Currently a plugin is installed in two steps 
1. Run `npm install npm-plugin-name`
2. Run `npx clusteriocontroller/host plugin add npm-plugin-name`

The issue with this is that both of these commands works in more or less any directory, which leads to users installing the plugin in the wrong directory and then not understanding why it doesn't work. Better documentation could partially solve this.

bbassie:
Installing a plugin used to take `npm install` followed by `plugin add`, and both happily run in any directory. Run them somewhere else and the plugin lands in a stray node_modules while plugin-list.json gets written next to it, and nothing tells the user why the plugin never loads.

This adds `plugin install <package>` to clusteriocontroller, clusteriohost and clusterioctl. It runs `npm install --save <package>` and then reuses `loadPluginList` to discover what the package provides, so one command does both steps. The package spec is passed straight to npm, so versions, `file:` paths and tarballs all work.

Before doing anything it checks that the current directory is a Clusterio installation, meaning package.json is the one the installer writes (`clusterio-install`) or depends on @clusterio/controller, host or ctl. Outside of one it exits with an error naming the directory. `plugin add` with a bare package name gets the same check but only warns, since it does not create anything on its own. Relative and absolute paths to `plugin add` are unaffected, the integration tests still use those.

Reinstalling a plugin that is already listed prints "already installed" and exits 0. Installing a package that has no clusterio-plugin keyword fails after npm finishes, the existing "is not a clusterio-plugin" warning from `loadPluginList` says which package. The plugin list is only saved by `loadPluginList`, so a failed npm run leaves it untouched.

Tests run real `npm install` against test/file/test_plugin in a temp directory. It is a local path so there is no network involved and each run takes about 150 ms. The READMEs now show `plugin install` as the way to install published plugins and keep `plugin add` for plugins under development.

Closes #1008

### Changelog
```
### Features
- Added `plugin install <package>` command to controller, host and ctl which installs a plugin from npm and adds it to the plugin list. #1008

### Changes
- `plugin install` and `plugin add` with a package name now check that they are run from a Clusterio installation directory. #1008
```


packages/controller/README.md
```diff
@@ -25,19 +25,30 @@ Configure plugins available to be loaded by the controller.
...
 
+#### `plugin add <path>`
+
+Add plugin either by require path or relative/absolute path to the plugin directory.
+A relative path must start with ./ or ../ (or .\ and ..\ on Windows) otherwise it will be assumed to be a require path for an installed package in node_modules.
+This is mainly useful for plugins under development, packages published on npm are better installed with `plugin install`.
+
+For example, adding a plugin checked out next to the Clusterio installation:
+
+    npx clusteriocontroller plugin add ../my_plugin
```
Cooldude2606: This example should be `./external_plugins/my_plugin`

packages/ctl/README.md
```diff
@@ -27,19 +27,30 @@ Configure plugins available to be loaded by ctl.
...
 
+#### `plugin add <path>`
+
+Add plugin either by require path or relative/absolute path to the plugin directory.
+A relative path must start with ./ or ../ (or .\ and ..\ on Windows) otherwise it will be assumed to be a require path for an installed package in node_modules.
+This is mainly useful for plugins under development, packages published on npm are better installed with `plugin install`.
+
+For example, adding a plugin checked out next to the Clusterio installation:
+
+    npx clusterioctl plugin add ../my_plugin
```
Cooldude2606: As above.

packages/host/README.md
```diff
@@ -26,19 +26,30 @@ Configure plugins available to be loaded by the host.
...
 
+#### `plugin add <path>`
+
+Add plugin either by require path or relative/absolute path to the plugin directory.
+A relative path must start with ./ or ../ (or .\ and ..\ on Windows) otherwise it will be assumed to be a require path for an installed package in node_modules.
+This is mainly useful for plugins under development, packages published on npm are better installed with `plugin install`.
+
+For example, adding a plugin checked out next to the Clusterio installation:
+
+    npx clusteriohost plugin add ../my_plugin
```
Cooldude2606: As above.

packages/lib/src/shared_commands.ts
```diff
@@ -35,6 +39,65 @@ export function pluginCommand(yargs: any) {
...
+}
+
+async function packageNameFromSpec(packageSpec: string) {
+	try {
+		const packageJson = JSON.parse(await fs.readFile(path.join(packageSpec, "package.json"), "utf8"));
+		if (typeof packageJson.name === "string") {
+			return packageJson.name;
+		}
+	} catch {
+		// Not a local directory, treat it as name@version
```
Cooldude2606: Shouldn't we only say "not a local dir" if it does not exist. There are cases were we might get a permission error in which case we need to throw the error.

test/lib/shared_commands.js
```diff
@@ -0,0 +1,129 @@
...
+
+		it("should refuse to install outside a Clusterio installation", async function() {
+			await run("install", testPluginPath);
+			assert.equal(process.exitCode, 1);
+			assert.equal(pluginList.size, 0);
+			await assert.rejects(fs.access("node_modules"));
+		});
+
+		it("should install and add an npm plugin", async function() {
+			this.timeout(60000);
```
Cooldude2606: Can we make sure to call `slowTest` for these so they are skipped when the fast flag is present.

Cooldude2606 (changes requested):

packages/controller/README.md
bbassie: Changed to `./external_plugins/my_plugin` in 7c0f774.

packages/ctl/README.md
bbassie: Changed to `./external_plugins/my_plugin` in 7c0f774.

packages/host/README.md
bbassie: Changed to `./external_plugins/my_plugin` in 7c0f774.

packages/lib/src/shared_commands.ts
bbassie: Good catch. Now only ENOENT and ENOTDIR fall through to the name@version parse, anything else is rethrown. 7c0f774

test/lib/shared_commands.js
bbassie: Done, the four tests that run npm now call `slowTest(this)` and are skipped with FAST_TEST. 7c0f774

Cooldude2606 (approved):

bbassie:
Converted test/lib/shared_commands.js to ES module syntax on top of the master merge, that was what broke the test jobs.

PR #1019, opened 2026-09-10
Require a token to scrape /metrics

### LINKED ISSUE #1009: Protection for metrics endpoint
The /metrics endpoint needs to be protected with authentication/rate limiting.

bbassie:
Fixes #1009.

The `/metrics` endpoint was open to anyone who could reach the controller. That exposed cluster statistics, and since every scrape fans out a gather to all connected hosts, it also let anyone put load on the whole cluster at will.

Scraping now requires a user token for a user with the new `core.controller.metrics` permission. Missing or invalid tokens get 401, users without the permission get 403. The token is accepted as `Authorization: Bearer <token>` in addition to `X-Access-Token`, because Prometheus can only send the former. The Bearer form is accepted by the other token-protected endpoints too, since they share the same middleware.

For rate limiting I went with coalescing rather than rejecting: concurrent scrapes share one gather, so the hosts see at most one metrics request at a time no matter how many scrapers poll. A well-behaved Prometheus never notices, and a misconfigured one can no longer multiply host load.

The three connection gauges in the controller entrypoint now no-op when the controller has not been created, same as the other gauges in that file. That is what makes the endpoint testable against the mock controller.

Docs: added a "Prometheus Metrics" section to managing-a-cluster.md with the role/user/token recipe and the matching `prometheus.yml` snippet.

Full suite: 1683 passing, 1 failing (the known port-8080 conflict on my machine, unrelated).

### Changelog
```
### Changes
- Access tokens are also accepted as an `Authorization: Bearer` header on HTTP API endpoints. #1009

### Breaking Changes
- The `/metrics` endpoint requires a user token with the new `core.controller.metrics` permission. Concurrent scrapes share a single gather from the hosts. #1009
```

### DIFF
--- a/docs/managing-a-cluster.md
+++ b/docs/managing-a-cluster.md
@@ -282,12 +282,13 @@ The controller exports metrics for the whole cluster on the `/metrics` endpoint.
 Fetching it requires a token for a user with the `core.controller.metrics` permission, sent either as an `Authorization: Bearer <token>` header or as an `X-Access-Token` header.
 Requests without a valid token get a 401 response and requests for users lacking the permission get a 403 response.
 
-Create a role with just this permission, a user to hold it, and generate a token for that user:
+Create a role with just this permission, a user to hold it, and generate a token for that user.
+Give the user a name that is not a valid Factorio player name, such as one starting with `#`, so a player joining the cluster can never claim the account and its token:
 
     ctl> role create Prometheus --permissions core.controller.metrics
-    ctl> user create prometheus
-    ctl> user set-roles prometheus Prometheus
-    $ npx clusteriocontroller bootstrap generate-user-token prometheus
+    ctl> user create #prometheus
+    ctl> user set-roles #prometheus Prometheus
+    $ npx clusteriocontroller bootstrap generate-user-token '#prometheus'
 
 Then pass the token in the scrape job for the controller in `prometheus.yml`:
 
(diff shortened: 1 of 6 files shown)


docs/managing-a-cluster.md
```diff
@@ -274,3 +274,28 @@ Remote updates are enabled by default but can be disabled on a per-machine basis
...
+## Prometheus Metrics
+
+The controller exports metrics for the whole cluster on the `/metrics` endpoint.
+Fetching it requires a token for a user with the `core.controller.metrics` permission, sent either as an `Authorization: Bearer <token>` header or as an `X-Access-Token` header.
+Requests without a valid token get a 401 response and requests for users lacking the permission get a 403 response.
+
+Create a role with just this permission, a user to hold it, and generate a token for that user:
+
+    ctl> role create Prometheus --permissions core.controller.metrics
+    ctl> user create prometheus
```
Cooldude2606: Because there could be a player named "prometheus" we need to create service accounts using an invalid character. Typically I would use `#` so `#prometheus`

Cooldude2606 (changes requested):

docs/managing-a-cluster.md
bbassie: Done in 171ee43b. The recipe now uses `#prometheus` and says why, and the shell line quotes it since `#` would otherwise start a comment.

Cooldude2606 (approved):

PR #1020, opened 2026-09-11
Add typed hook system for plugins

bbassie:
Recreates #921 on a clean branch from master, as suggested in the closing comment there. The content is Cooldude2606's hook refactor plus the follow-up work from Cooldude2606/clusterio#1 (tests, templates, docs), squashed into one commit and adapted to the changes master picked up since: the ESM migration (#984), `packagePath` on plugin infos, and permissions on `PluginDeclaration` (#988).

What it does, same as #921:

* Adds `AsyncHook` and hook collections for the controller, host, instance, clusterioctl and web control objects. Plugins attach handlers to these instead of overriding `on*` methods.
* Entrypoints export a default `async function(context)`. Class exports still load through `fromContext` and log a deprecation warning.
* Replaces the `plugins` map with `loadedPlugins` on the core classes and improves the typing of the plugin APIs and web extension points.
* The `@clusterio/create` templates and `docs/writing-plugins.md` use the entrypoint function.
* New tests for `AsyncHook`, `loadPlugin` and the base plugin classes. `test/file/test_plugin` uses the function entrypoint for host and instance and keeps the class export for controller and ctl so both paths are covered by the integration tests.

Changes made while rebasing onto master:

* The node only loader stays in `plugin_loader.ts` as on master, with `loadPlugin` replacing the old `loadPluginClass`. The hooks branch had moved it into `plugin.ts`, which is also compiled for the browser and rejects the JSON import attribute the ESM loader uses. `plugin.ts` now only holds types, so the web bundle no longer pulls `node:fs` in through it.
* `SaveModule.fromPluginInfo` uses `pluginInfo.packagePath` instead of `require.resolve`.
* The JS template's `index.js` imported `messages.js` as a default export, which does not exist, so a generated JS plugin failed to load. It now uses `import * as Messages`, matching the TS template.
* The docs said entrypoint paths should not end in `.js`, which stopped being true with ESM.

Not done, same as before: the built in plugins still use class exports and log the deprecation warning on load. `AsyncHook.invoke` runs handlers concurrently where `invokeHook` ran them in order.

Verified with `pnpm test` (1706 passing, the only two failures are the port 8080 conflict on my machine and the plugin web bundle missing in a fresh worktree, which passes once built), `pnpm lint`, the controller webpack build, and by generating a full feature plugin from both templates, type checking the TS one and loading every entrypoint of the JS one under Node.

### Changelog

```
### Features
- Added typed hook system for plugin registration. (#730)

### Changes
- Plugins should now return a function to register handlers via hooks. (#730)
- Plugin templates from `@clusterio/create` use entrypoint functions and hooks. (#730)
- Improved typing for plugin APIs and webui extension points. (#730)
- Plugin hooks run concurrently instead of one plugin at a time. (#730)

### Breaking Changes
- Replaced `plugins` map with `loadedPlugins` set across core classes (except WebUi). (#730)
```


docs/writing-plugins.md
```diff
@@ -241,13 +249,13 @@ See [Configuration System](config-system.md) for more details on how this system
...
-If the plugin requires a certain feature to be enabled to function it should throw an error during init if this is not the case.
+If the plugin requires a certain feature to be enabled to function it should throw an error from the entrypoint if this is not the case.
 The most common such feature is the save patching, which can be disabled to run vanilla or scenarios not compatible with Clusterio.
 For example:
 
 ```js
-async init() {
-    if (!this.instance.config.get("factorio.enable_save_patching")) {
+export default async function(context) {
+    if (!context.instance.config.get("factorio.enable_save_patching")) {
```
Cooldude2606: Since the addition of feature flags like `SavePatching` this is now a bad example and is discouraged. Even more generally, invalid configuration can most often be detected through the config validators.

docs/writing-plugins.md
```diff
@@ -296,17 +304,17 @@ The older `lib.definePermission()` function still works, but a plugin should not
...
 For pushing data into Factorio there's RCON, which lets you send arbitrary Lua commands to invoke whatever code you want in the game.
-This is done by calling the `sendRcon` method on the plugin object.
+This is done by calling the `sendRcon` method on the instance.
 For example:
 
 ```js
-async onStart() {
-    let response = await this.sendRcon(
+instance.hooks.start.attach(plugin.name, async () => {
+    let response = await instance.sendRcon(
```
Cooldude2606: Let's encourage the use of `const` rather than `let`

docs/writing-plugins.md
```diff
@@ -589,14 +598,14 @@ const barMetric = new Gauge(
...
-barMetric.labels(String(this.instance.id)).set(someValue);
+barMetric.labels(String(instance.id)).set(someValue);
 ```
 
 Metrics are automatically registered to the default registry, and this default registry is automatically polled by the controller on hosts.
 This means that it's important that you place the definition of the metric at module level so that it's not created more than once over the lifetime of a host.
 Since the metrics remember their values and would continue to be exported after an instance is shutdown, there's code at instance shutdown that removes all the values where the `instance_id` label matches the id of the instance shut down.
 
-For statistics you need to update on collection there's an `onMetrics` hook on both controller and instance plugins that is run before the metrics in the default registry are collected.
+For statistics you need to update on collection there's a `metrics` hook on both controller and instance plugins that is run before the metrics in the default registry are collected.
```
Cooldude2606: We should add an example for this because this hook allows returning a value.

packages/web_ui/src/util/websocket.ts
```diff
@@ -51,9 +51,13 @@ export class Control extends lib.Link {
 	/** Roles of the account this control link is connected as. */
 	accountRoles: lib.AccountRole[] | null = null;
 
-	/** Plugins loaded in the web interface */
-	public plugins = new Map<string, BaseWebPlugin>();
-	public inputComponents: Record<string, InputComponent> = {};
+	public hooks = new WebPlugin.WebHooks(logger);
+	public plugins = new Map<string, WebPlugin.BaseWebPlugin>();
```
Cooldude2606, line 56: Can we add a deprecated lint to `plugins`

Cooldude2606 (approved):
Everything is as I recall, but with a much cleaner diff. There are some very minor changes, happy to merge once fixed.

docs/writing-plugins.md
bbassie: Rewrote the section to point at the `features` flags (which fail config validation for `<plugin>.load_plugin` when save patching or script commands are off) and at field validators with `dependsOn` for other cases. Also listed `features` in the plugin declaration properties since it was not documented. 8363b221

docs/writing-plugins.md
bbassie: Done, both `let` in the examples are `const` now. 8363b221

docs/writing-plugins.md
bbassie: Added two examples: updating a gauge in the hook, and returning `registry.collect()` from a registry outside the default one. 8363b221

packages/web_ui/src/util/websocket.ts
bbassie: Added a `@deprecated` tag on `plugins`. It still holds the class based plugins so `AboutModal` and subspace_storage keep working. 8363b221

Cooldude2606 (approved):

PR #1023, opened 2026-09-12
Bump dependency versions

bbassie:
Second of the chores Cooldude asked for. Every caret range in the root package.json, the pnpm-workspace.yaml catalog, the per-package package.json files and the create plugin template is raised to the current latest within its major. The lockfile is not committed, so most of these were already what CI installed; the floors just catch up. Majors were taken only where the change was mechanical and the suite stayed green.

**Majors taken.** eslint 10 with @eslint/js 10 and globals 17 (FlatCompat still works), nyc 18, @types/express 5 (express itself was already 5), @types/node 24, finalhandler 2, pidusage 4, style-loader 4, css-loader 7, yargs 18 in all packages (every call site already used the `yargs(argv)` factory), chalk 6 in lib and create (both are ESM now).

**Majors skipped.** mocha 12 fails 59 integration tests here, starting with a 30 s timeout in the `clusteriocontroller run` tests, and the process hangs after printing its summary, so it stays at 11. inquirer 10+ is a rewrite where `ui.Prompt.prototype.close` no longer exists, which would silently disable the win32 readline workaround in create.js, and I cannot verify that on Linux. react, antd 6, react-router-dom 7, @ant-design/icons 6, rc-field-form 2, typescript 7 and typebox are real migrations rather than chores and are left alone. webpack-merge keeps its exact 5.9.0 pin.

**Code changes.** The two `/api/stream/:id` handlers in routes.ts are typed `Request<{ id: string }>` because @types/express 5 types `req.params` values as `string | string[]`. eslint 10.10 started reporting `new this.constructor()` under new-cap, so the config gets `newIsCapExceptions: ["constructor"]`.

**Two things noticed on the way, not changed here.** pidusage 4 has a native fallback for the removed wmic, so the PowerShell workaround in `packages/host/src/Instance.ts` could probably go. And `clusteriohost should auto start instances with auto_start enabled` failed once locally with an ENOENT on a `.tmp.json` chmod inside safeOutputFile, two concurrent saves of one config sharing the same temp name; it passed on every rerun and looks like a pre-existing race from #1017, worth a separate look.

Verified locally on Node 24.21.0 with Factorio 2.1.17: build, lint (0 errors, the 4 pre-existing unused-directive warnings) and the full suite (1753 passing, 1 failing which is the port 8080 test that always fails on my machine).

### Changelog
```
### Changes
- Bumped dependencies, including eslint 10, yargs 18, chalk 6, @types/express 5 and @types/node 24. #1023
```


Cooldude2606 (commented):
Can we move `yargs` `@types/yargs` and any other now common versions into the workspace catalogue.

bbassie:
Done in the second commit: yargs, @types/yargs, semver and @types/semver are in the catalog now, and the direct @swc/core and react-router-dom references in the root and web_ui package.json point at the existing catalog entries too. The create template keeps literal versions since it is copied into projects outside the workspace.

Cooldude2606 (approved):

PR #1026, opened 2026-09-12
Give each safeOutputFile call its own temporary file

bbassie:
Follow-up to the flake noted in #1023. `safeOutputFile` writes to `name.tmp.ext` and renames it over the target, but two calls for the same file at the same time share that temporary. The second write truncates and rewrites the first call's temporary, and once the first rename moves it away the second call has nothing left to rename, so it fails with ENOENT. There is also a narrower window where the first rename lands while the second write is still in progress, which puts a half-written file at the target and defeats the point of the temporary.

The host does exactly this on `clusteriohost should auto start instances with auto_start enabled`, where the host saves an instance's config while the instance saves it too. It surfaced as `ENOENT chmod 'alt-instances/alt-start/instance.tmp.json'`. The chmod from #1017 made the window a little wider, but the race predates it: without the chmod the same interleaving fails on the rename instead.

Each call now gets its own temporary, `name.<random>.tmp.ext`, created with the exclusive flag, so concurrent writers never touch each other's file and no state is shared between calls. Two consequences of that: the temporary is always a fresh file, so the mode is applied on creation and the chmod for reused leftovers goes away, and a failed rename removes its temporary since nothing will overwrite it later. A process dying mid-write still leaves one behind, as before.

Concurrent writers in separate processes were never a problem here since no file has more than one owning process. Ordering between concurrent calls is whichever rename lands last, which is no weaker than before, where the second call crashed.

The new tests fire twenty concurrent writes at one path (fails on master with the rename ENOENT) and check a failed rename leaves nothing behind. Full suite locally: 1755 passing, 1 failing (the port 8080 test that always fails on my machine).

### Changelog
```
### Fixes
- Fixed concurrent saves of the same file racing on a shared temporary file. #1026
```

### DIFF
--- a/packages/lib/src/file_ops.ts
+++ b/packages/lib/src/file_ops.ts
@@ -137,11 +137,11 @@ export async function getTempFile(prefix = "tmp.", suffix = "", tmpdir = "./") {
  * file.  If the operation fails because the parent directory does not exist, it
  * attempts to create the directory, including any ancestors and then retries
  * the operation.  The name of the temporary file is the same as the target file
- * with the suffix `.tmp` added before the extension.
+ * with a random tag and the suffix `.tmp` added before the extension, so
+ * concurrent writes to the same file do not interfere with each other.
  *
- * If the operation fails it may leave behind the temporary file.  This
- * should not be too much of an issue as the next time the same file is
- * written the temporary will be overwritten and renamed to the target file.
+ * If the operation fails the temporary file is removed.  Should the process
+ * die in the middle of writing the temporary file it is left behind.
  *
  * @param file - Path to file to write.
  * @param data - Content to write.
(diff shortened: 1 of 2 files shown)


Cooldude2606 (changes requested):
This solution introduces file level state, this creates singlton behaviour that will break when multiple versions of the library exist. This is not a good pattern to design around. Please can you look for an alternative solution.

bbassie:
Replaced the queue with a stateless version: every call writes its own `name.<random>.tmp.ext` (created with the exclusive flag) and renames it into place, so nothing is shared between calls or library copies. The chmod for reused leftovers is gone with it since the temporary is always new, and a failed rename now unlinks its temporary. Same tests plus one for the failed-rename cleanup; full suite green locally.

Cooldude2606 (approved):

PR #1028, opened 2026-09-13
Log errors from instance config change handlers instead of exiting the host

bbassie:
Changing some instance config fields at the wrong moment kills the whole host process, and every instance on that host goes down with it. `_configFieldChanged` in the `Instance` constructor ran `updateFactorioSettings` and `updateFactorioWhitelist` with only a `finally` attached, and `handleUnhandledErrors` exits on an unhandled rejection.

Both of these were reproduced against a real host running Factorio 2.1.17 (base game), using the requests the web UI config editor sends:

1. Send `InstanceStartRequest`, then straight away `InstanceConfigSetFieldRequest(id, "factorio.enable_whitelist", <flipped value>)`. `updateFactorioWhitelist` calls `sendRcon` before the server has launched and the host exits with `fatal Unhandled rejection: Error: Expected state running,stopping but state is new`. The window is the mod sync before launch, so a mod pack that has to be downloaded makes it wide.
2. On a running instance, set the `factorio.settings` prop `tags` to `[1]`, which the config editor accepts as valid JSON. `applyAsConfig` called `tag.replace` outside its try block and the host exits with `fatal Unhandled rejection: TypeError: tag.replace is not a function`.

Both handlers now log the error on the instance logger and still invoke the hook, and tags are converted with `String()`. I ran the same two sequences against the fuzz cluster with this branch: the host stays up and the instance log shows `Error updating whitelist: Expected state running,stopping but state is new`.

Not changed here, open for input:
- A whitelist toggle during startup is logged and not applied. The server reads `enableWhitelist` when the instance is constructed, so the running server keeps the old setting until the next restart.
- `factorio.settings` props are not validated. With `tags: [1]` saved, the next start fails in Factorio with `Value must be a string in property tree at ROOT.tags[0]`. That error is clear, so I left validation for the TODO in `definitions.ts`.

### Changelog
```
### Fixes
- Fixed the host exiting when an instance's whitelist or server settings config changed while it was starting, or when tags contained non-string values. #1028
```

### DIFF
--- a/packages/host/src/Instance.ts
+++ b/packages/host/src/Instance.ts
@@ -61,7 +61,7 @@ function applyAsConfig(name: string) {
 			// Replace spaces with non-break spaces and delimit by spaces.
 			// This does change the defined tags, but there doesn't seem to
 			// be a way to include a space into a tag from the console.
-			value = value.map(tag => tag.replace(/ /g, "\u00a0")).join(" ");
+			value = value.map(tag => String(tag).replace(/ /g, "\u00a0")).join(" ");
 		}
 		try {
 			await instance.sendRcon(`/config set ${name} ${value}`);
@@ -199,9 +199,13 @@ export default class Instance extends lib.Link {
 			if (field === "factorio.shutdown_timeout") {
 				this.server.shutdownTimeoutMs = curr as number * 1000;
 			} else if (field === "factorio.settings") {
-				this.updateFactorioSettings(curr as any, prev as any).finally(hook);
+				this.updateFactorioSettings(curr as any, prev as any).catch(err => {
+					this.logger.error(`Error updating server settings:\n${err.stack}`);
+				}).finally(hook);
 			} else if (field === "factorio.enable_whitelist") {
-				this.updateFactorioWhitelist(curr as any).finally(hook);
+				this.updateFactorioWhitelist(curr as any).catch(err => {
+					this.logger.error(`Error updating whitelist:\n${err.stack}`);
+				}).finally(hook);
 			} else {
 				if (field === "factorio.max_concurrent_commands") {
 					this.server.maxConcurrentCommands = curr as number;
--- a/test/host/Instance.js
+++ b/test/host/Instance.js
@@ -4,7 +4,7 @@ import path from "node:path";
 import * as lib from "@clusterio/lib";
 import { PlayerStats, wait } from "@clusterio/lib";
 import Instance from "@clusterio/host/dist/node/src/Instance.js";
-import { MockConnector, MockServer } from "../mock.js";
+import { MockConnector, MockLogger, MockServer } from "../mock.js";
 
 const addr = lib.Address.fromShorthand;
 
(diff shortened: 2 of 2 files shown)


Cooldude2606 (commented):
The current changes are good and solve the identified issue, these should remain in place. Writing a full schema for `factorio.settings` is out of scope for this PR. However, there is an underlying issue with config changed handler that needs to be addressed. That is, if the handler is called during a transitory state such as "starting" then it should wait until the instance is "running" or "stopped" before continuing, but we must be careful to not hang it. There may be other similar cases where we should be waiting for transitory states to settle. Please can you investigate a solution for this.

Cooldude2606 (changes requested):

bbassie:
Thanks, that pointed at the real problem. I've pushed a follow-up commit that leaves the existing catch/`String(tag)` changes in place and adds the waiting behaviour; no `factorio.settings` schema.

**What it does.** `notifyStatus` is the only place `_status` changes, and every transitory state eventually settles (`notifyExit` calls `notifyStatus("stopped")`), so I hooked the wait there: a small per-instance list of waiters is resolved whenever the status becomes `running` or `stopped`. Work that needs a live server now goes through a helper that waits for the instance to settle and then only runs if it settled to `running`.

**Why it can't hang.** Nothing awaits the deferred work. The plugin hook is now invoked immediately for `factorio.settings` and `factorio.enable_whitelist` instead of being chained with `.finally`, which matches what every other config field already does and means a slow startup can no longer delay plugin notification. The list-update handlers are event handlers, and `Link._processEvent` attaches a `.catch` rather than awaiting, so waiting there doesn't block message processing either. The only way a waiter never resolves is a Factorio start that never finishes, and in that case there is nothing to apply anyway and the instance object is discarded on exit.

**Why settling to `stopped` skips the work.** There is no server to talk to, and the value is not lost: `factorio.settings` is written into server-settings.json by `writeServerSettings` at the next start, the admin/ban/whitelist files are written by `prepare()`, and `enableWhitelist` is re-read when the next `Instance` is constructed. So a config change made while stopped is now a silent no-op instead of a logged RCON error.

**Other similar cases.** I applied the same wait to the three controller-pushed list handlers (`handleInstanceAdminlistUpdateEvent`, `handleInstanceBanlistUpdateEvent`, `handleInstanceWhitelistUpdateEvent`). Those had the same shape and were worse: during `starting` the RCON rejection was swallowed by `_processEvent`'s catch, so promoting/banning/whitelisting someone while an instance was starting was silently dropped. They now apply once the instance is running.

I left these alone but they are the same class of problem, happy to do them in a follow-up if you want:

- `Host.handleInstanceUnassignInternalRequest` only sends a stop for `starting`/`running`. Unassigning during `creating_save` or `exporting_data` removes the instance from `assignedInstances` while Factorio keeps running and the connection stays in `instanceConnections`. I hit this while testing: the save finished, but the controller then rejected the reply because the instance was no longer assigned there, so the original request never resolved.
- `Instance.stop()` during early `starting` (server state still `new`/`init`) takes neither branch, so the status goes to `stopping` but nothing stops and the instance carries on to `running`. That is the #879 case.
- `Instance.kill()` in the same window throws `Expected state running,stopping,create but state is new` out of `server.kill()`.
- `InstanceManager.assignInstance` on the controller has no status gate at all, which is what lets the first case be reached.

**On how narrow the window actually is**, in case it affects how you want to weigh this: `prepare()` writes server-settings.json and the admin/ban/whitelist files before launching, and `server._state` becomes `running` the moment the process spawns, after which `sendRcon` just waits for rcon-ready and succeeds. So a change landing before `prepare()` is picked up by the written files, and one landing after the spawn applies over RCON regardless. The genuinely broken window is the sliver in between, which is milliseconds on a small save with cached mods but widens to seconds whenever `syncMods` has to fetch mods. That sliver is where the crash happened and where updates were silently dropped.

**Verified** against a real host on Factorio 2.1.17, toggling `factorio.enable_whitelist` immediately after a start request. Before this commit the instance log shows `Error updating whitelist: Expected state running,stopping but state is new` and the update is lost (the instance still reaches `running`). With the commit the same sequence logs no error and the update is applied once the instance is running. Unit tests cover deferral to `running`, dropping when it settles to `stopped`, the no-op while stopped, a deferred adminlist update, and the hook still firing promptly.

packages/host/src/Instance.ts
```diff
@@ -199,9 +205,13 @@ export default class Instance extends lib.Link {
 			if (field === "factorio.shutdown_timeout") {
 				this.server.shutdownTimeoutMs = curr as number * 1000;
 			} else if (field === "factorio.settings") {
-				this.updateFactorioSettings(curr as any, prev as any).finally(hook);
+				this._applyWhenRunning(
+					"server settings", () => this.updateFactorioSettings(curr as any, prev as any)
+				);
+				hook();
```
Cooldude2606: It seams odd that `hook` is not called for `factorio.shutdown_timeout`. I think we should make the hook invoke unconditional and not in the branches.

Cooldude2606 (changes requested):
Please fix the one issue below and the conflicts. Then for the other issues we can use a new PR as those have much wider impact.

bbassie:
Good catch, and you are right that it reads oddly. The hook is now invoked once, unconditionally, after the field specific side effect, rather than repeated inside the branches. That also fixes a pre-existing bug rather than just tidying: `factorio.shutdown_timeout` never invoked the hook at all, on master or on this branch, so plugins were never told when it changed.

```ts
this._configFieldChanged = (field: string, curr: unknown, prev: unknown) => {
	if (field === "factorio.shutdown_timeout") {
		this.server.shutdownTimeoutMs = curr as number * 1000;
	} else if (field === "factorio.max_concurrent_commands") {
		this.server.maxConcurrentCommands = curr as number;
	} else if (field === "factorio.settings") {
		this._applyWhenRunning(
			"server settings", () => this.updateFactorioSettings(curr as any, prev as any)
		);
	} else if (field === "factorio.enable_whitelist") {
		this._applyWhenRunning("whitelist", () => this.updateFactorioWhitelist(curr as any));
	}
	this.hooks.instanceConfigFieldChanged.invoke(field, curr, prev);
};
```

One thing I checked before leaving the bare `invoke(...)` unawaited, since an unhandled rejection here is exactly the bug this PR exists to fix: `AsyncHook._invokeHandler` catches and logs every handler error and `invoke` only awaits a `Promise.all` of those, so it cannot reject. The `_applyWhenRunning` calls stay fire and forget and the hook is deliberately not chained onto them, so a slow startup still cannot delay plugin notification.

Also merged master in, which was clean. #1034's guards in `_watchServerLogActions` and the changes here do not overlap.

New tests cover the hook firing for `factorio.shutdown_timeout` and exactly one invoke per change across all four fields. `test/host/Instance.js` is 41 passing, and I re-ran the whitelist-during-startup reproduction against a real host to confirm the deferral still behaves after the restructure.

On the other issues, agreed, a separate PR makes sense given the reach. For reference these are the ones I found while looking at this, all in the same "transitory state not accounted for" family:

- `Host.handleInstanceUnassignInternalRequest` only sends a stop for `starting`/`running`, so unassigning during `creating_save` or `exporting_data` drops the instance from `assignedInstances` while Factorio keeps running and the connection stays registered.
- `Instance.stop()` during early `starting` takes neither branch, because `server._state` is still `new`/`init`, so the status goes to `stopping` but nothing stops and the instance carries on to `running`. That is #879.
- `Instance.kill()` in that same window throws `Expected state running,stopping,create but state is new` out of `server.kill()`.
- `InstanceManager.assignInstance` has no status gate at all, which is what lets the first one be reached.

Happy to open that as its own PR, together with the web UI side of disallowing stop in those states if you want them together.

Cooldude2606 (approved):

PR #1029, opened 2026-09-13
Keep instances assignable after a failed assign or create

bbassie:
Two inputs leave an instance whose `instance.assigned_host` points at a host that does not have it. Assigning it to that host again then returns success and does nothing, because `assignInstance` returns early when the host is unchanged. The web UI assign modal preselects the current host, so clicking Assign looks like it worked.

1. The new host fails `InstanceAssignInternalRequest`. I reproduced this with a real host whose instances directory was read-only (`EACCES: permission denied, mkdir 'instances/a3'`). The controller had already set `assigned_host`, so the instance was left with `assigned_host=10` and status `unassigned`. Retrying after fixing the permissions returned ok, and starting it failed with `Instance with ID 3001 does not exist`.
2. `InstanceCreateRequest` with `"instance.assigned_host": 10` in the config was accepted and gave the same state. The config set requests already reject this field.

A failed send now leaves the instance unassigned before rethrowing, so the assign can be retried. Create rejects a non-null `instance.assigned_host` with the same message as the set requests. Null is still accepted so a client that serialises a whole config keeps working. Neither the web UI nor ctl sends the field on create.

With this branch running as the controller of the same test cluster, the failed assign leaves the instance unassigned, the retry assigns it to host 10 with status `stopped`, and the create request is rejected.

One trade-off to look at: if the host applied the assign but the reply was lost, the controller now treats the instance as unassigned while the host has it. When that host reconnects its `InstancesUpdateRequest` gets an unassign for the instance, the same way other mismatches are resolved.

### Changelog
```
### Fixes
- Fixed instances getting stuck on a host after a failed assign, and instance create requests no longer accept `instance.assigned_host`. #1029
```

### DIFF
--- a/packages/controller/src/ControlConnection.ts
+++ b/packages/controller/src/ControlConnection.ts
@@ -288,6 +288,9 @@ export default class ControlConnection extends BaseConnection {
 	}
 
 	async handleInstanceCreateRequest(request: lib.InstanceCreateRequest) {
+		if ((request.config["instance.assigned_host"] ?? null) !== null) {
+			throw new lib.RequestError("instance.assigned_host must be set through the assign-host interface");
+		}
 		const instanceConfig = new lib.InstanceConfig("controller");
 		if (request.cloneFromId) {
 			const baseInstance = this._controller.instances.get(request.cloneFromId);
--- a/packages/controller/src/InstanceManager.ts
+++ b/packages/controller/src/InstanceManager.ts
@@ -208,9 +208,17 @@ export default class InstanceManager {
 		instance.config.set("instance.assigned_host", hostId);
 
 		if (newHostConnection) {
-			await newHostConnection.send(
-				new lib.InstanceAssignInternalRequest(instanceId, instance.config.toRemote("host")),
-			);
+			try {
+				await newHostConnection.send(
+					new lib.InstanceAssignInternalRequest(instanceId, instance.config.toRemote("host")),
+				);
+			} catch (err) {
+				// Leave unassigned so the assign can be retried
+				instance.config.set("instance.assigned_host", null);
+				instance.status = "unassigned";
+				this.records.set(instance);
+				throw err;
+			}
 		} else {
 			instance.status = "unassigned";
 		}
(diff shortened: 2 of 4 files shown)


Cooldude2606 (approved):
On your two questions, `null` is the default and only allowed value for `assigned_host` when creating an instance, so it is fine to accept it as you have. As for the reply being lost, this is a rare edge case and so long as the state on reconnection settles to a consistent state (as your solution does) then we are good.

Cooldude2606 (changes requested):
It appears that the conflicts had material impact that need to be resolved before I can merge.

bbassie:
Thanks for the answers on both questions, and for updating the branch. Merged master in again on top of your merge commit rather than rebasing, so your update stays in the history.

The branch had gone stale rather than broken. Your merge resolved the #1027 collision correctly, but master has picked up #1030 and #1032 through #1035 since, and `test/controller/ControlConnection.test.js` gained two more describe blocks in that time, which is what re-conflicted.

That test file was the only conflict. I resolved it as a union: every block from master kept, this branch's `.handleInstanceCreateRequest()` block alongside them, and the imports merged into one statement per module rather than duplicated, since `no-duplicate-imports` is an error in the eslint config here. The diff of that file against master is now a single changed line (the import) and otherwise only additions, so none of the merged tests were disturbed.

One thing in the merge commit is not pure conflict resolution, so flagging it: two closing braces in the `.handleInstanceCreateRequest()` block were indented with spaces rather than tabs, which `@stylistic/indent` rejects. That was already the case before this merge, so it would have failed lint on the PR either way. I converted them to tabs. Happy to split that into its own commit if you would rather the merge stayed resolution only.

Neither source change was affected by any of this. `ControlConnection.ts` merges cleanly because the guard sits at the top of `handleInstanceCreateRequest` while the recent changes are in the constructor, `handleUserDeleteRequest` and `handleDebugDumpWsRequest`. `InstanceManager.assignInstance` on master is still byte for byte what this branch was cut from, so the try/catch applies unchanged.

I also re-ran the original reproduction against a real controller and host: an assign that fails because the host cannot create the instance directory now leaves the instance unassigned rather than stranded on that host, retrying the same assign afterwards succeeds and the instance comes up stopped, and a create carrying `instance.assigned_host` is rejected.

Cooldude2606 (approved):

PR #1030, opened 2026-09-13
Terminate control connections of deleted users

bbassie:
A deleted user keeps every permission for as long as their WebSocket session stays open, because `handleUserDeleteRequest` removed the user record but left their control connections running. Each ControlConnection holds its `User` object and checks permissions against it, so nothing changes for the live session. `handleUserRevokeTokenRequest` right above already terminates the user's connections; delete did not.

Reproduced against a running controller: create "victim" with an admin role, connect as victim, then have another admin send UserDeleteRequest("victim"). Before the fix victim stayed connected and could still list users and create and promote a new user "mallory".

After deleting, terminate the control connections belonging to the deleted user, the way revoke does. A user deleting itself is deferred one tick so the response goes out before its own session drops, matching how the web UI already treats revoking your own token as a SessionLost.

Not changed here: a user recreated with the same name accepts a token issued before the deletion, because `tokenValidAfter` starts at 0 on the new record. Fixing that cleanly means stamping `tokenValidAfter` in the create paths rather than in `createUser` (which bootstrap and the test harness rely on). Happy to do it in a follow-up.

### Changelog
```
### Fixes
- Fixed deleted users keeping their permissions on any open connection until they disconnected. #1030
```


Cooldude2606 (changes requested):
I think we should have consistency between revoke and delete, so lets have it be immediate. And for your second issue around `tokenValidAfter`, I agree that we need to set this to "now" rather than 0; however, it should definatlly be within `createUser` so all paths are covered. This should not impact `create-admin` since the token will be used after its creation, and the tests need to be updated to deal with this new default rather than relying on pre-signed tokens.

bbassie:
Both done, and I rebased onto master first since #1027 landed in the same two files and the PR had gone conflicting.

**Immediate terminate.** The self-delete special case is gone, so the loop is now the same shape as `handleUserRevokeTokenRequest`.

**`tokenValidAfter` in `createUser`.** Stamped there as you asked, so every path that creates a user is covered. `bootstrap create-admin` only creates the user, and `generate-user-token` and `create-ctl-config` are separate later invocations that sign after it exists, so they are unaffected. Since the check is `iat < tokenValidAfter`, a token signed in the same second as creation still works.

**Tests that relied on pre-signed tokens.** `test/integration/index.js` now declares `export let controlToken;` and assigns it in the root `before` hook right after `bootstrap create-admin test`. `test/integration/link.js` imports that instead of signing its own; its host tokens are untouched. The two sites in `clusterio.js` that sign tokens sign them at test runtime after the user exists, and both still pass.

**One consequence I fixed in the same PR.** With delete now immediate, an admin deleting their own account got a `SessionLost` rejection and the web UI showed "Error deleting user" for a delete that had worked. `UserViewPage` now handles that the way revoke already did. ctl has the same gap, but its revoke does not handle `SessionLost` either, so delete and revoke at least behave the same there; say the word if you want ctl done too.

**Verified against a real controller**, comparing master with this branch:

| | master | this branch |
|---|---|---|
| deleting your own account | session stayed open | `SessionLost`, session closed |
| deleting another connected user | session stayed open and their next request succeeded | session closed |
| token issued before a delete, used after recreating the same name | accepted | `Authentication failed: invalid token` |

Unit tests are 202 passing for `test/controller`. The integration suite is 274 passing with one failure, `clusterioctl run should start when no users are loaded if bypass option given`, which spawns a controller with no config so it binds the default port 8080 that is occupied on this machine. It fails the same way without these changes.

Two small things worth knowing. Users created from now on serialize `token_valid_after`, since `toJSON` emits it when non-zero; existing users keep `0`. And the recreation unit test compares against `tokenValidAfter - 60` rather than a genuinely older signed token, because `signUserToken` stamps `iat` in the same second as creation, so it exercises the comparison `WsServer` makes rather than a real old token. The live check above covers the real thing.

Cooldude2606 (approved):

PR #1031, opened 2026-09-13
Resolve instance stop when Factorio exits before RCON is ready

bbassie:
`FactorioServer.stop()` waits for the RCON connection when it isn't ready yet with `await new Promise(resolve => this.once("rcon-ready", resolve))`. If the Factorio process exits before RCON connects, that event never fires and stop() never resolves, so the control request that triggered the stop hangs until the session closes.

Reproduced against a real host: send InstanceExportDataRequest to a stopped instance, then InstanceStopRequest while it is exporting. `exportData` runs the server through `startScenario` and kills it before RCON connects, so `Instance.stop()` calls `server.stop()` which waits on rcon-ready forever. The export finishes and the instance reports stopped, but the stop request was still pending after 20 s. The same wait applies to Stop pressed while an instance is starting if Factorio dies before RCON connects (for example invalid server-settings.json), which the web UI offers for status "starting".

Wait on both rcon-ready and exit, removing whichever listener did not fire. The existing `if (this._state !== "stopping") return;` right after handles the exit case, since the exit handler has already reset the state.

`sendRcon` has the same rcon-ready wait and can hang the same way (a concurrent request while the server is being killed during export); that needs a slightly different change and is left for a follow-up.

### Changelog
```
### Fixes
- Fixed a stop request or RCON command hanging when the Factorio process exited before its RCON connection was established. #1031
```

### DIFF
--- a/packages/host/src/server.ts
+++ b/packages/host/src/server.ts
@@ -1405,7 +1405,19 @@ export class FactorioServer extends events.EventEmitter<FactorioServerEvents> {
 		// complete before the RCON connection can be used
 		if (!this._rconReady) {
 			// Not using events.once here to avoid throwing on error events.
-			await new Promise<void>(resolve => this.once("rcon-ready", resolve));
+			// The process may also exit before RCON connects.
+			await new Promise<void>(resolve => {
+				const onRconReady = () => {
+					this.off("exit", onExit);
+					resolve();
+				};
+				const onExit = () => {
+					this.off("rcon-ready", onRconReady);
+					resolve();
+				};
+				this.once("rcon-ready", onRconReady);
+				this.once("exit", onExit);
+			});
 		}
 
 		// The Factorio server may have decided to get ahead of us and
--- a/test/host/server_stop.js
+++ b/test/host/server_stop.js
@@ -1,0 +1,62 @@
+import assert from "node:assert/strict";
+import events from "node:events";
+import path from "node:path";
+
+import { FactorioServer } from "@clusterio/host/dist/node/src/server.js";
+
+
+// Stand-in for the Factorio child process, only exit and kill are used here.
+class FakeProcess extends events.EventEmitter {
+	killed = false;
+	kill() {
+		this.killed = true;
+	}
+}
+
+function createStartedServer() {
+	const server = new FactorioServer(path.join("test", "file", "factorio"), path.join("temp", "test", "server"), {});
+	server._state = "running";
+	server._rconReady = false;
+	server._server = new FakeProcess();
+	server._watchExit();
+	return server;
+}
+
+// Fail instead of hanging if stop() never resolves.
+async function waitForStop(stopped) {
+	let timeoutId;
+	const timeout = new Promise((resolve, reject) => {
+		timeoutId = setTimeout(() => reject(new Error("stop() did not resolve")), 500);
+	});
+	try {
+		await Promise.race([stopped, timeout]);
+	} finally {
+		clearTimeout(timeoutId);
+	}
+}
+
+describe("host/src/server", function() {
+	describe("FactorioServer.stop()", function() {
+		it("returns if the process exits before RCON is ready", async function() {
+			const server = createStartedServer();
+			const stopped = server.stop();
+			server._server.emit("exit", 1, null);
+			await waitForStop(stopped);
+			assert.equal(server._state, "init");
+			assert.equal(server.listenerCount("rcon-ready"), 0);
+		});
+
+		it("stops the server if RCON becomes ready", async function() {
+			const server = createStartedServer();
+			const stopped = server.stop();
+			server._rconReady = true;
+			server.emit("rcon-ready");
+			await new Promise(resolve => setImmediate(resolve));
+			assert.equal(server.listenerCount("exit"), 0);
+			const fakeProcess = server._server;
+			assert.equal(fakeProcess.killed, true);
+			fakeProcess.emit("exit", 0, null);
+			await waitForStop(stopped);
+		});
+	});
+});


Cooldude2606 (changes requested):
Current code is good. Can you include the fix for `sendRcon` here as it is the same root cause. I am unsure which "server-state wait" you mean within `Instance.stop` but it does seam like a good idea to disallow stop (and only allow kill) when in these states. This will require updating the web ui so should be done in a seperate PR.

bbassie:
Added the `sendRcon` fix here as you asked, and pulled the paired wait into a small private helper so `stop()` and `sendRcon()` share it rather than repeating the same twelve lines.

`sendRcon` used `events.once(this, "rcon-ready")`, which does reject on an `"error"` event, so at first glance an unexpected exit looks covered. It isn't: `_watchExit` deliberately skips the `"error"` emit while `_state === "stopping"`, and `kill()` sets `stopping` before sending SIGKILL. That is exactly what `exportData` does, so a `sendRcon` waiting for RCON during an export never settled. Now the wait ends on either `rcon-ready` or `exit`, and since `_resetState()` has already nulled `_rconClient` by then, the existing `if (!this._rconClient)` check turns it into an "RCON connection lost" rejection instead of a hang.

**One behaviour change worth your opinion.** On an *unexpected* exit, `_watchExit` does emit `"error"`, and `events.once` used to reject `sendRcon` with that `EnvironmentError` ("Factorio server unexpectedly shut down with code N"). Now the exit arm wins and the caller gets the flatter "RCON connection lost". It is still a rejection rather than a hang, and the descriptive cause is not lost because `Instance` keeps a permanent `"error"` listener that logs it, but the text the caller sees is worse in that one path. `stop()` has had this property since the first commit, so the two are at least consistent. I can make the helper report which arm fired so `sendRcon` rethrows the richer error, if you would rather have that.

**Verified against a real host** on Factorio 2.1.17, by sending an RCON command while an instance is exporting data. On master the request was still pending after 30 seconds; with this branch it rejects after about 1.6 seconds with "RCON connection lost", and the export still completes normally.

**Sorry about the "server-state wait" phrase**, that was from my own notes and not something I should have left in a sentence aimed at you. What the PR body meant is the `if (this._state !== "stopping") { return; }` guard immediately after the wait in `stop()`. When the wait ends because the process exited, `_watchExit` has already called `_resetState()`, which puts `_state` back to `"init"`, so that guard makes `stop()` return rather than trying to quit a server that is already gone.

**On disallowing stop in transitory states**, agreed, and agreed it should be its own PR since the web UI has to change with it. For what it's worth the current behaviour there is worse than just an error: `Instance.stop()` during early `starting` takes neither branch, because `server._state` is still `new`/`init`, so the status flips to `stopping` but nothing actually stops and the instance carries on to `running`. Pressing Kill in that same window throws `Expected state running,stopping,create but state is new` out of `server.kill()`. Happy to pick that up separately, together with the web UI side, if you want it.

### Changelog
```
### Fixes
- Fixed a stop request or RCON command hanging when the Factorio process exited before its RCON connection was established. #1031
```

Cooldude2606:
I agree that the flatter "RCON connection lost" is a behaviour regression as the wait exists because no rcon connection was present; we should restore the previous message from `EnvironmentError`.

bbassie:
Agreed, and restored. `_waitForRconReadyOrExit()` now returns the error that came with the exit, if there was one, and `sendRcon` rethrows it. So an unexpected exit surfaces the original `EnvironmentError` again, for example "Factorio server unexpectedly shut down with code 1", and the flat "RCON connection lost" is now only reached on the kill path, where `_watchExit` deliberately emits no error because the state is already `stopping`. That is the case where there genuinely is no richer reason to report.

The wait attaches a plain `on("error")` listener that only records the last error, with one cleanup removing all three listeners whichever arm fires. `_watchExit` emits `"error"` before `"exit"`, so the error is always recorded by the time the exit arm resolves.

I left `stop()` returning rather than rethrowing. Its contract is that the process is gone, and after an exit the `if (this._state !== "stopping") { return; }` guard already treats that as success, so throwing would turn a completed shutdown into a failure for every caller. The error is not lost either way, since `Instance` keeps a permanent `"error"` listener that reports it.

One deliberate difference from the pre-PR behaviour, in case you would rather have it the other way. `events.once` rejected on any `"error"` event; the new wait only reports one if an exit actually follows, and otherwise keeps waiting for `rcon-ready`. I checked every `emit("error")` in `server.ts` and they are all paired with an `emit("exit")` except the two catch blocks in `create` and `startScenario`, which run before `_watchExit` is installed and so cannot be racing an RCON wait. If you would rather any error reject immediately, that is a one line change.

Master merged in cleanly; it had not touched `server.ts`.

`test/host/server_stop.js` is 5 passing, covering the unexpected exit surfacing the original message, the kill path still giving "RCON connection lost", and `stop()` still returning in both. Reverting just the source makes the new case fail with the flat message, as expected. I also re-ran the export reproduction against a real host to confirm that path still settles quickly rather than hanging.

packages/host/src/server.ts
```diff
@@ -1318,6 +1318,34 @@ export class FactorioServer extends events.EventEmitter<FactorioServerEvents> {
...
+			const onError = (err: Error) => {
+				exitError = err;
+			};
+			const onRconReady = () => {
+				cleanup();
+				resolve(undefined);
+			};
+			const onExit = () => {
+				cleanup();
+				resolve(exitError);
```
Cooldude2606: It seams odd to resolve to an error, should we use `reject` here?

bbassie:
Fair point, resolving with an error is an odd shape. Changed it to reject.

The reason it was a resolve is that the helper has two callers that want opposite things: `sendRcon` wants the error to propagate, and `stop()` must not throw, since its job is done once the process is gone. Returning the error let each caller decide, but it hid that intent inside a convention. With a rejection the intent is visible at each call site instead, which does read better.

The one thing worth looking at in this diff is `stop()`. It now needs an explicit catch, because without one a shutdown that used to succeed would start failing for every caller. That is the regression this change could introduce, so there is a test asserting `stop()` still resolves when the process exits with an error emitted, and I checked it fails if the catch is removed.

Observable behaviour is unchanged. An unexpected exit still surfaces the original `EnvironmentError`, the kill path still ends as "RCON connection lost", and `stop()` still returns in both cases. The helper rejects with the error that came with the exit when there was one, and with `Error("RCON connection lost")` when there was not, which is exactly what `sendRcon` used to throw a line later, so its call site is now a bare `await`.

I kept the `if (!this._rconClient)` throw below it. It is still reachable on the resolve arm: the RCON client's `error` and `end` handlers and the `connect()` catch each null the client while setting `_rconReady` and emitting `rcon-ready`, so the wait can resolve with no client.

`test/host/server_stop.js` is 6 passing, and I re-ran the export reproduction against a real host: an RCON command sent during an export still rejects in about 1.6 seconds with "RCON connection lost" rather than hanging, and the export still completes.

Cooldude2606 (approved):

PR #1036, opened 2026-09-13
Reject save detail updates for instances not on the sending host

bbassie:
`HostConnection.handleInstanceSaveDetailsUpdatesEvent` trusts the `instanceId` in the event and in each `SaveDetails` without checking the instance is assigned to the host that sent the event. `InstanceSaveDetailsUpdatesEvent` allows a host source, and `validateIngress` only checks the source host id, not the payload, so any authenticated host can rewrite the controller's save records for an instance on a different host:
- send saves for another instance's id and they get stored in `controller.saves`;
- send empty `updates` with another instance's id and the delete-missing sweep wipes that instance's entire controller-side save list.

Confirmed against a running cluster: a second host that owns nothing added `HACKED.zip` to an instance running on the first host, then emptied that instance's save list. The real files on the owning host are untouched, but the controller's view is corrupted until that host resyncs.

Add an ownership check: if `event.instanceId`, or any save in `event.updates`, is for an instance not assigned to this host, log a bogus-update warning (matching `handleInstanceStatusChangedEvent`) and drop the event. The normal path, a host reporting saves for its own instance, is unchanged.

I drop the whole event on any foreign entry rather than filtering per entry: a well-behaved host reports one instance's saves at a time, so a foreign entry means a buggy or hostile host, and dropping everything keeps the delete-missing sweep simple. Switching to a per-entry filter later is small if incremental multi-instance updates are ever needed.

Related, not fixed here: `handleLogMessageEvent` spreads `event.info` and only overrides `host_id`/`host_name`, so a host can tag log entries with any `instance_id`. Left for a separate decision.

### Changelog
```
### Fixes
- Fixed a host being able to modify the controller's stored save list for, or forge log messages attributed to, instances assigned to other hosts. #1036
```

### DIFF
--- a/packages/controller/src/HostConnection.ts
+++ b/packages/controller/src/HostConnection.ts
@@ -296,6 +296,23 @@ export default class HostConnection extends BaseConnection {
 	}
 
 	async handleInstanceSaveDetailsUpdatesEvent(event: lib.InstanceSaveDetailsUpdatesEvent) {
+		// A host may only report saves for instances assigned to it. Otherwise
+		// any host could inject save records for, or wipe the save list of, an
+		// instance on a different host. See handleInstanceStatusChangedEvent.
+		const assignedHere = (instanceId: number) => {
+			const instance = this._controller.instances.get(instanceId);
+			return instance !== undefined && instance.config.get("instance.assigned_host") === this.id;
+		};
+
+		if (event.instanceId !== undefined && !assignedHere(event.instanceId)) {
+			logger.warn(`Got bogus save updates for instance id ${event.instanceId}`);
+			return;
+		}
+		if (event.updates.some(save => !assignedHere(save.instanceId))) {
+			logger.warn(`Got bogus save updates from host id ${this.id}`);
+			return;
+		}
+
 		const updates: lib.SaveDetails[] = [];
 		const deletes: lib.SaveDetails[] = [];
 		for (const save of event.updates) {
--- a/test/controller/HostConnection.test.js
+++ b/test/controller/HostConnection.test.js
@@ -1,0 +1,78 @@
+import assert from "node:assert/strict";
+import * as lib from "@clusterio/lib";
+import { HostConnection } from "@clusterio/controller";
+
+function makeSave(instanceId, name) {
+	return new lib.SaveDetails(instanceId, "file", name, 0, 0, false, false, 0, false);
+}
+
+// Mock saves datastore backed by a Map keyed on SaveDetails.id.
+class MockSaves {
+	constructor(saves = []) {
+		this.saves = new Map(saves.map(s => [s.id, s]));
+	}
+
+	get(id) { return this.saves.get(id); }
+
+	values() { return this.saves.values(); }
+
+	setMany(saves) { for (const s of saves) { this.saves.set(s.id, s); } }
+
+	deleteMany(saves) { for (const s of saves) { this.saves.delete(s.id); } }
+}
+
+// Mock instances manager where instance id -> assigned host id.
+function makeInstances(assignments) {
+	return {
+		get(id) {
+			if (!assignments.has(id)) {
+				return undefined;
+			}
+			const host = assignments.get(id);
+			return { config: { get: name => (name === "instance.assigned_host" ? host : undefined) } };
+		},
+	};
+}
+
+describe("controller/src/HostConnection", function() {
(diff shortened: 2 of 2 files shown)


Cooldude2606 (changes requested):
Lets include the assignment check for log message event as it is the same root cause.

bbassie:
Added the same assignment check to log messages, and lifted the `assignedHere` closure into a shared private method so both handlers use it.

Worth spelling out why the log case was exploitable, since the schema makes it look like it should not be. `LogMessageEvent.jsonSchema` declares `info` as only `{ level, message }`, but the Ajv instance in `lib/src/schema.ts` is built without `removeAdditional` and the schema has no `additionalProperties: false`, so unknown keys are neither stripped nor rejected, and `fromJSON` passes `json.info` straight through. That is not hypothetical: real instance log lines arrive carrying `instance_id`, `instance_name`, `host_id` and `host_name`, which is how instance logs get attributed at all.

**I sanitise rather than drop**, which is the one place this differs from the save handler, so please push back if you would rather have consistency. I strip `instance_id` and `instance_name` and keep the line as a host level message. The reassignment race is real and already noted in this file: the comment in `handleInstanceStatusChangedEvent` says updates arrive for instances "reassigned or deleted while the connection to the host it was originally on was down". Dropping would silently discard a host's output during exactly the migration you would want to read afterwards. Demoting also grants the host nothing it did not already have, since it can log arbitrary host level lines anyway, so the only thing taken away is the false attribution. A bogus save record is different because it corrupts stored state, so rejecting outright is right there.

The warning fires once per connection, since a misbehaving host could otherwise log in a loop.

**Verified against a real controller.** A second host that owns nothing sent a log message tagged with an instance assigned to the first host. On master the controller stored it as `{"host_id":40,"instance_id":1002,"instance_name":"i2","level":"error",...}`, so the line was attributed to another host's instance while carrying a `host_id` that does not own it. With this branch the same message is stored as `{"host_id":40,"level":"error",...}`, keeping the host's output but dropping the false attribution.

On the consumer side: nothing in `plugins/` handles log entries, and `LogConsole` in the web UI renders only level and message. Routing happens in `lib.logFilter`, which keys on `instance_id`, so a stripped line moves from an instance subscription to a host subscription rather than disappearing. `WebConsoleFormat` prints `instance_name` only when `instance_id` is set, which is why both are removed together.

One thing I left alone: on the accepted path I do not override `instance_name` from the controller's own instance record, so a host can still mislabel the name of an instance it legitimately owns. It seemed out of scope, but it is a one line change if you want it.

`test/controller` is 206 passing.

Cooldude2606:
I think logging a warning for a misaligned name would be beneficial, as well as correcting it based on the controller's state. Also as host connections can be long lived, a single warning will get quickly burried, we should have a timeout which resets this flag after 5 minutes. For clarity, the misaligned name should use the same flag to stop warning spam.

bbassie:
All three done.

The name a host sends for an instance it does own is now checked against the controller's record and corrected when it differs, warning with both values. The lookup that decides assignment returns the instance record now rather than a boolean, so the authoritative name comes from that same lookup instead of a second one.

The warn-once flag became a timestamp, `_lastLogMessageWarnMs`, compared against `Date.now()` with a five minute interval, so warnings resurface on a long lived connection instead of latching for its lifetime. It is a timestamp rather than a `setTimeout` so there is no timer to clean up on disconnect and nothing keeping the process alive. Both warnings, bogus instance id and misaligned name, go through one `_warnLogMessage` helper and so share the suppression, as you asked.

Verified against a real controller with a host that legitimately owns an instance. A line carrying the correct name is logged untouched; a line claiming `TOTALLY-WRONG` for the same instance is stored as `owned-by-50`, with one warning: `Got log message for instance id 5050 named TOTALLY-WRONG which is named owned-by-50 on the controller`. `test/controller` is 220 passing, and reverting just the source makes the three new cases fail.

One question on the edge of your request. A line that carries `instance_id` but no `instance_name` at all now counts as misaligned: the name gets filled in from the controller and it warns, so the message reads "named undefined". Filling it in seems right either way, but I am not sure the warning is. Every real instance log line I have seen carries both fields, so this should not fire in practice, and I left it warning on the basis that it is a host sending something that disagrees with the controller. If you would rather it fill the name silently when none was sent, that is a one line change.

Master merged in cleanly, nothing near `HostConnection.ts`.

Cooldude2606 (approved):

PR #1038, opened 2026-09-13
Run TypeScript files without transpiling them

Hornwitser:
Node.js supports running TypeScript files without transpiling them to JavaScript first by the way of typestripping. That means it replaces all the type syntax with whitespace and then runs the code. For this to work we can only use erasable syntax, and we need to fix all imports to reference the TypeScript files instead of the transpiled JavaScript files in `dist/node`.

This is a work in progress showing all the changes needed for this to work.

### Changelog
```
### Changes
- Remove transpilation of TypeScript into JavaScript in favor of running the .ts files directly. #1038
```


Hornwitser:
Given that this more or less conflicts with everything and doesn't cause any user visible changes I will wait with rebasing and finishing this until the list of other open PRs have been reduced.

Cooldude2606:
Other prs are waiting until after 2.0; will be more than happy to review this after 2.0 is released.

PR #1040, opened 2026-09-14
Require config files to be explicitly created

### LINKED ISSUE #548: Creating config files should be an explicit operation

Hornwitser:
The behaviour of creating the config file if it's missing while convenient causes confusion and problems when the tools are run in the wrong folder.  They end up creating a shadow installation and misdirect the user with nonsensical error messages, instead pointing to the real problem which is that the tool can't find the configuration needed to actually run.

Require a config file to be present at startup and refuse to run if this file is missing.  To support the odd use-case of creating the config when missing a `--create-config` flag is added. To conveniently generate a config file to edit out from the "config create" command is added.

Closes #543.
Resolves #548.

### Changelog
```
### Breaking Changes
- Config files now have to be explicitly created. Use the `--create-config` flag to get the old behaviour of creating it if missing. #548
```


packages/lib/src/bootstrap.ts
```diff
@@ -0,0 +1,50 @@
...
+		await fs.unlink(configLockPath).catch(() => {}); // ignore error, file might not exist
+	}
+
+	const isCreateCommand = args._[0] === "config" && args._[1] === "create";
+	const create = args.createConfig || isCreateCommand;
+
+	let config;
+	const configLock = new LockFile(configLockPath);
+	const message = `${create ? "Creating" : "Loading"} config at ${args.config}`;
+	if (location === "control") {
```
Cooldude2606, line 29: I would suggest making this `!create && location === "control"` so that when creating a config it will output the file path.

Cooldude2606 (approved):
One point for consideration, all the code LGTM but the CI does need to be fixed before merging.

packages/lib/src/bootstrap.ts
Hornwitser: Good idea. I've added it to the code.

